PR #11, opened 2019-12-20
allow whitespaces for attributes after '='

imartinat:
Hi,

I had a problem to compile a file because I had a whitespace for an attribute between '=' and '{{' . And the line number given by the parser for this error was several lines before the line with the error, so it took me a while to figure out the error. I though it will be nice to allow whitespaces in the parser to avoid this problem in the future.
Thanks.

### DIFF
--- a/lib/surface/translator/parser.ex
+++ b/lib/surface/translator/parser.ex
@@ -89,10 +89,10 @@ defmodule Surface.Translator.Parser do
     |> concat(whitespace)
     |> optional(
       choice([
-        ignore(string("=")) |> concat(attribute_expr),
-        ignore(string("=")) |> concat(attribute_value),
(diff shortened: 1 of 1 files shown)


msaraiva:
Hi @imartinat!

Thanks for the PR. We're currently working on a new parser that will provide more accurate error messages and better whitespaces handling. I'll let you know when we're finished so we can check if we still need to merge this or not.

Thanks a lot for contributing!

imartinat:
Hi Marlus, 

Thank you for your quick answer. Thank you so much for this great project. I was annoyed by the parser and the line numbers for the errors so I am happy to hear that there will be a new parser. Let me know when it's ready.
Thanks again.

msaraiva:
@imartinat the new parser on master handles those whitespaces now. Thanks a lot for the PR.

PR #37, opened 2020-02-16
Fixes #36

### LINKED ISSUE #36: Support for Live Layouts
LiveView 0.6+ comes with support for Live Layouts; markup that can be used to wrap whatever is rendered by a LiveView. While we can currently use `mount` to specify which Live Layout to use in a `Surface.LiveView`, what's missing is the ability to specify the Live Layout through the `using` macro[ as it's possible to do](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#module-live-layouts) in Phoenix LiveView.

This pattern is useful if a project has its own `using` directive for Live Views and wants to use the same Live Layout most of the time without having to specify it for every LiveView.

wrren:
Allows the Surface LiveView `using` macro to pass options to the Phoenix LiveView `using` macro, which enabled the specification of a default Live Layout.

### DIFF
--- a/lib/surface/live_view.ex
+++ b/lib/surface/live_view.ex
@@ -30,7 +30,7 @@ defmodule Surface.LiveView do
 
   """
 
-  defmacro __using__(_) do
+  defmacro __using__(opts) do
     quote do
       use Surface.BaseComponent, translator: Surface.Translator.LiveViewTranslator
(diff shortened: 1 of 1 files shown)


msaraiva:
@wrren Thanks! ❤️

PR #51, opened 2020-03-02
Don't use Macro.underscore() to convert to kebab case

### LINKED ISSUE #50: css_class transforms periods in class names into slashes
```
<div class={{ "m-1.5": true }}>
</div>
```

Currently generates the following html:
```html
<div class="m-1/5">
</div>
```

However, I'd expect for it to generate the following html:
```html
<div class="m-1.5">
</div>
```

The cause is this function in surface.ex (line 301):
```elixir
  # TODO: Find a better way to do this
  defp to_kebab_case(value) do
    value
    |> to_string()
    |> Macro.underscore()
    |> String.replace("_", "-")
  end
```

Something like the following might give similar functionality without relying on `Macro.underscore()`:
```elixir
  defp to_kebab_case(value) do
    value
    |> to_string()
    |> String.replace(~r/([a-z])([A-Z0-9])/, "\\1-\\2")
    |> String.downcase()
  end
```

and should fix this specific issue; however I'm not sure if any code is relying on anything else that `Macro.underscore()` does as I'm not familiar with that.

For more context on why this is an issue for me, I'm using tailwind css which has class names that include periods.

lnr0626:
Fixes #50 

I'm not confident this is the correct solution as I know it doesn't support everything that `Macro.underscore()` does. I removed the `0-9` that I originally suggested in the issue as it caused one of the css class tests to fail

### DIFF
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -305,8 +305,8 @@ defmodule Surface do
   defp to_kebab_case(value) do
     value
     |> to_string()
-    |> Macro.underscore()
-    |> String.replace("_", "-")
+    |> String.replace(~r/([a-z])([A-Z])/, "\\1-\\2")
+    |> String.downcase()
   end
 
   defp runtime_error(message) do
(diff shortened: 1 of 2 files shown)


msaraiva:
@lnr0626 thanks for the PR!

PR #57, opened 2020-03-12
add helpers for live patch/live redirect

### LINKED ISSUE #54: Create wrapper components for live_patch and friends

lnr0626:
I went for two separate components instead of one with a property as I assumed the phoenix team had reasons for moving away from the single `live_link` helper to separate `patch`/`redirect` helpers

Resolves #54


msaraiva:
@lnr0626 sorry for taking so long to merge it. I was on a small vacation :)

Thanks for the PR!

PR #61, opened 2020-03-31
Add missing comma in .formatter.exs.

mathewdgardner:
Noticed this when I was trying to import the new export.

Also adding `slot/1` and `slot/2` to the export.


msaraiva:
@mathewdgardner thanks!

PR #62, opened 2020-04-01
Drop camelCase to kebab-case auto-conversion

zamith:
Why:

* There are css architectures that use uppercase class names in their
  naming convention, for example, SUITCSS (https://suitcss.github.io/).
  Therefore we don't want to restrict that.

* Allowing binary or atom css classes to go through untouched, even in
  lists. When they're defined as a property we lose control over exactly
  how it's going to be named and that can also be improved, but not as
  big of a deal.

As a side note, I noticed that the project does not use `mix format`.
Any particular reason?

### DIFF
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -301,14 +301,6 @@ defmodule Surface do
     [{:safe, "\""}, value, {:safe, "\""}]
   end
 
-  # TODO: Find a better way to do this
-  defp to_kebab_case(value) do
-    value
-    |> to_string()
-    |> String.replace(~r/([a-z])([A-Z])/, "\\1-\\2")
-    |> String.downcase()
-  end
-
   defp runtime_error(message) do
     stacktrace =
       self()
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -199,7 +199,7 @@ defmodule Surface do
         {class, true} ->
           [to_kebab_case(class) | classes]
         class when is_binary(class) or is_atom(class) ->
-          [class | classes]
+          [to_string(class) | classes]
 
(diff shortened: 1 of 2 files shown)


msaraiva:
Hi @zamith!

Thanks for the PR.

The camel case to kebab case auto-conversion will be removed very soon (before the next alpha version) so I believe the changes in this PR  will no longer be necessary (please, let me know if it still does). The whole discussion can be found [here](https://github.com/msaraiva/surface/issues/52).

> I noticed that the project does not use mix format. Any particular reason?

Yeah, I have mixed feelings about the formatter. I think some of the decisions it makes are really bad, especially regarding indentation. However, I understand the importance of a unified format for all developers so I will eventually start using now as we starting to get more contributions. So I'll make sure I'll run it before releasing the next alpha version.

Cheers.

lib/surface.ex
```diff
@@ -199,7 +199,8 @@ defmodule Surface do
         {class, true} ->
           [to_kebab_case(class) | classes]
```
msaraiva, line 200:
```suggestion
          [to_string(class) | classes]
```

lib/surface.ex
```diff
@@ -199,7 +199,8 @@ defmodule Surface do
...
         class when is_binary(class) or is_atom(class) ->
-          [to_kebab_case(class) | classes]
+          [class | classes]
```
msaraiva, line 202:
```suggestion
          [to_string(class) | classes]
```

zamith:
@msaraiva Done. Also removed `to_kebab_case/1` as it's not used anymore

PR #67, opened 2020-04-15
Creates a TextInput component

zamith:
Why:

* There's a few components that are common enough that we can define
  ourselves, namely related with forms. Moreover that will eventually give
  us the chance to create form wrappers that abstract over repeated
  complexity of implementing said forms.

  Related to msaraiva/surface#32

* Creating a new namespace `Form`
* Adding a very simple `TextInput` component which tries to mimic
  `Phoenix.HTML.Form`. There are a couple of issues with different
  levels of priority that could be addressed before finalizing this:

  1. There is no way to pass a "blob" of properties along. This would be
  needed, because we can't/don't want to set a property in the component
  for every property the html element supports. (msaraiva/surface#2)
  2. It would be nice to be able to only set an attribute if the
  property is non empty. This is particularly interesting for non
  required properties. For example, if no class attribute is passed,
  it still sets `class=""`.

  This might be a moot problem after number 1 is solved, as we can
  simply not set class as property.

### DIFF
--- a/lib/surface/components/form/text_input.ex
+++ b/lib/surface/components/form/text_input.ex
@@ -1,12 +1,37 @@
 defmodule Surface.Components.Form.TextInput do
+  @moduledoc """
+  Generates a text input.
+
+  Provides a wrapper for Phoenix.HTML.Form's `text_input/3` function.
+
+  All options passed via `opts` will be sent to `text_input/3`, `value` and
+  `class` can be set directly and will override anything in `opts`.
+
+
+  ## Examples
+
+  ```
+  <TextInput form="user" field="name" opts={{ [autofocus: "autofocus"] }}>
+  ```
+  """
+
   use Surface.Component
 
   import Phoenix.HTML.Form
 
+  @doc "An identifier for the form"
   property form, :string, required: true
+
+  @doc "An identifier for the input"
   property field, :string, required: true
+
+  @doc "Value to pre-populated the input"
   property value, :string
+
+  @doc "Class or classes to apply to the input"
   property class, :css_class
+
+  @doc "Keyword with options to be passed down to `text_input/3`"
   property opts, :keyword, default: []
 
   def render(assigns) do
--- a/lib/surface/components/form/text_input.ex
+++ b/lib/surface/components/form/text_input.ex
@@ -1,18 +1,23 @@
 defmodule Surface.Components.Form.TextInput do
   use Surface.Component
 
+  import Phoenix.HTML.Form
+
+  property form, :string, required: true
   property field, :string, required: true
   property value, :string, default: ""
   property class, :css_class
 
   def render(assigns) do
     ~H"""
-    <input
-      type="text"
-      class={{ @class }}
-      id={{ @field }}
-      value={{ @value }}
-    />
+    {{
+      text_input(
+        String.to_atom(@form),
+        @field,
+        value: @value,
+        class: @class
+      )
+    }}
     """
   end
 end
--- a/lib/surface/components/form/text_input.ex
+++ b/lib/surface/components/form/text_input.ex
@@ -7,6 +7,7 @@ defmodule Surface.Components.Form.TextInput do
   property field, :string, required: true
   property value, :string, default: ""
   property class, :css_class
+  property opts, :keyword, default: []
 
   def render(assigns) do
     ~H"""
@@ -14,8 +15,10 @@ defmodule Surface.Components.Form.TextInput do
       text_input(
         String.to_atom(@form),
         @field,
-        value: @value,
-        class: @class
+        [
+          value: @value,
+          class: @class,
+        ] ++ @opts
       )
     }}
     """
--- a/lib/surface/components/form/text_input.ex
+++ b/lib/surface/components/form/text_input.ex
@@ -5,7 +5,7 @@ defmodule Surface.Components.Form.TextInput do
 
   property form, :string, required: true
   property field, :string, required: true
-  property value, :string, default: ""
+  property value, :string
   property class, :css_class
   property opts, :keyword, default: []
 
--- a/lib/surface/components/form/text_input.ex
+++ b/lib/surface/components/form/text_input.ex
@@ -1,0 +1,18 @@
+defmodule Surface.Components.Form.TextInput do
+  use Surface.Component
+
+  property field, :string, required: true
+  property value, :string, default: ""
+  property class, :css_class
+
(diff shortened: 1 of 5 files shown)


zamith:
@msaraiva This is just an example to understand if this is what you had in mind. Also, dynamic props would really help here. Not sure if that's something to be worked on before hand.

msaraiva:
Hi @zamith!

Thanks a lot for working on this ❤️

> There is no way to pass a "blob" of properties along. This would be
needed, because we can't/don't want to set a property in the component
for every property the html element supports. (#2)

For the `<TextInput>` and friends we just want to wrap the existing Phoenix functions so in our case we have `text_input(form, field, opts \\ [])` where, according to the docs, _"All given options are forwarded to the underlying input"_

So we could have something like:

```elixir
defmodule Surface.Components.Form.TextInput do
  use Surface.Component

  import Phoenix.HTML.Form

  property form, :string
  property field, :string
  property value, :string, default: ""
  property class, :css_class
  property opts, :keyword, default: []

  context get form, from: Form, as: :context_form
  context get field, from: Field, as: :context_field

  def render(assigns) do
    form = assigns[:form] || assigns[:contex_form]
    field = assigns[:field] || assigns[:contex_field]

    ~H"""
    {{
      text_input(
        String.to_atom(form),
        field,
        [value: @value, class: @class] ++ @opts
      )
    }}
    """
  end
end
```

A couple of remarks:

  * We don't have a type `:keyword` yet, however, it's just a matter of adding it to [this list](https://github.com/msaraiva/surface/blob/master/lib/surface/api.ex#L4-L22)
  * The golden rule to decide if we should create a specific property for an option is:
    * Do we need to customize it? e.g., add synctatic sugar like we do for css classes), if so, we create it, otherwise, we just forward as it is to the phoenix helper.

> It would be nice to be able to only set an attribute if the
property is non empty. This is particularly interesting for non
required properties. For example, if no class attribute is passed,
it still sets class="".

I totally agree. We do this already for boolean attributes. We should extend this behaviour when working with regular HTML tags. Feel free to open an issue for it so we can address it later.

zamith:
@msaraiva updated with the opts and added some tests.

I didn't add context yet because we still don't have form and field as existing components. This works as a standalone, albeit without any context of what surrounds it.

My idea was to merge these building blocks and then work on bringing them together, but I can add a form component in this PR if you think that's the way to go.

lib/surface/components/form/text_input.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Surface.Components.Form.TextInput do
+  use Surface.Component
+
+  import Phoenix.HTML.Form
+
+  property form, :string, required: true
+  property field, :string, required: true
+  property value, :string, default: ""
```
zamith, line 8: `value` does not fall under the rules you set from when we define a property. However, it will likely be used more often than other options, and this makes it more ergonomic than passing the keyword. What do you think?

msaraiva: Agreed 👍

lib/surface/components/form/text_input.ex
```diff
@@ -7,15 +7,18 @@ defmodule Surface.Components.Form.TextInput do
...
     ~H"""
     {{
       text_input(
         String.to_atom(@form),
         @field,
-        value: @value,
-        class: @class
+        [
+          value: @value,
+          class: @class,
```
msaraiva, line 20: I believe `text_input/3` can handle `class` and `value` as `nil`, right? Otherwise, we'll need to conditionally add them to `otps`.

msaraiva:
> I didn't add context yet because we still don't have form and field as existing components. This works as a standalone, albeit without any context of what surrounds it.
>
> My idea was to merge these building blocks and then work on bringing them together, but I can add a form component in this PR if you think that's the way to go.

No problem. We can add `<Form>` next.

> I realize that the way I'm testing adds yet another warning about an alias not being used, although it is being used but Elixir can't know about it. Is this a problem? How do we get rid of it?

Can you send more details? What alias and what message?

zamith:
Ah sorry, since it happens with other aliases as well, I assumed you were aware. 

```
warning: unused alias TextInput
  test/components/form/text_input_test.exs:4
```

Which I actually need, otherwise I get:

```
warning: cannot render <TextInput> (module TextInput could not be loaded)
  code:1: Surface.Components.Form.TextInputTest.TestLiveView_16419.render/1
```

I don't know enough about Elixir's compiler to know for sure, but I believe it can't understand that the alias is being used in the test, because it is only evaluated later.

lib/surface/components/form/text_input.ex
zamith: Not sure what mean. The first test for text_input creates one where I'm not passing a class or value and it works as expected. I actually just added a commit to remove the default for `value`, exactly so that it can be nil and `text_input` will ignore it. Is that what you meant?

msaraiva: Yes! Awesome.

msaraiva:
> because it is only evaluated later on the `Code.eval` on `render_live`

That's right. The compiler has no idea it will be used at runtime. So you have 2 options:

  1. Remove the alias and use the full name: 

```jsx
<Surface.Components.Form.TextInput form="user" ...>
```
  2. Use the alias after defining it: 

```elixir
alias Surface.Components.Form.TextInput; TextInput
```

I would probably go with `#2` but feel free to go with `#1` if you prefer :)

zamith:
#2 is fine. Will seem a bit weird for other people, maybe. But it seems to be the least intrusive.

zamith:
Removed all of the compiler errors if that's ok. :)

mathewdgardner:
> warning: unused alias TextInput
  test/components/form/text_input_test.exs:4

I also ran into this recently and got by it by telling `alias` [to not warn](https://hexdocs.pm/elixir/master/Kernel.SpecialForms.html#alias/2-warnings).

```elixir
alias Surface.Components.Form.TextInput, warn: false
```

Not sure I like this either but it might be more clear than 

```elixir
alias Surface.Components.Form.TextInput; TextInput
```

msaraiva:
@mathewdgardner I wasn't aware of `warn: false`. I think it's better since it makes our intention explicit. Thanks!

@zamith feel free to add `@doc` to each property. There's no need to get fancy, just a short description is fine. You can even copy/paste from `input_text` docs if you want. After that, I think we can merge it 🎉

msaraiva:
@zamith thanks a lot! ❤️

Issue #70, opened 2020-04-17
Cannot use some Phoenix helpers within curly braces


mathewdgardner:
Cannot use Phoenix HTML helpers with do blocks with `{{ }}`.

#### Environment

-   Elixir version: 1.10.2
-   OTP version: 22.3.2
-   Phoenix version: 1.4.16
-   Phoenix LiveView version: 0.11.0
-   Operating system: macOS Catalina 10.15.4

#### Contrived Example

```elixir
defmodule Foo do
  use Surface.Component
  import Phoenix.HTML.Link

  property links, :list, default: ["foo", "bar", "baz"]

  def render(assigns) do
    ~H"""
    <div>
      {{ for l <- @links do }}
        {{ link to: "#" do }}
          <span>{{ l }}</span>
        {{ end }}
      {{ end }}
    </div>
    """
  end
end
```

With this we get a compiler warning and error

```
Compiling 1 file (.ex)
warning: unexpected beginning of EEx tag "<%=" on end of expression "<%= end %>", please remove "=" accordingly
  lib/foo.ex:14


== Compilation error in file lib/foo.ex ==
** (SyntaxError) lib/foo.ex:14: unexpected token: end
    lib/eex/compiler.ex:101: EEx.Compiler.generate_buffer/4
    lib/eex/compiler.ex:54: EEx.Compiler.generate_buffer/4
    expanding macro: Surface.sigil_H/2
    lib/foo.ex:9: Parent.render/1
```

In this contrived example (intentionally not using the `:for` directive) we can move past the error by manually creating the anchor tag, but is this desired behavior?

If we remove the `do` block for `link/2` we lose the error but keep the warning.

Sorry @msaraiva, I don't mean to pummel you with issues. This is great package and I can't wait to see it reach v1.0.0!

zamith:
Since you're not really taking advantage of Surface (at from the example), have you tried using the Raw component?

```elixir
defmodule Foo do
  use Surface.Component
  import Phoenix.HTML.Link

  alias Surface.Components.Raw

  property(links, :list, default: ["foo", "bar", "baz"])

  def render(assigns) do
    ~H"""
    <div>
      <#Raw>
        <%= for l <- assigns[:links] do %>
          <%= link to: "#" do %>
            <span><%= l %></span>
          <% end %>
        <% end %>
      </#Raw>
    </div>
    """
  end
end
```

mathewdgardner:
[...]

Issue #15, opened 2020-01-04
Implement a compiler instead of a translator


msaraiva:
Generate the AST directly instead of an EEx template

mazz-seven:
How can I view what AST structure LiveView expects?

msaraiva:
Hi @mazz-seven!

You can view the translated code of any component or HTML tag by adding the `:debug` directive. 

Example:

```vue
<Hello name="John Doe" :debug/>
```
will print to the standard output something like:

```elixir
<% require Hello %>
<% props = put_default_props(%{ name: "John Doe"}, Hello) %>
<% props = Map.put(props, :__surface__, %{groups: %{__default__: %{binding: false, size: 0}}}) %>
<% props = Map.put(props, :context, context) %>
<% children_props = %{} %>
<%= live_component(@socket, Hello, Keyword.new(Map.merge(props, children_props))) %>
```

mazz-seven:
First, it's very cool to see how things are built behind the scenes. 

What should be the AST for `<Hello name="John Doe" :debug/>`? I thought there is a special AST that LiveView expects. And that you want to go from `<Hello name="John Doe" :debug/>` to AST.

lnr0626:
Is this something you're interested in getting help with?

msaraiva:
Hey @lnr0626!

This one is definitely going to take a lot of effort so YES, I'd love to get some help :)

Since it has a huge impact on almost everything we currently have, my plan is to create a separate project to implement the engine, something like `surface_engine`. As soon as the engine is able to translate all features we have, including directives and slots, we can then replace the translator with the new compiler.

I'll create the project with more information and some initial implementation sometime next week. Then we move on from that. WDYT?

Thank you for your help!

lnr0626:
Sounds great!

msaraiva:
@lnr0626 I finally started working on this. What we need to do is to create something similar to the [EEX Compiler](https://github.com/heliohead/elixir/blob/master/lib/eex/lib/eex/compiler.ex). The problem is that since EEX doesn't care about the structure of the template (there's no hierarchy), the `EEx.Tokenizer.tokenize/3` can easily generate a list of sequential tokens, but Surface, on the other hand, needs to know the structure of the template so our parser must generate a tree of nodes instead. 

[...]

PR #71, opened 2020-04-17
Add Link component.

mathewdgardner:
Why:

This provides a component that wraps `Phoenix.HTML.link/2` and passes through all options given to it. This is needed so that we can add to our built-ins that provide developers all the tools they need to write component based applications.

This is a follow up to #70

### DIFF
--- a/lib/surface/components/link.ex
+++ b/lib/surface/components/link.ex
@@ -1,0 +1,60 @@
+defmodule Surface.Components.Link do
+  @moduledoc """
+  Generates a hyperlink.
+
+  Provides a wrapper for Phoenix.HTML.Link's `link/2` function.
+
+  All options passed via `opts` will be sent to `link/2`, `label` and `class` can
+  be set directly and will override anything in `opts`.
+
+  ## Examples
+  ```
+  <Link
+    label="user"
+    to="/users/1"
+    class="is-danger"
+    opts={{ method: :delete, data: [confirm: "Really?"] }}
+  />
+
+  <Link
+    to="/users/1"
+    class="is-link"
+  >
+    <span>user</span>
+  </Link>
+  ```
+  """
+
+  use Surface.Component
+
+  import Phoenix.HTML.Link, only: [link: 2]
+
+  @doc "Place to link to"
+  property to, :string, required: true
+
+  @doc "Class or classes to apply to the link"
+  property class, :css_class
+
+  @doc "Keyword with options to be passed down to `link/2`"
+  property opts, :keyword, default: []
+
+  @doc """
+  The label for the generated `<a>` alement, if no content (default slot) is
+  provided.
+  """
+  property label, :string
+
+  @doc """
+  The content of the generated `<a>` element. If no content is provided,
+  the value of property `label` is used instead.
+  """
+  slot default
+
+  def render(assigns) do
+    children = ~H"<slot>{{ @label }}</slot>"
+
+    ~H"""
+    {{ link [to: @to, class: @class] ++ @opts, do: children }}
+    """
+  end
+end
--- a/test/components/link_test.exs
+++ b/test/components/link_test.exs
@@ -1,0 +1,59 @@
+defmodule Surface.Components.LinkTest do
+  use ExUnit.Case
+
+  alias Surface.Components.Link, warn: false
+
+  import ComponentTestHelper
+
+  describe "Without LiveView" do
+    test "creates a link with label" do
+      code = """
+      <Link label="user" to="/users/1" />
+      """
+
+      assert render_live(code) =~ """
+             <a href="/users/1">user</a>
+             """
+    end
+
+    test "creates a link without label" do
+      code = """
+      <Link to="/users/1" />
+      """
+
+      assert render_live(code) =~ """
+             <a href="/users/1"></a>
+             """
+    end
+
+    test "creates a link with default slot" do
+      code = """
+      <Link to="/users/1"><span>user</span></Link>
+      """
+
+      assert render_live(code) =~ """
+             <a href="/users/1"><span>user</span></a>
+             """
+    end
+
+    test "setting the class" do
+      code = """
+      <Link label="user" to="/users/1" class="link" />
+      """
+
+      assert render_live(code) =~ """
+             <a class="link" href="/users/1">user</a>
+             """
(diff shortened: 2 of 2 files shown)


mathewdgardner:
@msaraiva  Any idea how to truly sync my fork with your base?

test/components/link_test.exs
```diff
@@ -0,0 +1,39 @@
...
+  use ExUnit.Case
+
+  alias Surface.Components.Link, warn: false
+
+  import ComponentTestHelper
+
+  describe "Without LiveView" do
+    test "creates a link" do
+      code = """
+      <Link text="user" to="/users/1" />
```
zamith, line 11: This looks good as a first version, but have you given thought as to adding support for children instead of `text`?

mathewdgardner: Good catch, I started off with a default slot. I'll give it a whirl.

msaraiva: Maybe it's a good idea to rename `text` to `label` and create the default slot the same way as [LiveRedirect](http://surface-demo.msaraiva.io/builtincomponents/LiveRedirect). See source [here](https://github.com/msaraiva/surface/blob/master/lib/surface/components/live_redirect.ex#L48).

zamith: You mean by always using the `link/2` version that takes a block?

mathewdgardner: I'm not sure if it's intended but this works for children but breaks the other use cases.

```elixir
    children = ~H"""
    <slot />
    """

    ~H"""
    {{ link [to: @to, class: @class] ++ @opts, do: children }}
    """
```

 Is there a way we can know if a slot is being used to match off of? If default slot is used, use `link/2` with do block, else use `link/2` without do block and require `text` property.

msaraiva: If the default slot is not set, `@inner_content`, i.e., `assigns[:inner_content]` should be `nil`. Please let me know if it doesn't.

mathewdgardner: This works for all cases so far:

```elixir
  def render(assigns) do
    case assigns.inner_content do
      nil ->
        ~H"""
        {{ link @text, [to: @to, class: @class] ++ @opts }}
        """

      _ ->
        children = ~H"""
        <slot />
        """

        ~H"""
        {{ link [to: @to, class: @class] ++ @opts, do: children }}
        """
    end
  end
```

msaraiva: Sweet! 

I wonder, however, why the following doesn't work for both cases:

```elixir
  children = ~H"""
  <slot>{{@text}}</slot>
  """

  ~H"""
  {{ link [to: @to, class: @class] ++ @opts, do: children }}
```
I'll check later. You can go with the current implementation since it works :)

mathewdgardner: Ha well that does work and I like it better so I'm going to steal it! Less is more

mathewdgardner: But it does add superfluous newlines `\n`, any idea why this happens?

zamith: That is added at the parser level. Every rendered output of the `H` sigil will have a new line, which makes sense most of the time. I created a PR with support for an `i` option to make the newline go away. The idea being that the rendered output is to be inlined.

Not sure if it is a good idea, but it solve this problem.

mathewdgardner: Yea I have a bunch of components that piece together uses of `~H` and I always get a bunch of white space with it. Wasn't sure why theses weren't trimmed.

zamith: I’d be interested with which cases those might be. I’m not sure my PR makes sense, but maybe it does.

You can try it in your examples, see what you think

msaraiva: There are few cases where an extra `\n` may be added. As @zamith pointed out, using `~H"""` is one of them. You should be able to avoid this one by adding `\` to the end of the text as mentioned here https://github.com/msaraiva/surface/pull/72#issuecomment-615835594. Another option is to use `~H"..."` or `~H(...)`, etc.

There are other scenarios where additional spaces might be added too but those cases are related to code injection during translation. Most of them will be tackled in #15.

So, I wouldn't worry much about those extra spaces for now. As soon we have #15 done, we'll certainly come back to this issue.

msaraiva:
BTW, I'll push a small change to the translator adding syntactic sugar for properties of type `:keyword`. This way we can pass the options without the `[...]`, just like we do for `:css_class` and `:props`.

```
<Link opts={{ method: :delete, csrf_token: "token" }} />
```

lib/surface/components/link.ex
```diff
@@ -0,0 +1,62 @@
...
+  All options passed via `opts` will be sent to `link/2`, `label` and `class` can
+  be set directly and will override anything in `opts`.
+
+  ## Examples
+  ```
+  <Link
+    label="user"
+    to="/users/1"
+    class="is-danger"
+    opts={{ [method: :delete, data: [confirm: "Really?"]] }}
```
msaraiva, line 16: If you rebase your code, you'll be able to pass the options without the outer brackets:

```
opts={{ method: :delete, data: [confirm: "Really?"] }}
```

test/components/link_test.exs
msaraiva: I just realized that adding `\` does not work for any of the `~H`, `~L` and `~E` sigils :(

@mathewdgardner if you want to get rid of those `\n` in the tests, you can replace the code with:

```elixir
  children = ~H"<slot>{{ @label }}</slot>"

  ~H"{{ link [to: @to, class: @class] ++ @opts, do: children }}"
```

lib/surface/components/link.ex
mathewdgardner: @msaraiva Rebased and squashed

lib/surface/components/link.ex
```diff
@@ -0,0 +1,60 @@
...
+  property label, :string
+
+  @doc """
+  The content of the generated `<a>` element. If no content is provided,
+  the value of property `label` is used instead.
+  """
+  slot default
+
+  def render(assigns) do
+    children = ~H"<slot>{{ @label }}</slot>"
```
zamith, line 54: Nice 👍

msaraiva:
@mathewdgardner thanks! ❤️

PR #72, opened 2020-04-17
Adds an option for inline parsing of elements

zamith:
Why:

* In some situations we might was to render Surface code to inline into
  other pieces of Surface code

* Adding an optional `i` to the H sigil that removes the final new line

Note: Not even sure if we want this, but it came up in #71 and I decided to give it a go.

### DIFF
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -58,8 +58,10 @@ defmodule Surface do
   @doc """
   Translates Surface code into Phoenix templates.
   """
-  defmacro sigil_H({:<<>>, _, [string]}, _) do
+  defmacro sigil_H({:<<>>, _, [string]}, opts) do
     line_offset = __CALLER__.line + 1
+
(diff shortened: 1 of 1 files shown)


msaraiva:
Hi @zamith!

Is there a similar `i` modifier in `~L` and `~E`? If so, I have no problem adding this feature to keep consistency between them. Otherwise, I think we can just use `\` at the end which achieves the same goal. Example:

```elixir
  ~H"""
  <div>
    <span>Whatever</span>
  <\div>\
  """

zamith:
[Doesn't seem like it](https://github.com/phoenixframework/phoenix_live_view/blob/v0.12.0/lib/phoenix_live_view/helpers.ex#L320). I guess that workaround is fine. I haven't actually needed it yet, to be honest.

I was trying to figure out where the ` \n` was coming from which is also a good opportunity for me to learn a bit more about the internals of Surface. I'm more than happy to close this PR.

PR #76, opened 2020-04-19
Adds Submit button

zamith:
Why:

* In order to have a functional form this is missing and should be part
  of the built in components

* Adding a Submit component that forwards options to Phoenix.HTML.Form
* Ensuring it works with children via a default slot

### DIFF
--- a/lib/surface/components/form/submit.ex
+++ b/lib/surface/components/form/submit.ex
@@ -1,0 +1,33 @@
+defmodule Surface.Components.Form.Submit do
+  @moduledoc """
+  Generates a submit button to send the form.
+
+  All options are forwarded to the underlying `Phoenix.HTML.Form.submit/3`
+  """
+
+  use Surface.Component
+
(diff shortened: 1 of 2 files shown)


lib/surface/components/form/submit.ex
```diff
@@ -0,0 +1,33 @@
...
+
+  All options are forwarded to the underlying `Phoenix.HTML.Form.submit/3`
+  """
+
+  use Surface.Component
+
+  import Phoenix.HTML.Form, only: [submit: 2]
+
+  @doc "The text to be used in the button"
+  property text, :string
```
msaraiva, line 13: We should either call it `label` to keep it consistent with the other components or change all `label`s to `text`. Although the helpers call it `text`, I believe `label` describes better. WDYT?

zamith: Makes sense. Phoenix calls it value which I don't think makes sense here, so I was debating what to call it.

msaraiva:
@zamith thanks! ❤️

PR #78, opened 2020-04-20
Fix unicode conversion error

MrYawe:
Fixes https://github.com/msaraiva/surface/issues/77

From the [documentation](https://hexdocs.pm/elixir/IO.html#iodata_to_binary/1):
> Notice that this function (iodata_to_binary/1) treats integers in the given IO data as raw bytes and does not perform any kind of encoding conversion. If you want to convert from a charlist to a UTF-8-encoded string, use chardata_to_string/1 instead. For more information about IO data and chardata, see the "IO data" section in the module documentation.

### DIFF
--- a/lib/surface/translator.ex
+++ b/lib/surface/translator.ex
@@ -47,7 +47,7 @@ defmodule Surface.Translator do
     |> prepare(caller)
     |> prepend_context()
     |> translate(caller)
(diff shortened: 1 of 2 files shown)


msaraiva:
@MrYawe thanks! ❤️

PR #79, opened 2020-04-21
implements input_* form elements

alexandrubagu:
I've implemented all input_* form elements @msaraiva  @zamith

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,3 @@ erl_crash.dump
 
 # Ignore package tarball (built via "mix hex.build").
 surface-*.tar
-
--- a/lib/surface/components/form/color_input.ex
+++ b/lib/surface/components/form/color_input.ex
@@ -20,7 +20,7 @@ defmodule Surface.Components.Form.ColorInput do
   import Phoenix.HTML.Form, only: [color_input: 3]
   import Surface.Components.Form.Utils
 
-  alias Surface.Components.Form, warn: false
+  alias Surface.Components.Form
 
   @doc "An identifier for the form"
   property form, :form
--- a/lib/surface/components/form/color_input.ex
+++ b/lib/surface/components/form/color_input.ex
@@ -37,6 +37,21 @@ defmodule Surface.Components.Form.ColorInput do
   @doc "Keyword list with options to be passed down to `color_input/3`"
   property opts, :keyword, default: []
 
+  @doc "Triggered when the component loses focus"
+  property blur, :event
+
+  @doc "Triggered when the component receives focus"
+  property focus, :event
+
+  @doc "Triggered when the component receives click"
+  property capture_click, :event
+
+  @doc "Triggered when a button on the keyboard is pressed"
+  property keydown, :event
+
+  @doc "Triggered when a button on the keyboard is released"
+  property keyup, :event
+
   context get form, from: Form, as: :form_context
 
   def render(assigns) do
--- a/lib/surface/components/form/color_input.ex
+++ b/lib/surface/components/form/color_input.ex
@@ -57,13 +57,14 @@ defmodule Surface.Components.Form.ColorInput do
   def render(assigns) do
     form = get_form(assigns)
     props = get_non_nil_props(assigns, [:value, :class])
+    event_opts = get_events_to_opts(assigns)
 
     ~H"""
     {{
       color_input(
         form,
         String.to_atom(@field),
-        props ++ @opts
+        props ++ @opts ++ event_opts
       )
     }}
     """
--- a/lib/surface/components/form/color_input.ex
+++ b/lib/surface/components/form/color_input.ex
@@ -1,23 +1,23 @@
-defmodule Surface.Components.Form.TextInput do
+defmodule Surface.Components.Form.ColorInput do
   @moduledoc """
-  Generates a text input.
+  Generates a color input.
 
-  Provides a wrapper for Phoenix.HTML.Form's `text_input/3` function.
+  Provides a wrapper for Phoenix.HTML.Form's `color_input/3` function.
 
-  All options passed via `opts` will be sent to `text_input/3`, `value` and
+  All options passed via `opts` will be sent to `color_input/3`, `value` and
   `class` can be set directly and will override anything in `opts`.
 
 
   ## Examples
 
   ```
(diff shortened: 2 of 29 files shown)


msaraiva:
Hi @alexandrubagu! 

Awesome job 🎉

I have a question for discussion before we merge this. As far as I could see, some of the inputs have the same properties and generate the same `<input...>`. The only difference is the `type`. Do you think it would be better to have a single component for those cases with a `type` property?, e.g. `<Input type="date"/>`, etc? Or do you feel that each one of them deserves its own component?

For the record, I'm not sure yet which approach is better so I'm trying to see the pros and cons :)

@zamith feel free to share your thoughts too.

.gitignore
```diff
@@ -22,3 +22,4 @@ erl_crash.dump
 # Ignore package tarball (built via "mix hex.build").
 surface-*.tar
 
+.elixir_ls
```
zamith, line 25: If this is something that you want to ignore but is relevant only to your setup you can use `.git/info/exclude`. Or do you think this should be here for everyone?

lib/surface/components/form/color_input.ex
```diff
@@ -0,0 +1,56 @@
...
+  Provides a wrapper for Phoenix.HTML.Form's `color_input/3` function.
+
+  All options passed via `opts` will be sent to `color_input/3`, `value` and
+  `class` can be set directly and will override anything in `opts`.
+
+
+  ## Examples
+
+  ```
+  <TextInput form="user" field="name" opts={{ [autofocus: "autofocus"] }}>
```
zamith, line 14:
```suggestion
  <ColorInput form="user" field="name" opts={{ [autofocus: "autofocus"] }}>
```

lib/surface/components/form/range_input.ex
```diff
@@ -0,0 +1,62 @@
...
+  Provides a wrapper for Phoenix.HTML.Form's `range_input/3` function.
+
+  All options passed via `opts` will be sent to `range_input/3`, `value` and
+  `class` can be set directly and will override anything in `opts`.
+
+
+  ## Examples
+
+  ```
+  <RangeInput form="volume" field="percent" opts={{ [min: "0", max: "100"] }}>
```
zamith, line 14:
```suggestion
  <RangeInput form="volume" field="percent" min="0" max="100" />
```

Since you've added the min and max as explicit properties, maybe the example could reflect that.

.gitignore
msaraiva: One nice alternative is to add `.elixir_ls` to `~/.gitignore_global`. That's what I do so I don't have to set it up in every project.

msaraiva:
Regarding "creating own component" vs "using a generic input". One advantage of the current implementation (having one component for each function) is that we end up with a 1:1 mapping between them, which may be more intuitive for users.

I've just noticed we've missed events for both, the inputs and the form. I believe for the inputs we have at least `phx-blur`, `phx-focus`, `phx-click`, `phx-capture-click`, `phx-keydown`, `phx-keyup` and for the form, `phx-change` and `phx-submit`.

I guess we should define them without the `phx-` prefix, like:

```elixir
property blur, :event

property focus, :event
```

**Note**: It's important that we add the events explicitly, otherwise, Surface will not set the `phx-target` automatically and the default target will be the parent live view.

.gitignore
alexandrubagu: Thanks, I didn't know.

lib/surface/components/form/range_input.ex
alexandrubagu: The first attempt was to pass `min` and `max` to options. Thanks

zamith:
Agree. It was in my plans to add those. As for the 1:1 mapping, I don't think there is much different in terms of being intuitive from:

```
<TextInput ...>
<PasswordInput ...>
```
to
```
<Input type="text" ...>
<Input type="password" ...>
```

It both makes it "simpler" in terms of maintainability for us and more closely resembles actual HTML. That being said I don't have a strong preference either way, the main reason I leaning towards a generic component is having less code to deal with (E.g. we want to add the events, it would be changes to one file, instead of 10)

alexandrubagu:
@msaraiva @zamith  I've updated the PR but I don't know how to proceed further. From my point of view I prefer to have them in separate components as they are defined in `Phoenix.HTML.Form`.

Another idea is to have `Surface.Components.Form.GenericInput` and we could use metaprogramming to generate modules with the body from `GenericInput` for these input elements
```
[
  DateInput,
  ColorInput,
  TextInput,
  PasswordInput,
  EmailInput,
  HiddenInput,
  TimeInput,
  TelephoneInput
]
```

lib/surface/components/form/range_input.ex
```diff
@@ -0,0 +1,77 @@
...
+  <RangeInput form="volume" field="percent" min: "0" max: "100"/>
+  ```
+  """
+
+  use Surface.Component
+
+  import Phoenix.HTML.Form, only: [range_input: 3]
+  import Surface.Components.Form.Utils
+
+  alias Surface.Components.Form, warn: false
```
msaraiva, line 23: You shouldn't need `warn: false` here since `Form` is being used in `context get`. In case you remove it and you start receiving warnings, please let me know, it's a bug.

msaraiva:
@alexandrubagu I took a look at the specification of `input` at https://www.w3schools.com/tags/tag_input.asp

It looks like some the attributes are only applicable to some of the types. For instance, an input of type `date` might have attributes `step`, `min` and `max` but no `placeholder` nor `maxLength`. An input of type `text` has `placeholder` and `maxLength` but no `step`, `min` or `max`. That means having one component per type would give us the opportunity to improve UX by adding those properties explicitly. 

So let's move on with separate components. Thanks a lot, @alexandrubagu for your work and sorry for coming back and forth on this.

alexandrubagu:
@msaraiva we are in early stages I don't see this as a problem. 

I've updated the PR according to discussion from https://github.com/msaraiva/surface/issues/82

lib/surface/components/form/range_input.ex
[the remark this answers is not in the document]
alexandrubagu: In tests I've tried to remove `warn: false` and I got some warnings, so I kept them in place

msaraiva:
@alexandrubagu thanks! ❤️

Issue #73, opened 2020-04-18
Do not generate attributes for HTML tags if the values are `nil`


msaraiva:
> It would be nice to be able to only set an attribute if the
property is non empty. This is particularly interesting for non
required properties. For example, if no class attribute is passed,
it still sets `class=""`.

We do this already for boolean attributes. We should extend this behaviour when working with regular HTML tags as none of them accepts `nil` anyway.

PR #80, opened 2020-04-21
[Issue #73] Do not generate attributes for nil values

paultannenbaum:
Proposed fix for #73 

Works for both cases when an attribute string is empty, or has empty whitespace. Not very pretty but still getting up to speed on this codebase.

### DIFF
--- a/lib/surface/translator/tag_translator.ex
+++ b/lib/surface/translator/tag_translator.ex
@@ -169,7 +169,7 @@ defmodule Surface.Translator.TagTranslator do
 
   defp translate_attribute_assignment(key, value, %{spaces: spaces}) do
     [space1, space2, space3] = spaces
-    [space1, key, space2, "=", space3, wrap_unsafe_value(key, value)]
+    [space1, "<%= attr(\"", key, "\",", space2, value_to_code(value), ") %>", space3]
   end
 
   defp value_to_code({:attribute_expr, expr, _}) do
--- a/lib/surface/translator/tag_translator.ex
+++ b/lib/surface/translator/tag_translator.ex
@@ -81,8 +81,7 @@ defmodule Surface.Translator.TagTranslator do
   defp translate_attributes(attributes) do
     for {key, value, meta} <- attributes do
       value = replace_attribute_expr(value)
-
-      if value_is_empty_list?(value), do: [], else: translate_attribute(key, value, meta)
+      translate_attribute(key, value, meta)
     end
   end
 
@@ -203,11 +202,5 @@ defmodule Surface.Translator.TagTranslator do
 
   defp replace_attribute_expr(value) do
     value
-  end
-
-  defp value_is_empty_list?(value) do
-    whitespace_codepoint = 32
-
-    value |> is_list and value |> Enum.filter(&(&1 !== whitespace_codepoint)) |> Enum.empty?
   end
 end
--- a/lib/surface/translator/tag_translator.ex
+++ b/lib/surface/translator/tag_translator.ex
@@ -81,7 +81,8 @@ defmodule Surface.Translator.TagTranslator do
   defp translate_attributes(attributes) do
     for {key, value, meta} <- attributes do
       value = replace_attribute_expr(value)
-      translate_attribute(key, value, meta)
+
+      if value_is_nil?(value), do: [], else: translate_attribute(key, value, meta)
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/surface/translator/tag_translator.ex
```diff
@@ -200,4 +201,8 @@ defmodule Surface.Translator.TagTranslator do
   defp replace_attribute_expr(value) do
     value
   end
+
+  defp value_is_nil?(value) do
+    is_list(value) and value |> Enum.filter(&(&1 !== 32)) |> Enum.empty?
```
zamith, line 206: What about moving the `is_list` to a guard? I understand it would required adding another clause to the function, but it makes it a bit more clear what's happening, don't you think?

Also, if we are going to filter the empty spaces, what about moving the code 32 to a module variable like `@whitespace 32`, so it is clearer what this is doing?

zamith: In terms of the feature itself I'll leave it up to @msaraiva, but I'm not sure we need to support removing empty strings as well. Do you have a specific use case in mind where it would be useful?

paultannenbaum: > What about moving the is_list to a guard? I understand it would required adding another clause to the function, but it makes it a bit more clear what's happening, don't you think?

From this implementation, `value_is_nil?` needs to return a boolean, so adding a guard to the function would not pattern match on non list types, which would then need to be handled in a separate function. Is this what you are suggesting? I do agree this function is very ugly, and could prob be cleaned up. 

> Also, if we are going to filter the empty spaces, what about moving the code 32 to a module variable like @whitespace 32, so it is clearer what this is doing?

Yeah, thats prob a good idea for readability purposes. I'll let @msaraiva chime in on whether he thinks it is worth having this or not, as I agree it may not be worth it.

paultannenbaum: Updated, the code just a bit. I implemented your whitespace idea, but just made it a local function variable instead of a module variable, seemed like the better route.

msaraiva:
Hi @paultannenbaum!

Thanks a lot for working on this.

The proposed solution will not work as expected because `translate_attributes/1` runs at compile-time, so if you pass an expression and the expression is evaluated to `nil` at runtime, it will not remove the attribute. What we need is move that logic to runtime. Something like:

On [Surface.Translator.TagTranslator:168](https://github.com/msaraiva/surface/blob/d6045c6dc26c00737b3dfd33b0b7bc1ca71cdf4c/lib/surface/translator/tag_translator.ex#L168), replace the code with:

```elixir
  defp translate_attribute_assignment(key, value, %{spaces: spaces}) do
    [space1, space2, space3] = spaces
    [space1, "<%= attr(\"", key, "\",", space2, value_to_code(value), ") %>", space3]
  end
```

On [Surface](https://github.com/msaraiva/surface/blob/d6045c6dc26c00737b3dfd33b0b7bc1ca71cdf4c/lib/surface.ex), create our new `attr/2` function:

```elixir
  @doc false
  def attr(_key, nil) do
    []
  end

  def attr(key, value) do
    value = attr_value(key, value)
    [key, "=", quot(value)]
  end
```

Don't forget to add a test for the `nil` case:

```elixir
  test "attributes as nil" do
    assigns = %{id: nil}

    code = ~H"""
    <p id={{ @id }}>Hello World</p>
    """

    assert render_static(code) =~ """
            <p >Hello World</p>
            """
  end
```

### Remarks

  * The solution will work with any HTML attribute except `class`, `style` and `phx-*` since they have their own implementation for `translate_attribute_assignment/3`. That means we need to update them too in order to have the same behaviour. We can certainly leave that to another PR, if you prefer.
  * Don't worry about the extra space in `<p >`. It's not an invalid HTML. We will get rid of it after #15.
  * I believe one or two tests of the translator might fail since they test against the generated `eex`. But it's just a matter of replacing them with the updated code that uses `attr(...)`.

Please let me know if you need any further assistance. 

Thanks!

msaraiva:
BTW, I think we should remove the attribute only when the value is `nil`. After all, empty strings are still valid strings.

lib/surface/translator/tag_translator.ex
zamith: > Is this what you are suggesting?

Yes. Having two clauses. But with @msaraiva's suggested changes there's no need for this.

msaraiva:
Resolved in https://github.com/msaraiva/surface/issues/105.

PR #84, opened 2020-05-05
Implements datetime_local_input, reset, radio_button and textarea

alexandrubagu:
@msaraiva Can you take a look over this PR ?

Summary:
-  implements textarea
- implements radio_button
- implements reset
- implements datetime_local_input
- improves input moduledocs

### DIFF
--- a/lib/surface/components/form/color_input.ex
+++ b/lib/surface/components/form/color_input.ex
@@ -11,7 +11,7 @@ defmodule Surface.Components.Form.ColorInput do
   ## Examples
 
   ```
-  <ColorInput form="user" field="name" opts={{ [autofocus: "autofocus"] }}>
+  <ColorInput form="user" field="color" opts={{ [autofocus: "autofocus"] }}>
   ```
   """
 
--- a/lib/surface/components/form/color_input.ex
+++ b/lib/surface/components/form/color_input.ex
@@ -11,46 +11,14 @@ defmodule Surface.Components.Form.ColorInput do
   ## Examples
 
   ```
-  <ColorInput form="user" field="color" opts={{ [autofocus: "autofocus"] }}>
+  <ColorInput form="user" field="color" opts={{ autofocus: "autofocus" }}>
   ```
   """
 
-  use Surface.Component
+  use Surface.Components.Form.Input
 
   import Phoenix.HTML.Form, only: [color_input: 3]
   import Surface.Components.Form.Utils
-
-  alias Surface.Components.Form
-
-  @doc "An identifier for the form"
-  property form, :form
-
-  @doc "An identifier for the input"
-  property field, :string, required: true
-
-  @doc "Value to pre-populated the input"
-  property value, :string
-
-  @doc "Class or classes to apply to the input"
-  property class, :css_class
-
-  @doc "Keyword list with options to be passed down to `color_input/3`"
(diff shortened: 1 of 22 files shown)


msaraiva:
Hi @alexandrubagu!

I think the time has come to centralize those properties/events/context definitions. We could start with something simple like creating a module `Input` and moving all default definitions there.

```elixir
defmodule Surface.Components.Form.Input do
  @moduledoc false

  defmacro __using__(_) do
    quote do
      use Surface.Component

      alias Surface.Components.Form

      @doc "An identifier for the form"
      property form, :form

      @doc "An identifier for the input"
      property field, :string, required: true

      @doc "Value to pre-populated the input"
      property value, :string

      @doc "Class or classes to apply to the input"
      property class, :css_class

      @doc "Keyword list with options to be passed down to `color_input/3`"
      property opts, :keyword, default: []

      property blur, :event

      property focus, :event

      @doc "Triggered when the component receives click"
      property capture_click, :event

      @doc "Triggered when a button on the keyboard is pressed"
      property keydown, :event

      @doc "Triggered when a button on the keyboard is released"
      property keyup, :event

      context get form, from: Form, as: :form_context
    end
  end
end
```

Then each input type could just `use` that module to get all default definitions and then, if necessary, add any specific property for that type.

```elixir
defmodule Surface.Components.Form.ColorInput do
  @moduledoc """
  Generates a color input.
  ...
  """
  use Surface.Components.Form.Input

  import Phoenix.HTML.Form, only: [color_input: 3]
  import Surface.Components.Form.Utils

  def render(assigns) do
    ...
  end
end
``` 

We can still improve it in the future if we decide to have an `<Input>` component. For now, I think that would be the best choice to minimize all that boilerplate. WDYT?

lib/surface/components/form/color_input.ex
```diff
@@ -11,7 +11,7 @@ defmodule Surface.Components.Form.ColorInput do
   ## Examples
 
   ```
-  <ColorInput form="user" field="name" opts={{ [autofocus: "autofocus"] }}>
+  <ColorInput form="user" field="color" opts={{ [autofocus: "autofocus"] }}>
```
msaraiva, line 14: You can remove the brackets in all examples if you want :)

```
<ColorInput form="user" field="color" opts={{ autofocus: "autofocus" }}>
```

alexandrubagu: Indeed :+1:

alexandrubagu:
I wanted to suggest if we can extract this into a generic input because code is very repetitive. I will do this. 

Thanks

alexandrubagu:
@msaraiva I've tried to centralize the code as you described but I've discovered that moving `context get form, from: Form, as: :form_context` in a separate module doesn't work. Works only if `context get ...` is in same file as the input definition.

What do you suggest ? Centralize only properties or is a bug in the current implementation?

Thanks,
Alex

msaraiva:
@alexandrubagu It's probably a bug. My suggestion is to centralize `properties` + `events` (which is most of the stuff anyway) and leave `context` out for now. I'll take a look later and we can fix it in a separate PR.

Thanks!

alexandrubagu:
@msaraiva Done. Thanks !

msaraiva:
@alexandrubagu awesome. Thanks! ❤️

PR #97, opened 2020-07-19
Add support for co-located .sface templates

### LINKED ISSUE #22: Support loading external templates and css files
Implement a default `render/1` that automatically loads files with the same base name + `.sface` and `.css`. 

Example:

```
live/
├── components/
│   ├── card.css
│   ├── card.ex
│   ├── card.sface
│   ├── hero.css
│   ├── hero.ex
│   └── hero.sface
│
...
```

For `*.sface` files, the implementation is straightforward, however, for `*.css` we should probably wait until we have scoped styles (#1).

lnr0626:
Resolves #22 

I patterned this closely off how live view handles the same functionality, but I couldn't figure out how to write tests for it. It should be somewhat straightforward to add support for stylesheets once we add that functionality.

I added the before compile hooks to the individual types of component instead of BaseComponent because MacroComponent uses BaseComponent as well and I figured that shouldn't have a render function automagically defined.

 I could see adding support for falling back to a `.html.leex`, but haven't added that in. WDYT?

### DIFF
--- a/lib/surface/component.ex
+++ b/lib/surface/component.ex
@@ -36,13 +36,13 @@ defmodule Surface.Component do
       end
 
     quote do
+      @before_compile Surface.Renderer
       use Phoenix.LiveComponent
       use Surface.BaseComponent, translator: unquote(translator)
       use Surface.API, include: [:property, :slot, :context]
       import Phoenix.HTML
 
       @behaviour unquote(__MODULE__)
-      @before_compile Surface.Renderer
       @before_compile Surface.ContentHandler
 
       if unquote(translator) == Surface.Translator.SlotableTranslator do
--- a/lib/surface/component.ex
+++ b/lib/surface/component.ex
@@ -42,6 +42,7 @@ defmodule Surface.Component do
       import Phoenix.HTML
 
       @behaviour unquote(__MODULE__)
+      @before_compile Surface.Renderer
       @before_compile Surface.ContentHandler
 
       if unquote(translator) == Surface.Translator.SlotableTranslator do
--- a/lib/surface/live_component.ex
+++ b/lib/surface/live_component.ex
@@ -55,6 +55,7 @@ defmodule Surface.LiveComponent do
       use Phoenix.LiveComponent
       use Surface.BaseComponent, translator: Surface.Translator.LiveComponentTranslator
 
+      @before_compile Surface.Renderer
       @before_compile unquote(__MODULE__)
 
       use Surface.API, include: [:property, :slot, :data, :context]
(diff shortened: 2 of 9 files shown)


msaraiva:
Hey @lnr0626!

Fantastic job!

> I couldn't figure out how to write tests for it.

You can try to create a component for each type along with their respective `.sface` templates and save them in the `test/support` folder, then you can use them your tests just like you'd use any other component. Something like:

```elixir
defmodule Surface.RendererTest do
  use ExUnit.Case, async: true

  alias Surface.Components.ComponentWithExternalTemplate
  alias Surface.Components.LiveComponentWithExternalTemplate
  alias Surface.Components.LiveViewWithExternalTemplate

  import ComponentTestHelper

  defmodule View do
    use Surface.LiveView

    def render(assigns) do
      ~H"""
      <ComponentWithExternalTemplate/>
      <LiveComponentWithExternalTemplate/>
      <LiveViewWithExternalTemplate/>
      """
    end
  end

  test "Component rendering external template" do
    assert render_live(View) =~ """
           the rendered content of the component
           """
  end
  
  test "LiveComponent rendering external template" do
    assert render_live(View) =~ """
           the rendered content of the live component
           """
  end
  
  test "LiveView rendering external template" do
    assert render_live(View) =~ """
           the rendered content of the live view
           """
  end
end
```

> I could see adding support for falling back to a `.html.leex`, but haven't added that in. WDYT?

I think we could start accepting only `.sface` files for now. We can then add support for `.html.leex` in the future.

lnr0626:
Thanks! That did it.

msaraiva:
@lnr0626 thanks! ❤️

matthewp:
This looks great. One question, do you need your top-level component to be inlined in render() as these tests show or can it be external as well? If so what would that look like?

lnr0626:
@matthewp you should not need to have the top level component inlined

matthewp:
@lnr0626 great, how would you render it then? (sorry if this seems obvious, I'm new to Phoenix)

lnr0626:
This doesn't change how you use the views/components, so you should be able to use them like normal. If you're using the view within the router, the you can do:
```
defmodule MyRouter do
  ...
  live "/route", MyView
  ...
end
```

if you're using it within another surface component/view, then you should be able to do something like:
```
defmodule MyOtherView do
  ...
  def render(assigns) do
    ~H"""
    <MyView />
    """
  end
end
```

If you're running into issues feel free to create a new issue or join the elixir slack - it's hard to see questions and responses on this PR because it's already closed

PR #104, opened 2020-08-10
[Issue #73] Don't render attribute on nil value

edgarmiadzieles:
Since https://github.com/msaraiva/surface/pull/80 seems to have been under construction for more than 2 months with no recent updates I propose a similar solution in this PR with fixed tests.

### DIFF
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -336,8 +336,12 @@ defmodule Surface do
   end
 
   @doc false
-  def phx_event(_phx_event, value) when is_binary(value) do
-    value
+  def phx_event(_phx_event, nil) do
+    []
+  end
+
+  def phx_event(phx_event, value) when is_binary(value) do
+    [phx_event, "=", quot(value)]
   end
 
   def phx_event(phx_event, value) do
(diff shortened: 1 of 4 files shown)


msaraiva:
Hey @edgarmiadzieles!

Thanks so much for this PR. I need to ask you to put this on hold since we're about to replace the current translator with a new compiler. Unfortunately, the changes in this PR will not work with the new changes. The good news is that the new model provides a much more consistent way for handling this kind of features. The work is already being reviewed and should be merged anytime today. In case you still want to contribute and add this feature after we merge the compiler, please let me know, I'll be glad to give you some directions so you can implement it using the new model.

Cheers.

edgarmiadzieles:
> Hey @edgarmiadzieles!
> 
> Thanks so much for this PR. I need to ask you to put this on hold since we're about to replace the current translator with a new compiler. Unfortunately, the changes in this PR will not work with the new changes. The good news is that the new model provides a much more consistent way for handling this kind of features. The work is already being reviewed and should be merged anytime today. In case you still want to contribute and add this feature after we merge the compiler, please let me know, I'll be glad to give you some directions so you can implement it using the new model.
> 
> Cheers.

Thanks for the feedback @msaraiva . I would definitely like to see this feature in the near future so I will definitely take a look at the changes and try to implement this in the new codebase.

PR #105, opened 2020-08-11
[Issue #73] Don't render attribute on nil value

edgarmiadzieles:


msaraiva:
@edgarmiadzieles thanks! ❤️

I'm leaving https://github.com/msaraiva/surface/issues/73 open since we still need to handle the `class` attribute, however, I think we should change a couple of things before doing that. My idea is to centralize all runtime handling like this one in a single function. Something like:

In `eex_engine.ex`:

```elixir
  defp to_html_attributes([
         %AST.Attribute{name: name, value: [%AST.AttributeExpr{value: value} = expr]}
         | attributes
       ]) do
    new_expr =
      quote generated: true do
        Surface.attr(unquote(to_string(name)), unquote(value))
      end

    [%{expr | value: new_expr} | to_html_attributes(attributes)]
  end
```

In `surface.ex`:

```elixir
  @doc false
  def attr(_attr, nil) do
    ""
  end

  def attr(name, value) do
    # Not sure yet if this is best way to do this
    Phoenix.HTML.raw([" ", name, "=", ~S("), attr_value(name, value), ~S(")])
  end
```

And then since all attributes go through this function, including `class` and `phx-*` we have a single place to handle `nil` values (or whatever we need) and we can use `attr_value` to customize the value as needed, for instance:

```elixir
def attr_value("class", value) do
  # Instead of injecting `css_class` earlier at compile-time. We handle it here, at runtime.
  css_class(value)
end

def attr_value(":on-" <> event, value) when event in @events do
  event_value(value)
end

# fallback for any other attribute
def attr_value(name, value) do
  ...
end
```

@lnr0626 WDYT?

lnr0626:
@msaraiva would the idea be to have `Helpers.attribute_expr_to_quoted!` normalize the different types into whatever they should be (e.g. make sure `:map` is a map, `:keyword` is a keyword list, `:list` is a list, `:css_class` is a keyword list, etc.), and then use `attr_value` when those are passed as an attribute `Tag`/`VoidTag`?

It would probably be worthwhile to pass the type through to that `attr_value` method instead of relying on the attribute names themselves. We'd also have to handle this for the `%AST.DynamicAttribute`s, though I don't think that would be an issue. something like (assuming we add the type):
```elixir
  defp to_html_attributes([
         %AST.DynamicAttribute{name: name, expr: %AST.AttributeExpr{value: value} = expr}
         | attributes
       ]) do
    new_expr =
      quote generated: true do
        for {name, {type, value}} <- unquote(value) do
            Surface.attr(to_string(name), type, value)
        end
      end

    [%{expr | value: new_expr} | to_html_attributes(attributes)]
  end
```

Currently the attributes named `:on-phx-<event>` are handled by the `Surface.Directive.Events` directive, which handles converting them from and `:event` into one or two dynamic properties with the appropriate name (`phx-<event>` and `phx-target`). would you envision changing that?

msaraiva:
@lnr0626 I want to do two things:

1. Minimize the amount of code that needs to be written for each individual type. 
2. Avoid generating code that represents runtime logic.

### Examples

1. Instead of many implementations of [Surface.Compiler.Helpers.attribute_expr_to_quoted!/4](https://github.com/msaraiva/surface/blob/298bde311289683956d79e9d2930214e3880292e/lib/surface/compiler/helpers.ex#L26), I want to have just one generic implementation that normalizes the values of the expression and set the `value` field.

2. Instead of many implementations of `to_html_attributes` generating defferent [logic](https://github.com/msaraiva/surface/blob/298bde311289683956d79e9d2930214e3880292e/lib/surface/compiler/eex_engine.ex#L680-L697), I want to move any runtime type-specific logic `Surface.attr(...)`

The goal is to make the code easier to maintain and adapt to new features, avoiding changing things in many different places depending on the type. I believe the only type that might need a different approach is the `:generator` since it holds an incomplete expression, that needs to be handled at compile-time. 

I'm doing some experiments right now. I'll let you know when I have something to show.

Issue #113, opened 2020-08-14
Defining update/2 in liveComponent results in error


edgarmiadzieles:
Whenever I try to define a update/2 in surface livecomponent it results in an error like

```
[error] GenServer #PID<0.8245.0> terminating
 ** (KeyError) key :__surface__ not found in: %{id: "bottom_modbar", selected_users: []}
     (app) lib/app/live/module/show/module_bottom_modbar.ex:1: App.ModuleLiveView.ModuleBottomModbar.update/2
     (phoenix_live_view) lib/phoenix_live_view/utils.ex:392: Phoenix.LiveView.Utils.maybe_call_update!/3
     (phoenix_live_view) lib/phoenix_live_view/diff.ex:235: anonymous fn/3 in Phoenix.LiveView.Diff.update_component/3
     (phoenix_live_view) lib/phoenix_live_view/diff.ex:184: Phoenix.LiveView.Diff.with_component/5
     (phoenix_live_view) lib/phoenix_live_view/diff.ex:234: Phoenix.LiveView.Diff.update_component/3
     (phoenix_live_view) lib/phoenix_live_view/channel.ex:99: Phoenix.LiveView.Channel.handle_info/2
     (stdlib) gen_server.erl:637: :gen_server.try_dispatch/4
     (stdlib) gen_server.erl:711: :gen_server.handle_msg/6
     (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
```

The first time it calls update I can see __surface__ being present, but its seems like Surface.LiveComponent quoted update doesn't have it any longer.

msaraiva:
Resolved in #114.

PR #114, opened 2020-08-14
[Issue  #113] Check if key __surface__ is present in assigns

edgarmiadzieles:
Let me know if I misundestood this issue. But it seems that if assigns change there might not be \__surface__ around the second time.


msaraiva:
@edgarmiadzieles thanks! ❤️

PR #115, opened 2020-08-15
Add bahaviour TypeHandler and implement it for common types

msaraiva:
@lnr0626 this PR tries to move any code that depends on a specific type to a single handler. Currently (and certainly before the compiler), this logic is spread through the whole codebase which makes maintaining the code much harder. 

The PR also makes the values of properties and attributes consistent as they are all normalized before we start to manipulate them. 

Expressions can now be also normalized before we use them so we don't have to add hacks for each type like wrapping the values in `[...]` or `identity(...)` anymore. All expressions should be normalized to lists of `clauses` + `opts`, if it doesn't, there's a syntax error. Then each type handler knows how to convert that information into the actual value. This provides a consistent way to add syntactic sugar for types like `css_class`, `:event`, `map`, `keyword`, etc.

There is still some code to migrate to the new handlers but I believe we're almost there. I think it's better to create the PR already so we can discuss improvements before merging it. I also added a few `# TODO: [Type]` indicating some points in the code that either still needs to be migrated or is marked for discussion.

Please, feel free to add suggestions, comments or event implement some of the TODOs, if you want :)

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -64,6 +64,8 @@ defmodule Surface.Compiler do
     :typemustmatch
   ]
 
+  @phx_event_attributes Surface.Directive.Events.phx_events() |> Enum.map(&String.to_atom/1)
+
   @void_elements [
     "area",
     "base",
@@ -489,6 +491,7 @@ defmodule Surface.Compiler do
 
     attr_value =
       case type do
+        # TODO: [Type] Move this logic to type handlers
         type when type in [:string, :css_class, :any] ->
           %AST.Text{
             value: ""
@@ -574,10 +577,14 @@ defmodule Surface.Compiler do
     ]
   end
 
+  # TODO: [Type] Move this logic to TypeHandler
   defp determine_attribute_type(nil, :class, _meta), do: :css_class
 
   defp determine_attribute_type(nil, name, _meta) when name in @boolean_tag_attributes,
     do: :boolean
+
+  defp determine_attribute_type(nil, name, _meta) when name in @phx_event_attributes,
+    do: :phx_event
 
   defp determine_attribute_type(nil, _name, _meta), do: :string
 
@@ -658,10 +665,11 @@ defmodule Surface.Compiler do
     collect_attr_values(attribute_name, meta, values, type, {[codepoint | codepoint_acc], acc})
   end
 
-  defp attr_value(name, :event, value, meta) do
+  # TODO: [Type] Find a way remove this guard or move this logic to Surface.Types
+  defp attr_value(name, type, value, meta) when type in [:css_class, :map, :keyword, :event] do
     %AST.AttributeExpr{
       original: value,
-      value: Helpers.attribute_expr_to_quoted!(Macro.to_string(value), name, :event, meta),
+      value: Helpers.attribute_expr_to_quoted!(Macro.to_string(value), name, type, meta),
       meta: meta
     }
   end
--- a/lib/surface/compiler/eex_engine.ex
+++ b/lib/surface/compiler/eex_engine.ex
@@ -102,6 +102,7 @@ defmodule Surface.Compiler.EExEngine do
     |> maybe_print_expression(comprehension)
   end
 
+  # TODO: Should we move this logic to Surface.Directive.If?
   defp to_expression(
          %AST.If{condition: %AST.AttributeExpr{value: condition}, children: children} =
            conditional,
@@ -399,10 +400,10 @@ defmodule Surface.Compiler.EExEngine do
     !!value
   end
 
-  defp to_prop_expr([%AST.AttributeExpr{value: expr}], :boolean) do
-    quote generated: true do
-      !!unquote(expr)
-    end
+  defp to_prop_expr([%AST.AttributeExpr{value: value, meta: meta}], type)
+       # TODO: [Type] Remove this guard after `update_prop_expr/3` works for all types
+       when type in [:css_class, :map, :keyword, :event, :boolean] do
+    Surface.TypeHandler.update_prop_expr(type, value, meta)
   end
 
   defp to_prop_expr([%AST.AttributeExpr{value: {_, value}}], :list), do: value
@@ -613,6 +614,8 @@ defmodule Surface.Compiler.EExEngine do
          %AST.Attribute{name: name, type: type, value: [%AST.Text{value: value}]}
          | attributes
        ])
+       # TODO: [Type] Should we have something like Surface.TypeHandler.literal_to_html(...) which
+       # would allow type specific validation? e.g. we could warn on disable="false".
        when type == :boolean or is_boolean(value) do
     if value do
       [
@@ -626,28 +629,12 @@ defmodule Surface.Compiler.EExEngine do
   end
 
   defp to_html_attributes([
-         %AST.DynamicAttribute{expr: %AST.AttributeExpr{value: value_expr} = expr} | attributes
+         %AST.DynamicAttribute{expr: %AST.AttributeExpr{value: expr_value} = expr} | attributes
        ]) do
     value =
       quote generated: true do
-        for {name, {type, value}} <- unquote(value_expr) do
-          case type do
-            :boolean ->
-              if value do
-                [" ", to_string(name)]
-              else
-                []
-              end
-
-            _ ->
-              [
-                " ",
-                to_string(name),
-                unquote(Phoenix.HTML.raw("=\"")),
-                value,
-                unquote(Phoenix.HTML.raw("\""))
-              ]
-          end
+        for {name, {type, value}} <- unquote(expr_value) do
+          Surface.attr(name, type, value)
         end
       end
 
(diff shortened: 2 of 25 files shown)


lnr0626:
This is great! I took a quick look through, and really like the direction. 

I'll look through a bit more in-depth and take a stab at some of the TODOs this weekend

msaraiva:
I forgot to mention, this should resove https://github.com/msaraiva/surface/issues/73 and https://github.com/msaraiva/surface/issues/103 too.

msaraiva:
> I'll look through a bit more in-depth and take a stab at some of the TODOs this weekend

@lnr0626 I'm starting working on [these TODOs](https://github.com/msaraiva/surface/blob/a0a110764181ce64b13596875b220edcf7faf99a/lib/surface/compiler/helpers.ex#L26-L93). In case you've picked any one of them, please let me know. Let's avoid having both of us working on the same stuff at the same time :)

lnr0626:
I'll let you know, I haven't had a chance to start yet

msaraiva:
@lnr0626 I just added the last types needed to get everything working using the new type handlers. We still have 6 remaining `# TODO: [Type]` which we need to decide the best way to tackle them. Since you're more familiar with the related code, I'd love to hear your thoughts on how you think we should adjust the API to address them.

Although I believe the new TypeHandler API can certainly be improved, it was a good first round and it was nice to get more familiar with the new compiler.

lib/surface/compiler/eex_engine.ex
```diff
@@ -723,6 +671,10 @@ defmodule Surface.Compiler.EExEngine do
     ]
   end
 
+  # TODO: [Type] We should probably convert an attribute value containing interpolation
```
lnr0626, line 674: I agree with this - normalizing the attribute values to either a single `AST.Text` or a single `AST.AttributeExpr` would simplify handling this throughout the rest of the code. 

It may make sense to handle lists of values in the `TypeHandler`, instead of directly in the compiler - the handling for them is somewhat type specific (i.e. a string makes sense to just join, but what does it mean for a css_class property to be specified as `"str_1 {{ @str_2 }}"`, and some (keyword, map, list) are probably just plain errors). I could see adding another method (maybe `list_expr_to_quoted!`?) that delegated handling of individual `AttributeExpr` to `expr_to_quoted!`, then called a similar method on the type handler to convert the list of `Text` and `AttributeExpr` into a single `AttributeExpr`.

This could potentially handle the empty list case as well (where the user specifies an empty string as the value - i.e. `class=""`), and could allow us to convert the list of expressions and/or text nodes into a single expression or text node.

lnr0626: We'd have to add a way for the type handlers to process literal values (similar to your suggested `literal_to_html`, but used at the same stage as `expr_to_quoted!` instead of `attr_to_html`

lib/surface/compiler/eex_engine.ex
```diff
@@ -613,6 +614,8 @@ defmodule Surface.Compiler.EExEngine do
          %AST.Attribute{name: name, type: type, value: [%AST.Text{value: value}]}
          | attributes
        ])
+       # TODO: [Type] Should we have something like Surface.TypeHandler.literal_to_html(...) which
```
lnr0626, line 617: We should be able to rely on `Surface.attr` here (or directly on TypeHandler.attr_to_value`) instead of inserting the result directly. The one thing I'm unsure about is because that uses `Phoenix.HTML.raw`, we may have to unwrap that. If the unwraping became an issue, we could go for the `literal_to_html` route and just have that return the same thing as attr_to_html (but without the call to raw).

An alternative is to remove that call to `Phoenix.HTML.raw` in the type handler, and just call that in `Surface.attr` on the result of `attr_to_html`

lib/surface/compiler/eex_engine.ex
```diff
@@ -399,16 +400,10 @@ defmodule Surface.Compiler.EExEngine do
...
-    end
+  defp to_prop_expr([%AST.AttributeExpr{value: value, meta: meta}], type) do
+    Surface.TypeHandler.update_prop_expr(type, value, meta)
   end
 
-  defp to_prop_expr([%AST.AttributeExpr{value: {_, value}}], :list), do: value
-
-  defp to_prop_expr([%{value: value}], :string), do: value
-
   defp to_prop_expr(values, :string) do
```
lnr0626, line 407: This variant should also be removed once we combine all values in a list into a single

lib/surface/compiler/eex_engine.ex
msaraiva: > what does it mean for a css_class property to be specified as `"str_1 {{ @str_2 }}"`, and some (keyword, map, list) are probably just plain errors)

I think we shouldn't accept that at all! An attribute value wrapped in `""` is a string and as such, it should behave just like any other string in Elixir. The interpolation should work the same way as `#{}` works, i.e. it should accept only:

  * strings
  * integers representing Unicode code points
  * a list containing one of these three elements
  *  values of types that implement the `String.Chars` protocol

We could even drop support for `{{...}}` inside attributes values to force the user to use `#{}` for interpolation. That would make it work just like React, which only accept interpolation on attributes using JS expressions like: 

```
title={"Hello " + name}
````
 or 
```js
title={`Hello ${name}`}
```

I find the `{{...}}` nicer than `#{}` in the context of a template because it provides a standard way for interpolation. However, I'd have no problem removing it if it would make Surface more consistent :)

> We'd have to add a way for the type handlers to process literal values (similar to your suggested `literal_to_html`, but used at the same stage as `expr_to_quoted!` instead of `attr_to_html`.

I agree. That would also allow us to easily provide static type-based validation for literals, which is pretty cool.

lib/surface/compiler.ex
```diff
@@ -574,11 +577,15 @@ defmodule Surface.Compiler do
     ]
   end
 
+  # TODO: [Type] Move this logic to TypeHandler
```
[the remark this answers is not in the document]
msaraiva: The question is: should we just move the code as it is to `Surface.TypeHandler` or would it be better to delegate to each type the ability to match its own type? Something like:

```elixir
defmodule Boolean do
  use SurfaceTypeHandler
  ...
  @impl true
  def match_type?(nil, name, _meta) do
    name in @boolean_tag_attributes
  end
end
```

I'll be fine by just moving the code to `Surface.TypeHandler`. I was just wondering if it would be the best approach.

lib/surface/compiler/eex_engine.ex
```diff
@@ -102,6 +102,7 @@ defmodule Surface.Compiler.EExEngine do
     |> maybe_print_expression(comprehension)
   end
 
+  # TODO: Should we move this logic to Surface.Directive.If?
```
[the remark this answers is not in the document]
msaraiva: It makes sense. I didn't have the chance to take a deep look at those yet so this TODO is basically a note for my future self. That's why I didn't add it as a `# TODO [Type]`. I have no intention to change that in this PR so you can just ignore it for now. I'll elaborate on that later when I'm more familiar with the code :)

lib/surface/compiler/eex_engine.ex
msaraiva: Yeah, I think we'll need to make a few experiments to see which approach would be more suitable.

lib/surface/compiler/helpers.ex
```diff
@@ -23,210 +23,8 @@ defmodule Surface.Compiler.Helpers do
...
-        value when is_list(value) ->
-          value
-
-        value ->
-          raise "invalid value for property \"#{unquote(name)}\". Expected a :list, got: #{
-                  inspect(value)
-                }"
-      end
-    end
+  def attribute_expr_to_quoted!(value, attribute_name, type, meta) do
```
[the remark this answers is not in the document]
msaraiva: Probably :)

lib/surface/compiler/eex_engine.ex
```diff
@@ -399,16 +400,10 @@ defmodule Surface.Compiler.EExEngine do
     !!value
   end
 
-  defp to_prop_expr([%AST.AttributeExpr{value: expr}], :boolean) do
-    quote generated: true do
-      !!unquote(expr)
-    end
+  defp to_prop_expr([%AST.AttributeExpr{value: value, meta: meta}], type) do
+    Surface.TypeHandler.update_prop_expr(type, value, meta)
```
msaraiva, line 404: I have the feeling we might not even need the `update_prop_expr` callback at all. What if we just use the `expr_to_quoted` callback to generate different expressions for attributes and properties based on the `module` argument? For instance:

```elixir
def expr_to_quoted(type, attribute_name, clauses, opts, module, original) do
  ast = 
    quote generated: true do
      Surface.TypeHandler.expr_to_value!(
        unquote(type),
        unquote(attribute_name),
        unquote(clauses),
        unquote(opts),
        unquote(module),
        unquote(original)
      )
      
  if module do
    # For properties
    quote generated: true do
      some_wrapper_function(unquote(ast))
    end
  else
    # For attributes
    ast
  end
end
```

I also think we might need to pass a few options to this function to enable the caller to customize the behaviour. For instance, we might have a directive that requires the attribute of type `:list` to be wrapped in `Surface.attr` because it will be treated as runtime value, however, another directive may require the original unwrapped value so it can manipulate the AST at compile-time.

Another alternative is to have separate callbacks for properties and attributes like `attr_expr_to_quoted` and `prop_expr_to_quoted`. I'm not sure yet which is one is better. We'll need to try it out to find out :)

lib/surface/compiler.ex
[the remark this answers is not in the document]
msaraiva: Let's just leave this one as it is. This is just about the compiler detecting the types, not handling them.

miguel-s:
Hi @msaraiva @lnr0626!

I believe this change may have introduced a bug with the built-in components. If I understand this PR correctly, now the `class` property of type `:css_class` converts the input into a list of strings and passes it to the Phoenix helper, which in turn takes that list and joins the strings without space.

Example with the built-in `NumberInput`:

```Elixir
<NumberInput
  opts={{
    autofocus: true,
    step: format_number_input(@currency, :step),
    placeholder: format_number_input(@currency, :placeholder),
    inputmode: format_number_input(@currency, :inputmode),
    pattern: format_number_input(@currency, :pattern)
  }}
  class={{
    "form-input block w-full pr-12 sm:text-sm sm:leading-5",
    "pl-9": String.length(@currency.symbol) == 1,
    "pl-12": String.length(@currency.symbol) > 1
  }}
/>
```

turns into (notice the "class" attribute)

```HTML
<input class="form-inputblockw-fullpr-12sm:text-smsm:leading-5pl-9" id="amount_component_amount" inputmode="decimal" name="amount_component[amount]" pattern="" placeholder="0.00" step="0.01" type="number" autofocus="">
```

PR #117, opened 2020-08-22
New directives: `:props` for components and `:attrs` for tags

### LINKED ISSUE #2: Dynamic props
Similar to React but with valid Elixir syntax:

```jsx
<Component {{..., props}} foo="override" />
```

lnr0626:
These have similar capabilities. `:props` allows providing a map of props to a component which is evaluated at runtime, `:attrs` allows providing a map of attributes for html tags which are evaluated at run time. For both of these, if a prop/attribute of the same name has a value explicitly provided, that will override any dynamic/runtime values provided via `:props`/`:attrs`.

It's a slightly different approach from the current pr
- the props are added as normal props instead of via a `__dynamic_props__` map
- the attributes have to be explicitly passed to the tag (either via `:attrs` or by specifying the attribute explicitly)
- props which aren't specified in the component will emit a runtime warning

As is, I think the remaining bit that's not yet addressed is being able to send props down to underlying components without redefining them all. I'll work on adding the `import_props` macro, which I think should address that.

Resolves #2 
Closes #68

### DIFF
--- a/lib/surface/compiler/eex_engine.ex
+++ b/lib/surface/compiler/eex_engine.ex
@@ -241,7 +241,6 @@ defmodule Surface.Compiler.EExEngine do
   defp handle_dynamic_props(nil), do: []
 
   defp handle_dynamic_props(%AST.DynamicAttribute{expr: %AST.AttributeExpr{value: expr}}) do
-    # TODO: how much processing should we actually do on these values?
     quote generated: true do
       for {name, {type, value}} <- unquote(expr) do
         {name, value}
--- a/lib/surface/compiler/eex_engine.ex
+++ b/lib/surface/compiler/eex_engine.ex
@@ -197,6 +197,7 @@ defmodule Surface.Compiler.EExEngine do
          %ast_type{
            module: module,
            props: props,
+           dynamic_props: dynamic_props,
            templates: templates
          } = component,
          buffer,
@@ -204,6 +205,8 @@ defmodule Surface.Compiler.EExEngine do
        )
        when ast_type in [AST.Component, AST.SlotableComponent] do
     props_expr = collect_component_props(module, props)
+
+    dynamic_props_expr = handle_dynamic_props(dynamic_props)
 
     {do_block, slot_meta, slot_props} = collect_slot_meta(component, templates, buffer, state)
 
@@ -224,7 +227,7 @@ defmodule Surface.Compiler.EExEngine do
         unquote(module),
         Surface.build_assigns(
           unquote(assigns_expr),
-          unquote(props_expr),
+          Keyword.merge(unquote(props_expr), unquote(dynamic_props_expr)),
           unquote(slot_props),
           unquote(slot_meta),
           unquote(module)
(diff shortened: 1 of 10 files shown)


msaraiva:
> I think it would probably feel more natural to be able to reference the `Button` props directly when using the `EnhancedButton` (i.e. `<EnhancedButton label="some label" />` compared to `<EnhancedButton button_props={{ label: "some label" }} />`); though the second is potentially a bit simpler to implement.

I agree. The enhancing an existing component is probably the most common case we'll find so I think if we had to choose a single approach, I would go with `<EnhancedButton label="some label" />`.

> I think this is ready for review - I address the comment about `:debug`, and added the runtime type validation for components.

@lnr0626 I can't see those changes. The last commit is https://github.com/lnr0626/surface/commit/ae4dce28c48f10888a52c9d1cdd4cac497aefff3. Did you push them? :)

lnr0626:
> Did you push them? :)

Whoops, pushed now :-)

lib/surface/compiler/eex_engine.ex
```diff
@@ -224,7 +227,7 @@ defmodule Surface.Compiler.EExEngine do
         unquote(module),
         Surface.build_assigns(
           unquote(assigns_expr),
-          unquote(props_expr),
+          Keyword.merge(unquote(props_expr), unquote(dynamic_props_expr)),
```
msaraiva, line 230: I wonder if we could wrap this logic in a separate function or maybe even move it to `Surface.build_assign` altogether in order to have a cleaner `live_component` call. Currently, the generated code gets too verbose with the two nested `for`:

```jsx
<DivWithProps content="static content" :props={{ class: "text-xs", hidden: false, content: "dynamic props content" }} :debug/>
```
Generates:

```elixir
live_component(
  @socket,
  Surface.DirectivesTest.DivWithProps,
  Surface.build_assigns(
    @__surface__[:context] || [],
    Keyword.merge(
      [class: nil, hidden: nil, content: "static content"],
      for(
        {name, {type, value}} <-
          for(
            {name, value} <-
              Surface.TypeHandler.expr_to_value!(
                :map,
                ":props",
                [],
                [class: "text-xs", hidden: false, content: "dynamic props content"],
                Surface.DirectivesTest.DivWithProps,
                " class: \"text-xs\", hidden: false, content: \"dynamic props content\" "
              ) || [],
            not Enum.member?([:content], name)
          ) do
            {name,
             Surface.TypeHandler.runtime_prop_value!(Surface.DirectivesTest.DivWithProps, name, value, "DivWithProps")}
          end
      ) do
        {name, value}
      end
    ),
    [],
    [__default__: %{size: 0}],
    Surface.DirectivesTest.DivWithProps
  ),
  []
)
```

Is it feasible?

lnr0626: I pushed up a change that moved both handling the prop defaults and the dynamic props into `Surface.build_assigns` which generates a much cleaner `live_component` call:

```elixir
live_component(
  @socket,
  Surface.DirectivesTest.DivWithProps,
  Surface.build_assigns(
    @__surface__[:context] || [],
    [content: "static content"],
    Surface.TypeHandler.expr_to_value!(
      :map,
      ":props",
      [],
      [class: "text-xs", hidden: false, content: "dynamic props content"],
      Surface.DirectivesTest.DivWithProps,
      " class: \"text-xs\", hidden: false, content: \"dynamic props content\" "
    ),
    [],
    [__default__: %{size: 0}],
    Surface.DirectivesTest.DivWithProps,
    "DivWithProps"
  ),
  []
)
```

It also cleaned up/simplified the directive code a bit :-)

PR #120, opened 2020-08-24
take advantage of :attrs for Live{Patch,Redirect}

lnr0626:
This allows folks to add attributes like `aria-*` attributes to the generated anchor tag. I went for `opts` to be consistent with the `Link` component.

i don't think we have any other components currently where this is necessary

### DIFF
--- a/test/components/live_redirect_test.exs
+++ b/test/components/live_redirect_test.exs
@@ -1,0 +1,93 @@
+defmodule Surface.Components.LiveRedirectTest do
+  use ExUnit.Case, async: true
+
+  alias Surface.Components.LiveRedirect, warn: false
+
+  import ComponentTestHelper
+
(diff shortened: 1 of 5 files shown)


lnr0626:
working on adding some tests for `LivePatch` and `LiveRedirect`

test/components/live_redirect_test.exs
```diff
@@ -0,0 +1,93 @@
...
+      </div>
+      """
+    end
+
+    def handle_event(_, _, socket) do
+      {:noreply, socket}
+    end
+  end
+
+  describe "Without LiveView" do
```
msaraiva, line 24: I think those `describe "Without LiveView"` that we have in many test cases don't make sense anymore. We should stop adding them to new tests and later remove them from the other tests.

PR #121, opened 2020-08-26
Fix typo when consuming context

bwittenbrook3:
Fixes missing `.` when consuming `ctx_assigns`. 

Great work on the library so far, really enjoying the developer ergonomics this offers.

### DIFF
--- a/lib/surface/content_handler.ex
+++ b/lib/surface/content_handler.ex
@@ -61,7 +61,7 @@ defmodule Surface.ContentHandler do
   defp data_content_fun(assigns, name, index) do
     fn
       {args, ctx_assigns} ->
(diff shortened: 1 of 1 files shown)


msaraiva:
@bwittenbrook3 thanks for the PR! We missed a test for that scenario. I took a look at [test/context_through_slot_test.exs](https://github.com/msaraiva/surface/blob/24d1162cc25004df9d40a11dd65fca98d7b65c64/test/context_through_slot_test.exs) and we don't have a test for context working with named slots, only `default`. In case you can write one, I'd appreciate 🙂

Otherwise, I'll add it sometime next week.

Cheers.

PR #123, opened 2020-09-01
Slot change tracking fixes

lnr0626:
This doesn't make drastic changes to the API - `:let` and `:prop` still work the exact same. The only difference is that bound props within a template have to be accessed as assigns instead of as variables. I did not change the `:let` binding syntax, but that's something we can do if we decide it's worth it.

As a plus, it removes the kind of wonky `ctx_assigns` variable, and instead just relies on rebinding that assign using `inner_content`

There are still a few tests failing, but I think I have an idea why that is.

### DIFF
--- a/lib/surface/compiler/eex_engine.ex
+++ b/lib/surface/compiler/eex_engine.ex
@@ -140,7 +140,7 @@ defmodule Surface.Compiler.EExEngine do
     slot_content_expr =
       if name == :default do
         quote generated: true do
-          unquote(at_ref(:inner_content)).(unquote(slot_prop_expr))
+          @inner_content.(unquote(slot_prop_expr))
         end
       else
         slot_name_expr = at_ref(name)
@@ -162,7 +162,7 @@ defmodule Surface.Compiler.EExEngine do
     name_to_check = if name == :default, do: :__default__, else: name
 
     quote generated: true do
-      if Enum.member?(unquote(at_ref(:__surface__)).provided_templates, unquote(name_to_check)) do
+      if Enum.member?(@__surface__.provided_templates, unquote(name_to_check)) do
         unquote(slot_content_expr)
       else
         unquote(default_value)
@@ -185,7 +185,7 @@ defmodule Surface.Compiler.EExEngine do
 
     quote generated: true do
       live_render(
-        unquote(at_ref(:socket)),
+        @socket,
         unquote(module),
         unquote(props_expr)
       )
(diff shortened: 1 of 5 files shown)


lnr0626:
update: the tests are now passing :-)

msaraiva:
@lnr0626 I created [this test](https://github.com/msaraiva/surface/blob/ms-fix-change-tracking/test/slot_change_tracking_test.exs#L60) to check if change tracking was working when using slots. As suspected it doesn't. 

Initially, I thought it was because we were passing `@__surface__[:context]` to `Surface.Surface.build_assigns` but it's not. Even when passing just `[]`, the test fails. I've been trying to find out the problem since then without success. This could even be another LV limitation. Who knows?! 

I need to fix this first before adding `<Context>`, otherwise it will be impossible to know if we're introducing new change tracking related bugs or not.

msaraiva:
Ok. The problem was here https://github.com/msaraiva/surface/blob/ms-fix-change-tracking/lib/surface/content_handler.ex#L101

The `for` disables change tracking. I'll fix it for the `default` slot with `size: 1` (the most common case) and then we can take a closer look later and see if it can be fixed for size > 1, assuming, of course, it's not an LV issue (I believe it is).

msaraiva:
It turns out, passing `@__surface__[:context]` also disables the change tracking as initially suspected :)

I believe this one can only be fixed if we analyze the tree at compile-time and only pass that information if both of these conditions apply:

1. The node's parent or any of its ancestors is setting context assigns.
2. The node itself or any of its descendent is using any of those context assings.

lnr0626:
is that true of any assign? i.e. would it still be the case if we replaced all usages of `@__surface__.context`/`@__surface__[:context]` with something like `@__context__` or `@__surface_context__`?

lnr0626:
just played around with that test a bit, and it does look like removing the `__surface__` assign passed to `inner_content` makes the test you created pass. I tried moving the context into an assign (named `__context__`) instead of a value within the `__surface__` map, and it has the same results (you can see that here https://github.com/lnr0626/surface/tree/lr-ctx-instead-of-surface).

It looks like we just need to remove it from the call to `inner_content` instead of the call to `build_assigns` if those two conditions apply

msaraiva:
> is that true of any assign?

It's true for any assign that is received as slot props using `inner_content.(...)` like ` __surface__`.

If we pass a received assign to a child component, change tracking is disabled for that component.

msaraiva:
> It looks like we just need to remove it from the call to `inner_content` instead of the call to `build_assigns` if those two conditions apply

Yeah, that would solve this particular problem indeed. But additionally not passing the whole context to `build_assigns` would add some extra optimization since a component that doesn't need any of the context assigns would not update when one of them changes.

PR #126, opened 2020-09-03
add phx-page-loading to boolean attributes list

lnr0626:
This appears to be used as a boolean attribute, https://hexdocs.pm/phoenix_live_view/js-interop.html#loading-state-and-errors references this attribute

### DIFF
--- a/lib/surface/type_handler.ex
+++ b/lib/surface/type_handler.ex
@@ -66,7 +66,8 @@ defmodule Surface.TypeHandler do
     :required,
     :reversed,
     :selected,
-    :typemustmatch
+    :typemustmatch,
+    :"phx-page-loading"
   ]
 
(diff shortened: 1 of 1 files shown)


msaraiva:
@lnr0626 Did you have a real case where using `phx-page-loading` didn't act as expected? I think after we merged the new TypeHandler, any attributes receiving `nil` or `false` will be suppressed. For instance:

```
<input phx-page-loading={{ false }} phx-foo="bar" />
<input phx-page-loading={{ nil }} phx-foo="bar" />
```
will be translated to:
```
<input phx-foo="bar">
<input phx-foo="bar">
```

I believe we don't have to add attributes to that list anymore. What am I missing?

lnr0626:
It's working as expected, this just cleans up the generated html when it's truthy - it'll currently generate `phx-page-loading=""`.

That does work as expected, so this isn't necessary.

PR #128, opened 2020-09-03
Allow properties to be accumulated

msaraiva:


lnr0626 (approved):
saw one thing i had a question about, but other than that looks good!

lib/surface/compiler/eex_engine.ex
```diff
@@ -251,11 +251,25 @@ defmodule Surface.Compiler.EExEngine do
...
 
   defp collect_component_props(module, attrs) do
-    attrs
-    |> Enum.filter(fn %AST.Attribute{name: prop_name} -> module.__validate_prop__(prop_name) end)
-    |> Enum.map(fn %AST.Attribute{name: prop_name, type: type, value: expr} ->
-      {prop_name, to_prop_expr(expr, type)}
-    end)
+    {props, props_acc} =
+      Enum.reduce(attrs, {[], %{}}, fn attr, {props, props_acc} ->
+        %AST.Attribute{name: prop_name, type: type, type_opts: type_opts, value: expr} = attr
```
lnr0626, line 256: I thought i had posted this, but apparently not :-/

It looks like `attr` isn't used except this line, would it make sense to move this pattern match to the function signature? Unsure if that would make the signature hard to read

msaraiva: Yup. I usually avoid pattern matching against many fields in function heads/clauses, unless of course, the pattern is necessary to differentiate which function/clause should be called. I prefer to have 2 separate easy-to-read lines than having the formatter doing this:

```elixir
  {props, props_acc} =
    Enum.reduce(attrs, {[], %{}}, fn %AST.Attribute{
                                        name: prop_name,
                                        type: type,
                                        type_opts: type_opts,
                                        value: expr
                                      },
                                      {props, props_acc} ->
      cond do
        !module.__validate_prop__(prop_name) ->
          {props, props_acc}

        ...
      end
    end)
```

or this:

```elixir
    {props, props_acc} =
      Enum.reduce(attrs, {[], %{}}, fn
        %AST.Attribute{name: prop_name, type: type, type_opts: type_opts, value: expr},
        {props, props_acc} ->
          cond do
            !module.__validate_prop__(prop_name) ->
              {props, props_acc}

            ...
          end
      end)

```

PR #129, opened 2020-09-03
Context component

msaraiva:
This PR replaces the context API with a new component `<Context/>`.

**Note:** There are many compile-time validations we can do for contexts. I believe it's better to add them in a separate PR. The metadata we'll need for those validations is probably the same that we'll need to optimize change tracking when using contexts.


test/context_test.exs
```diff
@@ -84,6 +89,19 @@ defmodule ContextTest do
...
+  #   code = """
+  #   <OuterUsingInnerContent>
+  #     <Inner/>
+  #   </OuterUsingInnerContent>
+  #   """
+
+  #   assert render_live(code) =~ """
+  #          <span id="field">field from OuterUsingInnerContent</span>\
+  #          """
+  # end
```
msaraiva, line 103: @lnr0626 I'm not sure why this test fails yet. I suspect the ContentHandler is broken for this case.

lib/surface/components/context.ex
```diff
@@ -0,0 +1,140 @@
...
+  end
+
+  defp context_assigns_kw(context, sets, gets) do
+    consolidated_sets =
+      Enum.reduce(sets, %{}, fn set, acc ->
+        {key, value} = normalize_set(set)
+        Map.put(acc, key, value)
+      end)
+
+    updated_context = Map.merge(context, consolidated_sets)
```
lnr0626, line 99:
```suggestion
    updated_context =
      Enum.reduce(sets, context, fn set, acc ->
        {key, value} = normalize_set(set)
        Map.put(acc, key, value)
      end)
```

This should work the same and doesn't require the merge call; though merge is a bit more explicit :-)

lib/surface/content_handler.ex
```diff
@@ -72,7 +74,7 @@ defmodule Surface.ContentHandler do
 
       # TODO [Context]: Update this to be only the appropriate assigns for context
       case args[:__context__] || assigns.__context__ do
-        [] ->
+        %{} ->
```
lnr0626, line 77:
```suggestion
        context when context == %{} ->
```

I believe just `%{}` will always match as that will pattern match any map

test/context_test.exs
lnr0626: I think the issue is that context sets are no longer added to the overall components assigns, and so the ContentHandler doesn't have any way of knowing the updated value of `__context__`. in the following code snippet args[:__context__] would be nil, and so it would fall back to the components assign.__context__:

https://github.com/msaraiva/surface/blob/dc2dc34c7452f81475fc8938de7bbbd712c4f0ef/lib/surface/content_handler.ex#L88-L102

The only way to make this work correctly (that I can think of) is to require passing `__context__: @__context__` when using `inner_content`. We could attempt to inject this automatically (as we already know where the `inner_content` calls are), but that feels like it would be error prone and complex. We could also just document this and strongly recommend always using `<slot />`.

lib/surface/components/context.ex
```diff
@@ -0,0 +1,140 @@
...
+  def get(assigns, key, opts) do
+    {key, _as} = normalize_get({key, opts})
+    assigns.__context__[key]
+  end
+
+  defp slot_kw() do
+    [__slot__: {:__default__, 0}]
+  end
+
+  defp context_assigns_kw(context, sets, gets) do
```
lnr0626, line 92: should we log a warning if both get and set are empty?

lnr0626 (commented):
I added a few suggestions for your consideration, and had a question: do we have any tests for using context with named slots? I looked through and didn't see any, but could have easily missed something

lib/surface/components/context.ex
msaraiva: Yeah, but I think this is one of the many things that can be validated at compile-time. I'll create a new issue so we can list them all and see which ones we should consider for `v0.1.0`. This is certainly one of them.

lib/surface/content_handler.ex
msaraiva: 🤦‍♂️

test/context_test.exs
msaraiva: > We could also just document this and strongly recommend always using `<slot />`

We could but, in this case, we need to find a way to use `<slot/>` when working with multiple named slots. Currently, the only way to render those slots is traversing them and calling `inner_content` explicitly:

```jsx
<tr :for={{ item <- @items }}>
  <td :for={{ col <- @cols }}>
    {{ col.inner_content.(item: item, info: info) }}
  </td>
</tr>
```

We need to accept something like:

```jsx
<tr :for={{ item <- @items }}>
  <td :for={{ col <- @cols }}>
    <slot instance={{ col }} props={{ item: item, info: info }}/>
  </td>
</tr>
```

There's no `instance` attribute defined in the HTML's `<slot>` specification so maybe we should add it as a directive `:instance` or maybe even use the existing `name`:

```jsx
<tr :for={{ item <- @items }}>
  <td :for={{ col <- @cols }}>
    <slot name={{ col }} props={{ item: item, info: info }}/>
  </td>
</tr>
```

Not sure.

msaraiva:
> do we have any tests for using context with named slots?

We should! But since using contexts + `inner_content` is not working, we probably don't :)

I'll take a look at it.

test/context_test.exs
lnr0626: overloading `name` seems like a reasonable approach - though it'd be hard to validate at compile time that the slot was actually defined

lnr0626: I agree - using `slotable` here would feel a bit weird, but I can't quite verbalize why. I think it might have something to do with "slotable" being an adjective - it feels a bit more natural to say "the slotable entry/item" or even just "the entry/item" than "the slotable".

I prefer either `entry` or `item`; but don't have a strong preference between the two, so let's do `:entry`

For directive vs. attribute - are you trying to avoid defining non-standard attributes?

msaraiva: > For directive vs. attribute - are you trying to avoid defining non-standard attributes?

Yes. One thing I want to avoid is confusion when the user needs to remember if something is a `property`, `attribute` or `directive`. Today we have a few very basic rules:

  1. Components have properties which have flexible semantics, i.e. the component's author defines its runtime behaviour.
  2. HTML tags have attributes which are not flexible because we don't own them. We can still provide some syntact surgar but never change its semantics.
  5. Any aditional behaviour that can't be provided using a property or is not already provided by an attribute has to be a directive.

The only attributes we explicitly add behaviour to are `slot` and `name` of `<template>` and `<slot>`, respectvely. However, we only do that because that's exactly the expected behaviour as defined in their specs. We don't change the semantics of `slot` nor `name`.

In order to keep those rules unbroken for `<slot>`, we have two options:

  1. Add the `:entry` directive
  2. Use a component `<Slot>`(or macro component `<#Slot>`) instead of the standard HTML tag `<slot>`

msaraiva: I wonder if we could simplify the whole thing and just create a macro for it. Something like:

In `Surface.ex`:

```elixir
  defmacro render_slot(slot, props \\ []) do
    quote do
      unquote(slot).inner_content.([__context__: @__context__] ++ unquote(props))
    end
  end
```

Then using it:

```jsx
<td :for={{ col <- @cols }}>
  {{ render_slot(col, item: item, info: info) }}
</td>
```

The macro should just replace the call with:

```elixir
col.inner_content.([__context__: @__context__] ++ [item: item, info: info])
```

I tried the code above but it didn't work. I got this:

```
warning: undefined module attribute @__context__, please remove access to @__context__ or explicitly set it before access
```

The generated code is:

```elixir
for(col <- @cols) do
  arg0 = Phoenix.LiveView.Engine.to_safe(render_slot(col, [item: item, info: info]))
  {:safe, ["<span>\n    ", arg0, "\n  </span>"]}
end
```

Apparently the macro didn't expand correctly.

lnr0626: That wouldn't let us handle template `:let` the way that we do know; however I don't know why that didn't expand correctly.

lnr0626: Hm. I would have expected that call to the macro to be expanded

lnr0626: ohh, it's because the macro is expanded after it's passed through the phoenix live view engine - which is what handles changing the `@name` syntax into `assigns.name` and setting up change tracking based on that value.

lnr0626: An idea - we could potentially use the `Context` component to update the `inner_content` assigns to add in `__context__` and pass all other arguments through - it would likely require compiler support to pass the slot assigns and the `@inner_content` assign, and I'm unsure of the complexity; but it would allow continued use of `@inner_content` and `col.inner_content`.... I'm also unsure of the implications for change tracking (though assuming that was an issue, we could potentially track usage of `@inner_content` and `@slot_name` assigns and only do that logic if either of those were referenced)

We could test this out by just passing those assigns in manually (i.e. `inner_content={{ @inner_content }}`), and basically rebind them to be: `fn args -> original_func.([__context__: updated_ctx] ++ args)`. for the slots, we'd have to map over them and update the inner contents within to something similar.

lnr0626: We'd only have to do that within `<Context>` that used set

lnr0626: That would also allow us to (slightly) simplify the conversion code for `<slot />` as we would no longer have to inject `@__context__` (in essence we could delete https://github.com/msaraiva/surface/blob/master/lib/surface/compiler/eex_engine.ex#L133-L136)

It also would make `<slot :props={{ props }} />` identical to `slot_name.inner_content(props)`, and 
```
<slot :props={{ props }}>
  <!-- default content -->
</slot>
```
identical to 
```
if slot_name do
  slot_name.inner_content(props)
else
   "<!-- default content -->"
end
```

msaraiva: Ok. I implemented the `<slot :entry={{...}}/>` and play with it for a while and the result is that I didn't like it :(

Its use is not intuitive and It makes the code much more complex than it should.

> it would likely require compiler support to pass the slot assigns and the @inner_content assign

If you're talking about changing `inner_content` calls at compile-time then I think it's probably the way to go. I might be wrong but I have a feeling that if we to identity `inner_content` calls at compile-time and change them accordingly, the whole thing will become much simpler. Maybe even the whole `ContentHandler` could disappear. 

I'm back to the drawing board.

lib/surface/compiler.ex
```diff
@@ -220,18 +221,20 @@ defmodule Surface.Compiler do
   defp convert_node_to_ast(:slot, {_, attributes, children, node_meta}, compile_meta) do
     meta = Helpers.to_meta(node_meta, compile_meta)
 
-    with name when not is_nil(name) and is_atom(name) <-
-           attribute_value(attributes, "name", :default),
-         {:ok, props, _attributes} <-
-           collect_directives([Surface.Directive.SlotProps], attributes, meta) do
+    # TODO: Validate attributes with custom messages
+    name = attribute_value(attributes, "name", :default)
+    index = attribute_value_as_ast(attributes, "index", %Surface.AST.Text{value: 0}, compile_meta)
```
msaraiva, line 226: Another nice thing to implement `<slot>` as a component is that we could have all this validation for free by just defining `property index, :integer`. Maybe the new `transform` callback along with a few tweaks might allow us to do that soon.

lib/surface/compiler.ex
```diff
@@ -414,6 +418,26 @@ defmodule Surface.Compiler do
     end)
   end
 
+  defp attribute_value_as_ast(attributes, attr_name, default, meta) do
```
msaraiva, line 421: We should probably have a place to centralize this kind of functions. Maybe `AST.Attribute`? I don't like the name `value_as_ast` but using just `value` is confusing. I'd love to have a nice name that could represent `AST.Text | AST. AttributeExpr` but I don't :)

lib/surface/compiler/eex_engine.ex
```diff
@@ -135,17 +136,31 @@ defmodule Surface.Compiler.EExEngine do
         Keyword.merge(unquote(props_expr), __context__: @__context__)
       end
 
+    slot_index =
+      case index_ast do
+        %AST.AttributeExpr{value: expr} -> expr
+        %AST.Text{value: value} -> value
+      end
```
msaraiva, line 143: Same thing here. Should we hava something like `AST.extract_value`?

msaraiva (commented):
@lnr0626 I replaced all calls to `inner_content` with `<slot>` or `<slot index={{...}}>`. All tests are passing. 

A few notes:

1. Using `@__inner_content__` instead of `@inner_content` didn't work. For some reason, the change tracking went crazy so I'm overriding the original one as before.
2. I didn't manage to remove the `ContentHandler` yet but I think there's room now to make its implementation much simpler. One thing that I don't like about it is that it makes debugging much harder. Since now we have all information we need at compile-time, we might be able to move some of that logic to `build_assigns` or something similar. Maybe we can generate the `inner_content` wrappers there an pass them along with the assigns/properties. This way the only thing we have to do at runtime would be to replace them. Was something like that you're suggestion yesterday?

I also left a couple of comments inline.

lib/surface/compiler.ex
lnr0626: Yeah! that would also simplify the compiler and EExEngine code a bit cause we'd be able to remove those cases

lib/surface/compiler.ex
lnr0626: Yeah - there are a couple of similar functions spread throughout the code for operating on the ast that I think would also fall into a similar category. I think moving these into the AST module would make sense, and I think we'll continue finding more of these types of things as we do more work with the AST

lib/surface/ast.ex
```diff
@@ -235,14 +235,16 @@ defmodule Surface.AST.Slot do
 
   ## Properties
       * `:name` - the slot name
+      * `:index` - the index of the slotable entry assinged to this slot
```
lnr0626, line 238:
```suggestion
      * `:index` - the index of the slotable entry assigned to this slot
```

lnr0626 (approved):
This looks good, and I think the Slot component would be worth playing around with. I think one thing we'd need to add is some way to specify that a component prop needs to be a compile time constant vs. resolved at runtime

One thing that came to mind looking at all of the `Enum.with_index(...)` calls, would it make sense to just have some index already embedded in the slot assigns? Something like adding an index property to the map - that would allow something like:

```
<td :for={{ %{index: index} <- @cols }}>
  <slot name="cols" index={{ index }} :props={{ item: item, info: info }}/>
</td>
```

for that last example. I could see that having some chance of conflicting with a user created property, so maybe a different name; but that's a little bit more fluid than having to call `with_index` to use non-default slots when there are multiple values

msaraiva:
> I could see that having some chance of conflicting with a user created property, so maybe a different name; 

Yeah, that's the reason. `index` is a very common name for a property the same way as `default` and `content`. That's why we have `@inner_content` and slots named `default` are renamed to `__default__`, allowing users to create properties with those common names. 

> but that's a little bit more fluid than having to call `with_index` to use non-default slots when there are multiple values

I agree. One thing that could help us here would be "modifiers" (similar to vue.js modifiers). In case you're not familiar with the concept, here are some examples:

```jsx
<td :for.index={{ index <- @cols }}>
  <slot name="cols" index={{ index }} :props={{ item: item, info: info }}/>
</td>
```

```jsx
<td :for.with_index={{ {col, index} <- @cols }}>
  Title: {{ col.title }}
  <slot name="cols" index={{ index }} :props={{ item: item, info: info }}/>
</td>
```

```jsx
<td :for.compact.uniq.with_index={{ {item, index} <- @items }}>
  Title: {{ item.title }}
  Content: <slot index={{ index }}/>
</td>
```

This last one would be the same as: 

```
{item, index} <- @items |> Enum.reject(&is_nil/1) |> Enum.uniq() |> Enum.with_index()
```

We don't have to implement all `Enum.*`, just the most common/useful.

It can also be handy for events, for instance:

```jsx
<TextInput :on-phx-keyup.enter="do_something_on_enter" />
```

PR #130, opened 2020-09-06
Potential solution for `inner_content` issue

lnr0626:
There are two commits on here - the first implements the solution I had in mind, the second shows how we could extend the directive to allow it to automatically wrap content within the `<Context>` component.

If we decided to keep the stuff on the second commit, some of it needs to be cleaned up a bit and we should do something similar for get so that they could be accessed similarly. Either way we may want to switch `:get` to a directive to be consistent between `set`/`get` usage

WDYT?


msaraiva (commented):
@lnr0626 nice work! 

I liked the `transform` solution. I think we can make it less verbose later by moving some of the functions that manipulate the AST to someplace else. Regarding extracting the transformation code to a separate `transformer` or `ast_processor` behaviour, I think it's too early to decide. We can wait for other cases and see if the abstraction is really necessary. For now, if we feel it's too verbose, we can still move it to a separate module and call it directly.

I also left a couple of minor comments for your consideration.

msaraiva:
@lnr0626 the more I look at the amount of complexity we need to add in order to make `slots` + `contexts` + `inner_content` + `change tracking` workings nicely, the more I feel that we should take a step back. It seems to me that overriding multiple `inner_content` in multiple places (Context and ContentHandler) will bite us in the future.

I think we should try to find a way to leave all `inner_content` untouched and instruct the user to always use `<slot>` for those cases. That would be much easier to implement, maintain and avoid conflicts with LV. The problem would be how to render named slots. One idea could be to use the index instead of injecting the `inner_content` into the entry. Something like:

```jsx
<td :for={{ {col, index} <- Enum.with_index(@cols) }}>
  Title: {{ col.title }}
  Content: <slot name="cols" index={{ index }}/>
</td>
```

That would also make it much easier to collect slot meta info at compile-time since we're explicitly using the slot name instead of a variable. That will be important to improve change tracking too.

BTW, I wouldn't even care breaking the "we don't add attributes to existing HTML tags" rule and add the `index` attribute. Maintainability comes first in this case.

WDYT?

lnr0626:
That would be far simpler to implement and very likely far simpler to maintain. It also seems like a reasonable addition to the api

I think we should still consider something like a `transform` callback, as that's a fairly small addition and provides an interesting middle ground between a full blown macro component and a normal component which allows some compile time modifications to the generated AST; however, that should probably have some more discussion around it and can definitely wait until after `0.1.0` if we don't need it for the `Context` component.

I'll ahead and fix the attribute ast typespec (https://github.com/msaraiva/surface/pull/130/files#diff-0423625d36cdf062f1be6a888dc1c2cd) on master

msaraiva:
> I think we should still consider something like a transform callback

Definitely! I'm not only willing to keep the `transform` callback, but also investigate the possibility to use a similar approach for things like `<slot>`, `<template>` and maybe `<style>`. This way we could keep the compiler clean and additionally have a place to centralize all necessary implementation for each tag, just like we do for each component.

> that should probably have some more discussion around it and can definitely wait until after `0.1.0` if we don't need it for the `Context` component

I think we can use the `transform` as you implemented in `Context` for `v0.1.0` and discuss possible improvements for `v0.2.0`.

lnr0626:
> I think we can use the transform as you implemented in Context for `v0.1.0` and discuss possible improvements for `v0.2.0`.

Sounds good! I'll pull that into another PR and add some docs/tests for it

msaraiva:
> I'll pull that into another PR and add some docs/tests for it

Awesome! I'll try to implement the `<slot index={{...}}` and see how it goes.

PR #132, opened 2020-09-06
Test and potential solution for built-in components class issue

miguel-s:
This PR contains a test and a potential solution for issue https://github.com/msaraiva/surface/issues/131 using `NumberInput` as an example.

Please let me know if this looks ok to you, then I'll continue with the rest of built-in components.


msaraiva:
Hi @miguel-s!

Thanks for reporting the issue and submitting a PR. Since now all properties/attributes are normalized to their internal runtime values (e.g. :css_class is converted to a list of strings), I believe we'll need a more generic solution that can cover all types.

Surface's compiler is able to automatically generate code to translate any internal runtime value into a string whenever we pass those values directly as attributes of HTML tags, however, those input controls handle the properties manually so the compiler never has the chance to do its job.

What we need to do here is to expose a function to component authors (probably in the `Surface` module) that can be used to translate those values to strings.

This new function could be called `Surface.prop_to_opts` (I know it's a bad name but it's consistent with the existing `Surface.event_to_opts` function. We can change both names later) and the implementation basically uses [attribute_type_and_opts/3](https://github.com/msaraiva/surface/blob/1f370edffd1a529d3a775ab0434e482dd819b7e7/lib/surface/type_handler.ex#L201) to retrieve the type of the prop and then call [attr_to_html!/3](https://github.com/msaraiva/surface/blob/1f370edffd1a529d3a775ab0434e482dd819b7e7/lib/surface/type_handler.ex#L154) to convert the value.

We're currently making some changes to all input components in https://github.com/msaraiva/surface/pull/129 as we're replacing the context API. As soon as we finish it, I'll be back to this PR so we can add this fix.

Cheers.

miguel-s:
Thanks for the detailed explanation!

I'm not very familiar with the inner workings of Surface, so I'll take the time to understand how `Surface.event_to_opts` is implemented.

In the meanwhile, is it helpful if I continue adding (failing) tests to this PR?

lib/surface.ex
```diff
@@ -248,6 +248,31 @@ defmodule Surface do
...
+    html = Enum.at(html, -2)
+
+    # HACK:
+    # attribute_type_and_opts\3 gives us the internal surface type for a property e.g. :css_class
+    # How do we convert back from type to html attribute? E.g. :css_class -> :class
+    case type do
+      :css_class -> [{:class, html}]
+      _ -> [{type, html}]
+    end
+  end
```
msaraiva, line 274: Yeah, you're right. `Surface.TypeHandler.attr_to_html!` is not exactly what we need. I think we could add a new function `Surface.TypeHandler.attr_to_opts!` that can convert the values directly to opts:

```elixir
  def attr_to_opts!(type, name, value) do
    case handler(type).value_to_html(name, value) do
      {:ok, val} when val in ["", nil, false] ->
        []

      {:ok, val} ->
        [{name, to_string(val)}]

      {:error, message} ->
        IOHelper.runtime_error(message)
    end
  end
```

and then the implementation of `Surface.prop_to_opts` would be:

```elixir
  @doc false
  def prop_to_opts(_prop, nil, _caller) do
    []
  end

  def prop_to_opts(prop, value, caller) do
    module = caller.module
    meta = %{caller: caller, line: caller.line, node_alias: module}
    {type, _opts} = Surface.TypeHandler.attribute_type_and_opts(module, prop, meta)
    Surface.TypeHandler.attr_to_opts!(type, prop, value)
  end
```

Regarding the inputs, we should also rely on the type handlers so we need to change the [get_non_nil_props](https://github.com/msaraiva/surface/blob/ea4ca563846e765bffed508e515dcec006e01baa/lib/surface/components/form/utils.ex#L11) function to this:

```elixir
  defmacro get_non_nil_props(assigns, props) do
    quote do
      unquote(__MODULE__).get_non_nil_props(unquote(assigns), unquote(props), __ENV__)
    end
  end

  def get_non_nil_props(assigns, props, caller) do
    module = caller.module
    meta = %{caller: caller, line: caller.line, node_alias: module}
    Enum.reduce(props, [], fn prop, acc ->
      {key, value} = prop_value(assigns, prop)
      {type, _opts} = Surface.TypeHandler.attribute_type_and_opts(module, key, meta)
      internal_value = Surface.TypeHandler.expr_to_value!(type, key, [value], [], module, inspect(value))
      Surface.TypeHandler.attr_to_opts!(type, key, internal_value) ++ acc
    end)
  end
```

After adding the above changes, everything should work but you will see a bunch of warnings regarding an undefined property `pattern` for component `TelephoneInput`. That's great! It shows how important is to rely on the type handlers. The warnings are correct since we forgot to declare the `pattern` property. Add this to `Surface.Components.Form.TelephoneInput`:

```elixir
  @doc "A regular expression to validate the entered value"
  property pattern, :string
```

and you should not see any warnings anymore. 

Let me know in case I've missed anything.

lib/surface/components/link.ex
```diff
@@ -57,7 +57,7 @@ defmodule Surface.Components.Link do
     children = ~H"<slot>{{ @label }}</slot>"
 
     ~H"""
-    {{ link [to: @to, class: @class] ++ @opts ++ event_to_opts(@click, :phx_click), do: children }}
+    {{ link [to: @to] ++ prop_to_opts(__MODULE__, :class, @class) ++ @opts ++ event_to_opts(@click, :phx_click), do: children }}
```
msaraiva, line 60: You'll have to change this to pass the env instead of the module:

```
    {{ link [to: @to] ++ prop_to_opts(:class, @class, __ENV__) ++ @opts ++ event_to_opts(@click, :phx_click), do: children }}
```

msaraiva (changes requested):

lib/surface.ex
miguel-s: Worked just as you described, thanks for the detailed guide!

I only made a small change from `prop_to_opts(prop, value, caller)` to `prop_to_opts(prop_value, prop_name, caller)`, to be more consistent with `event_to_opts/2`.

msaraiva:
@miguel-s thanks! 🎉

PR #133, opened 2020-09-09
Directive cleanup

lnr0626:


lib/surface/directive/debug.ex
```diff
@@ -1,32 +1,31 @@
...
-      value: directive_value(value, Map.merge(meta, expr_meta)),
-      meta: Helpers.to_meta(attr_meta, meta)
-    }
+  use Surface.Directive,
+    extract: [
+      name: ":debug",
+      type: :static_list
+    ]
+
+  def handle_value(%AST.Text{value: true}, meta) do
```
msaraiva, line 8: I think we should rename `AST.Text` to `AST.Literal` in the near future. After all, it can hold not only text but also boolean and integer.

msaraiva (commented):
I like the general idea 👍

One thing you could try later is to minimize the generated code inside `extract_function/1`. Maybe we can move some of that code to pure functions and then just `quote` the necessary bits to delegate the execution.

lib/surface/directive.ex
```diff
@@ -1,19 +1,95 @@
...
+
+    %AST.AttributeExpr{
+      original: value,
+      value: Surface.TypeHandler.expr_to_quoted!(value, name, type, expr_meta),
+      meta: expr_meta
+    }
+  end
+
+  defp to_ast!(value, name, type, meta) do
+    Surface.TypeHandler.literal_to_ast_node!(type, name, value, meta)
```
lnr0626, line 93: I've noticed the signature of this is inconsistent with `expr_to_quoted!` for parameter ordering

lnr0626:
I'm going to close this, though I'll keep the branch around until i have some time to play around with it some more

PR #134, opened 2020-09-09
Simplify prop_to_opts/3

miguel-s:
Follow-up to PR https://github.com/msaraiva/surface/pull/132

I started using the changes we made in PR https://github.com/msaraiva/surface/pull/132 and found it quite cumbersome and ugly to always have to pass `__ENV__` manually when using `prop_to_opts/3`. So I looked for a solution where we can hide that away, which also led to an opportunity to simplify `get_non_nil_props/2`.

Please let me know what you think.

Cheers

### DIFF
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -249,6 +249,13 @@ defmodule Surface do
   end
 
   @doc false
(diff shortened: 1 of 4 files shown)


msaraiva:
@miguel-s thanks!

PR #135, opened 2020-09-10
Add support for a transform callback

lnr0626:
As discussed in #130. I added it to the base even though it's not currently used for macro components - though it's feasible to do that I couldn't think of a use case which isn't already covered by the `expand` function


msaraiva (approved):
Awesome! 🎉

Issue #106, opened 2020-08-11
[Proposal] Changes to the public API


msaraiva:
Before releasing our first stable version, I'd like to propose a few changes to the API: 

1. Rename the `property` function to just `prop`. 
**Reason:** We're already using `:props`, `__props__`, `__get_prop__` everywhere so let's make it more consistent.

1. Rename all `:on-phx-*` directives to just `:on-*`, e.g. instead of `:on-phx-click`, just `:on-click`.
**Reason:** Surface is already an abstraction on top of LV so I think it's pretty clear that `:on-*` is related to the `phx` bindings.

1. ~Rename `context set` and `context get` to `context_set` and `context_get`.
**Reason:** Simpler implementation, better formatting when not using `import_deps`, better tooling (upcoming autocomplete support)~

lnr0626:
We should also consider always adding an `id` property for `Surface.LiveComponent`s, currently we only add that property if the module exports a `handle_event`, but everywhere else we treat `Surface.LiveComponent` as if they are stateful.

This means that if you don't define a `handle_event` function, and use a property of type `event`, then you'll get an error because it tries to use `@myself`; which is only created when the component has an id.

We should also consider changing the type of the id props we inject to `:string` instead of `:integer`.

PR #136, opened 2020-09-10
ID prop changes

lnr0626:
the id changes I mentioned in #106

### DIFF
--- a/lib/surface/live_component.ex
+++ b/lib/surface/live_component.ex
@@ -67,7 +67,7 @@ defmodule Surface.LiveComponent do
       alias Surface.Components.Context
 
       @doc """
-      The id of the live component (required by LiveView).
+      The id of the live component (required by LiveView for stateful components).
       """
       property id, :string, required: true
     end
--- a/lib/surface/live_component.ex
+++ b/lib/surface/live_component.ex
@@ -64,21 +64,17 @@ defmodule Surface.LiveComponent do
 
       require Surface.ContentHandler
       @before_compile Surface.ContentHandler
-      Module.put_attribute(__MODULE__, :__is_stateful__, true)
       alias Surface.Components.Context
+
+      @doc """
+      The id of the live component (required by LiveView).
+      """
+      property id, :string, required: true
     end
   end
 
   defmacro __before_compile__(env) do
-    [maybe_quoted_id(env), quoted_mount(env), quoted_update(env), quoted_stateful(env)]
-  end
-
-  defp quoted_stateful(env) do
-    stateful = Module.get_attribute(env.module, :__is_stateful__, false)
-
-    quote do
-      def __is_stateful__(), do: unquote(stateful)
-    end
+    [quoted_mount(env), quoted_update(env)]
   end
 
   defp quoted_update(env) do
@@ -129,17 +125,6 @@ defmodule Surface.LiveComponent do
            |> Surface.init()
            |> assign(unquote(defaults))}
         end
-      end
-    end
-  end
-
-  defp maybe_quoted_id(env) do
-    if Module.defines?(env.module, {:handle_event, 3}) do
-      quote do
-        @doc """
-        The id of the live component (required by LiveView).
-        """
-        property id, :integer, required: true
       end
     end
   end
--- a/lib/surface/live_view.ex
+++ b/lib/surface/live_view.ex
@@ -41,7 +41,7 @@ defmodule Surface.LiveView do
       @before_compile unquote(__MODULE__)
 
       @doc "The id of the live view"
-      property id, :integer
+      property id, :string, required: true
 
       @doc """
       The request info necessary for the view, such as params, cookie session info, etc.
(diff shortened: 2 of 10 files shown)


lib/surface/live_component.ex
```diff
@@ -65,20 +65,16 @@ defmodule Surface.LiveComponent do
       @behaviour unquote(__MODULE__)
       require Surface.ContentHandler
       @before_compile Surface.ContentHandler
-      Module.put_attribute(__MODULE__, :__is_stateful__, true)
```
lnr0626, line 68: This was added in the compiler, but ended up not being used - instead we just use the `Surface.LiveComponent`/`Surface.Component` distinction to determine if it's stateful or not

lib/surface/live_component.ex
```diff
@@ -65,20 +65,16 @@ defmodule Surface.LiveComponent do
...
+
+      @doc """
+      The id of the live component (required by LiveView).
+      """
+      property id, :string, required: true
```
msaraiva, line 72: Should it be required? As far as I can remember LV only requires an id when using `handle_event`. Has that changed?

lnr0626: You're correct that LV only requires it when using `handle_event`; however, we're treating `LiveComponent` as if it's stateful and using the `@myself` assign. That assign is only created if the component has an id (and is therefore stateful). I pushed up a test showing the issue

msaraiva: I see. So ideally we could always define the property but make it required only when the component has either a `handle_event` or defines a property of type `:event`. Is there any other case?

lnr0626: Those are the only two cases I've run into - if we wanted to keep it optional when it wasn't strictly required by LV, that should be sufficient.

My rational on keeping it always required would be that we call `LiveComponent` stateful and `Component` stateless in our docs, and the thing that makes it stateful in LV is the `id` prop. I don't have a strong preference though 🙂

msaraiva: Having the `id` required when we know it's required is nice because we add static validation while LV will only complain at runtime. However, if LV allows LiveComponents without `id`, I think we should allow that too. I believe there must be cases when this is useful. I'm just not sure when. 🙂

lnr0626: that makes sense - i think my assumption has been that a `Surface.LiveComponent` without an id is functionally equivalent to a `Surface.Component`

msaraiva: It is equivalent today because of a limitation of LV which does not work properly when you have a `LiveComponent` inside a regular `Component` (I believe is the same issue that doesn't allow using `LiveComponent` inside `content_tag`). My initial goal was to make `Component` available to non-LV projects. Actually the first version didn't depend on LV at all. Unfortunately, I had to implement `Component` as a `LiveComponent` later just to "fix" this issue. 

I'd still love to have a real stateless component that could be used in non-LV projects. It would be very useful.

lnr0626: That makes sense - I'll see how complicated it'd be to switch it to required only if handle_event is exported or an :event property is used

msaraiva: > switch it to required only if handle_event is exported or an :event property is used

Regarding checking the `:event` property, that might be tricky. The `id` would be required only if the live component passes an event to a child component as a string, which is basically the same as passing `{{ ..., target: @myself }}`.

Maybe it's better to leave that one out for now. We can try it later for `v.0.2.0` or higher.

lnr0626: That sounds reasonable, it's easier to go from required to optional than optional to required. In that case, this PR may be ready to merge

lib/surface/live_view.ex
```diff
@@ -41,7 +41,7 @@ defmodule Surface.LiveView do
       @before_compile unquote(__MODULE__)
 
       @doc "The id of the live view"
-      property id, :integer
+      property id, :string, required: true
```
msaraiva, line 44: I think the `id` of a Liveview is required only if it's a child Liveview. According to the docs:

> When rendering a child LiveView, the :id option is required to uniquely identify the child. A child LiveView will only ever be rendered and mounted a single time, provided its ID remains unchanged. Updates to a child session will be merged on the client, but not passed back up until either a crash and re-mount or a connection drop and recovery. To force a child to re-mount with new session data, a new ID must be provided.

If that's true, I think we should make it required only if we can check if there's a parent, i.e. it's not a root LV. Otherwise, we should just leave it optional.

lnr0626: Won't it always be a child live view when our validation is run? I don't think we do any validation on the assigns when a `Sirfave.LiveView` is used from the router or a live_render call

Is your concern around the documentation we generate? Or validation that's too strict?

msaraiva: > Won't it always be a child live view when our validation is run? I don't think we do any validation on the assigns when a `Sirfave.LiveView` is used from the router or a live_render call

You're probably right. If that's true, then my concern is unfounded. Feel free to merge it :)

Issue #32, opened 2020-02-12
Add built-in form components


msaraiva:
Create built-in context friendly form components around Phoenix helpers.  

### List of Phoenix built-in form helpers

- [x] color_input (thanks to @alexandrubagu)
- [x] date_input (thanks to @alexandrubagu)
- [x] date_select (thanks to @alexandrubagu)
- [x] datetime_local_input (thanks to @alexandrubagu)
- [x] email_input (thanks to @alexandrubagu)
- [x] form_for (thanks to @zamith)
- [x] hidden_input (thanks to @alexandrubagu)
- [x] label
- [x] number_input (thanks to @alexandrubagu)
- [x] password_input (thanks to @alexandrubagu)
- [x] radio_button (thanks to @alexandrubagu)
- [x] range_input (thanks to @alexandrubagu)
- [x] reset (thanks to @alexandrubagu)
- [x] search_input (thanks to @alexandrubagu)
- [x] submit (thanks to @zamith)
- [x] telephone_input (thanks to @alexandrubagu)
- [x] text_input (thanks to @zamith)
- [x] textarea (thanks to @alexandrubagu)
- [x] time_input (thanks to @alexandrubagu)
- [x] url_input (thanks to @alexandrubagu)
- [x] checkbox (thanks to @miguel-s)
- [x] select (thanks to @miguel-s)
- [x] inputs_for (implemented as `<Inputs for={{...}} .../>`)
- [x] datetime_select (thanks to @miguel-s)
- [x] time_select (thanks to @miguel-s)
- [x] file_input (thanks to @miguel-s)
- [x] hidden_inputs_for (thanks to @miguel-s)
- [x] multiple_select (thanks to @miguel-s)
- [x] options_for_select (thanks to @miguel-s)

I'm not sure yet if it makes sense to create wrappers for all of them. Please let me know if you believe we should add/remove any helper to/from this list.

zamith:
Is the idea to create built-in components like `<TextField>` or `<Form>` that then wrap the existing components or something lower level than that?

msaraiva:
Hi @zamith!

So far, the only one we've found that would require a lower level implementation would be `inputs_for` as discussed [here](https://github.com/msaraiva/surface/issues/23#issuecomment-583168194). All the rest could, apparently, be just wrappers. 

Regarding `inputs_for`, in case we can't find a practical solution, we can still create a wrapper and use slot props to pass the form instance to the parent scope, instead of retrieving it from the context.

[...]

PR #137, opened 2020-09-11
Inputs component

msaraiva:
A wrapper for `inputs_for/3`.

Related to #32.

### Usage with context

```jsx
<Form for={{ :parent }}>
  <Inputs for={{ :children }}>
    <TextInput field="name" />
    <TextInput field="email" />
  </Inputs>
</Form>
```

### Usage with slot props

```jsx
<Form for={{ :parent }}>
  <Inputs for={{ :children }} :let={{ form: f }}>
    <TextInput form={{ @f }} field="name" />
    <TextInput form={{ @f }} field="email" />
  </Inputs>
</Form>
```

### DIFF
--- a/lib/surface/compiler/eex_engine.ex
+++ b/lib/surface/compiler/eex_engine.ex
@@ -123,7 +123,7 @@ defmodule Surface.Compiler.EExEngine do
          %AST.Slot{
            name: name,
            index: index_ast,
-           props: %AST.Directive{value: %AST.AttributeExpr{value: props_expr}},
+           props: props_expr,
            default: default
          },
          buffer,
@@ -390,7 +390,7 @@ defmodule Surface.Compiler.EExEngine do
          [
            %AST.Template{
              name: name,
-             let: %AST.Directive{value: %AST.AttributeExpr{value: let}},
+             let: let,
              children: children
            }
            | tail
@@ -411,7 +411,7 @@ defmodule Surface.Compiler.EExEngine do
            %AST.SlotableComponent{
              slot: name,
              module: module,
-             let: %AST.Directive{value: %AST.AttributeExpr{value: let}},
+             let: let,
              props: props,
              templates: %{default: default}
            }
--- a/lib/surface/compiler/eex_engine.ex
+++ b/lib/surface/compiler/eex_engine.ex
@@ -339,10 +339,7 @@ defmodule Surface.Compiler.EExEngine do
 
     slot_meta =
       for {name, size, infos} <- slot_info do
-        prop_assign_mappings =
-          Enum.map(infos, fn {let, _, _} ->
-            Enum.map(let, fn {prop_name, binding_name} -> {prop_name, binding_name} end)
-          end)
+        prop_assign_mappings = Enum.map(infos, fn {let, _, _} -> let end)
 
         meta_value =
           quote generated: true do
--- a/lib/surface/compiler/eex_engine.ex
+++ b/lib/surface/compiler/eex_engine.ex
@@ -341,7 +341,7 @@ defmodule Surface.Compiler.EExEngine do
       for {name, size, infos} <- slot_info do
         prop_assign_mappings =
           Enum.map(infos, fn {let, _, _} ->
-            Enum.map(let, fn {prop_name, {binding_name, _, _}} -> {prop_name, binding_name} end)
+            Enum.map(let, fn {prop_name, binding_name} -> {prop_name, binding_name} end)
           end)
 
         meta_value =
@@ -444,7 +444,7 @@ defmodule Surface.Compiler.EExEngine do
     end)
     |> Enum.map(fn %{generator: gen, name: name} ->
       case find_attribute_value(props, gen, nil) do
-        %AST.AttributeExpr{value: {binding, _}} ->
+        %AST.AttributeExpr{value: {{binding, _, _}, _}} ->
           {name, binding}
 
         _ ->
@@ -601,11 +601,13 @@ defmodule Surface.Compiler.EExEngine do
                         let: %AST.Directive{
                           module: Surface.Directive.Let,
                           name: :let,
-                          value: %AST.AttributeExpr{
-                            original: "",
-                            value: [],
-                            meta: meta
-                          },
+                          value: [
+                            %AST.AttributeExpr{
+                              original: "",
+                              value: [],
+                              meta: meta
+                            }
+                          ],
                           meta: meta
                         },
                         children: [require_expression],
--- a/lib/surface/components/form/inputs.ex
+++ b/lib/surface/components/form/inputs.ex
@@ -35,11 +35,11 @@ defmodule Surface.Components.Form.Inputs do
 
   def render(assigns) do
     ~H"""
-    <div :for={{ f <- inputs_for(get_form(assigns), @for, @opts) }}>
-      <Context set={{ :form, f, scope: Surface.Components.Form }}>
-        <slot :props={{ form: f }}/>
-      </Context>
-    </div>
+    <Context
+      :for={{ f <- inputs_for(get_form(assigns), @for, @opts) }}
+      set={{ :form, f, scope: Surface.Components.Form }}>
+      <slot :props={{ form: f }}/>
+    </Context>
     """
   end
 end
--- a/lib/surface/components/form/inputs.ex
+++ b/lib/surface/components/form/inputs.ex
@@ -1,0 +1,45 @@
+defmodule Surface.Components.Form.Inputs do
+  @moduledoc """
+  A wrapper for `Phoenix.HTML.Form.html.inputs_for/3`.
+
+  Additionally, adds the generated form instance that is returned by `inputs_for/3`
+  into the context, making it available to any child input.
+  """
+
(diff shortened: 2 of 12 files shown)


lib/surface/components/form/inputs.ex
```diff
@@ -0,0 +1,45 @@
...
+  See `Phoenix.HTML.Form.html.inputs_for/4` for the available options.
+  """
+  property opts, :keyword, default: []
+
+  @doc "The code containing the input controls"
+  slot default, props: [:form]
+
+  def render(assigns) do
+    ~H"""
+    <div :for={{ f <- inputs_for(get_form(assigns), @for, @opts) }}>
```
msaraiva, line 38: I had to use a `<div>` here because currently, there's no way to perform a `:for` without a tag. This is the first case I found where we need this. We need to find a nice way to achieve that. Here's a couple of options:

1. Use `<template :for={{...}}>...</template>` - This is similar to `Vue`'s approach. I believe this will require changes in the compiler/eex_engine to suppress the wrapping `<template>...</template>`.
1. Add a new component `<For>` or `<ForEach>`
1. Add a new directive to suppress the wrapping element. Something like: `<div :for={{}} :unwrap>`

test/components/form/inputs_test.exs
```diff
@@ -0,0 +1,69 @@
...
+  alias Surface.Components.Form, warn: false
+  alias Surface.Components.Form.Inputs, warn: false
+  alias Surface.Components.Form.TextInput, warn: false
+
+  import ComponentTestHelper
+
+  test "using generated form received as slot props" do
+    code = """
+    <Form for={{ :parent }} opts={{ csrf_token: "test" }}>
+      <Inputs for={{ :children }} :let={{ form: f }}>
```
msaraiva, line 13: I wonder if users might be confused between `for` and `:for`.

msaraiva: For the `<Inputs>` component we could just call it `field` but for the `<Form>` component I'm not sure. Should we maybe use separate properties? Like: 

```jsx
<Form changeset={{ @changeset }}>
and
<Form name={{ :my_form }}>
```

msaraiva: A separate discussion is whether we should use `<FormFor>` and `<InputsFor>` instead. 

Although using `<FormFor>` and `<InputsFor>` would be more intuitive for people already using Phoenix. I find them much less intuitive for other users. Besides, since we need to have a property like `for` to pass the value (as opposed to just pass it as an unnamed argument like in `form_for`), it seems to make sense to use just `<Form>`.

lib/surface/components/form/inputs.ex
lnr0626: The `transform` functionality would allow creating a `For`/`ForEach` component fairly trivially - that's what I had in mind for exploring the `case`/`cond` support

lnr0626: `:for` would work on the `Context` component directly:

```
<Context
  :for={{ f <- inputs_for(get_form(assigns), @for, @opts) }}
  set={{ :form, f, scope: Surface.Components.Form }}>
    <slot :props={{ form: f }} />
</Context>
```

test/components/form/inputs_test.exs
lnr0626: > I wonder if users might be confused between `for` and `:for`.

I could see that, though I think the context for using and describing the two is different enough that it's reasonable - i.e. `for` makes sense as a property name here, and that documentation is associated with the `Inputs` and `Form` components

> Should we maybe use separate properties?

🤔 if we were using separate properties, that would open users up to potentially specifying two different values for the same input. (i.e. `<Form name={{ :my_form }} changeset={{ @changeset }}>`). I prefer a single property so we don't even have to consider how to validate that.

> using `<FormFor>` and `<InputsFor>` would be more intuitive for people already using Phoenix

maybe more familiar or a better `1:1` mapping, but i would say that `<Form for={{ @changeset }}>` reads a bit better than `<FormFor for={{ @changeset }}>`; and if we stick with the property named `for`, keeps that fairly close mapping. imo the reason `form_for` works is because the first argument is the thing the form is for, so both read the same (at least to me)

lib/surface/components/form/inputs.ex
msaraiva: > The `transform` functionality would allow creating a `For`/`ForEach` component fairly trivially - that's what I had in mind for exploring the `case`/`cond` support

So you think we should go with `#2`. Right?

lnr0626: Yeah, I think #2 would be the way to go

msaraiva:
There's one last thing we need to decide before I can merge this, which is change the way `:let` maps the assigns. Currently, it works like this:

```jsx
<template slot="body" :let={{ item: user, info: my_info }}>
  User: {{ @user }}
  Info: {{ @my_info }}
</template>
```

The problem is that it gives the impression we have a pattern matching and `my_info` and `user` will be both variables. I think this is misleading since they will be assigns instead of variables. I see two alternatives:

### Option 1 - Replace the variable with an atom.

```jsx
<template slot="body" :let={{ item: :user, info: :my_info }}>
  User: {{ @user }}
  Info: {{ @my_info }}
</template>
```

### Option 2 - Use the generator like syntax

```jsx
<template slot="body" :let={{ @user <- item, @my_info <- info }}>
  User: {{ @user }}
  Info: {{ @my_info }}
</template>
```

Any other alternative?

lnr0626:
At first glance I like the third alternative the best; followed by the first. I think the second would be require us to add more validations because the generator syntax in elixir allows more things than we want to allow in `:let`

With the third option, we could also make `as` optional, which would mean to just use that prop name instead of overriding it. It also has a side benefit of allowing us to extend let down the road

I believe the following would also be valid, but haven't verified if it works correctly:

```
<template ... :let={{ item: @user, info: @my_info }}>
...
```

The reason I think this'll work correctly is because we're treating it like a keyword list, and you can use module attributes for the values of those keywords.

msaraiva:
I think using `:let={{ item: @user, info: @my_info }}` would be confusing since it looks like we're passing assigns instead of defining them. It seems to me that `#3` is our best option. The main benefit IMO, is that it uses the same syntax of `Context`'s `get`, which makes its use very intuitive.

lnr0626:
> I still need to check how to do the accumulate thing on directives though

`collect_directives` will extract all instances of `:let` by default, but won't try to accumulate them. We should be able to handle accumulating all of them in the [template_props](https://github.com/msaraiva/surface/blob/master/lib/surface/compiler.ex#L441-L456) function instead of returning the first one we find

msaraiva:
One case that might get confusing now is this one:

```jsx
    <Grid items={{ user <- @items }}>
      <Column>
        <b>Id: {{ @user.id }}</b>
      </Column>
    </Grid>
```

This one is tricky because it creates the variable `user` but also implicitly generates an assign `@user`. The value is passed by the parent `Grid` to the `Column` using `:props`:

```
<slot name="cols" index={{ index }} :props={{ item: item }}/>
```

 and then the value is implicitly mapped to the `@user` assign as:

```
<Column :let={{ :item, as: :user }}>
```

The problem is that the user won't see that `:let` above.

@lnr0626 I'm not sure if we can do any better than that regarding the syntax since it must be a valid generator, i.e. it accepts filters and might accept multiple clauses in the future. Let me know in case you have any better idea. Otherwise, I think we'll have to keep the syntax as it is and try to do a good job in the docs.

Not being able to pass slot props as variables because of change tracking issues sucks. The previous syntax was not only gorgeous but much more consistent :(

lnr0626:
i can't think of a way to keep the generator capabilities and provide a syntax that more intuitively links the assign with the binding name ☹️

msaraiva:
Ok. The syntax has been updated. It accepts both:

```jsx
<Column :let={{ :item }}>
```

and

```jsx
<Column :let={{ :item, as: :user }}>
```

@lnr0626 I'm not fully satisfied with the changes necessary to do the `accumulate` since it will only work for `:let`. However, that was the best approach to achieve what we needed with minimal changes. Ideally, we should have something similar to what we have for `property` using the `accumulate: true`.

lib/surface/compiler/eex_engine.ex
```diff
@@ -341,7 +341,7 @@ defmodule Surface.Compiler.EExEngine do
       for {name, size, infos} <- slot_info do
         prop_assign_mappings =
           Enum.map(infos, fn {let, _, _} ->
-            Enum.map(let, fn {prop_name, {binding_name, _, _}} -> {prop_name, binding_name} end)
+            Enum.map(let, fn {prop_name, binding_name} -> {prop_name, binding_name} end)
```
lnr0626, line 344: if this is already a keyword list of prop -> binding then we don't need this inner map anymore

lnr0626 (approved):
I see what you mean, though I think it'd be worthwhile to see more use cases where something like `accumulate` would be useful. I'm going to give something a try and see how it feels

lib/surface/compiler/eex_engine.ex
msaraiva: 🤦‍♂️

PR #138, opened 2020-09-12
Add index and with_index modifiers to :for

msaraiva:
This PR adds initial support for modifiers. Two modifiers were added to the `:for` directive, which are `with_index` and `index`.

### .index

```jsx
<td :for={{ {_col, index} <- Enum.with_index(@cols) }}>
  <slot name="cols" index={{ index }}/>
</td>
``` 

can be written as:

```jsx
<td :for.index={{ @cols }}>
  <slot name="cols" index={{ index }}/>
</td>
``` 

or

```jsx
<td :for.index={{ i <- @cols }}>
  <slot name="cols" index={{ i }}/>
</td>
``` 

### .with_index

```jsx
<td :for={{ {item, index} <- Enum.with_index(@items) }}>
  Title: {{ item.title }}
  Content: <slot name="cols" index={{ index }}/>
</td>
```
can be written as:

```jsx
<td :for.with_index={{ {item, index} <- @items }}>
  Title: {{ item.title }}
  Content: <slot name="cols" index={{ index }}/>
</td>
```

## Using multiple modifiers

Currently we only have `index` and `with_index` but in the near future we could add a few more. This way we could use them together, for instance:

```jsx
<td :for.compact.uniq.with_index={{ {item, index} <- @items }}>
  Title: {{ item.title }}
  Content: <slot name="cols" index={{ index }}/>
</td>
```

would be equivalent to:

```jsx
<td :for={{ {item, index} <- @items |> Enum.reject(&is_nil/1) |> Enum.uniq() |> Enum.with_index() }}>
  Title: {{ item.title }}
  Content: <slot name="cols" index={{ index }}/>
</td>
```

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -586,6 +586,8 @@ defmodule Surface.Compiler do
   defp collect_directives(handlers, [attr | attributes], meta) do
     {:ok, dirs, attrs} = collect_directives(handlers, attributes, meta)
 
+    attr = extract_modifiers(attr)
+
     directives =
       handlers
       |> Enum.map(fn handler -> handler.extract(attr, meta) end)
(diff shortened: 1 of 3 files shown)


lnr0626 (approved):
This looks good!

PR #139, opened 2020-09-15
treat let and slot props the same as other directives

lnr0626:
This removes a bit of special handling for these directives in the compiler. WDYT?


msaraiva:
I like it :)

PR #143, opened 2020-09-28
Update slot props and contexts to use variables

msaraiva:


msaraiva:
@lnr0626 I still need to implement the last changes to the `<Context>` API, which are renaming `set` to `put`, `set` to `use` and add the variable syntax to them. I'll work on that later. In the meantime, if you want to start taking a look at the changes, feel free to jump in. Pay attention that since there's no `ContentHandler` anymore, there might be more code to be removed.

lib/surface/compiler/eex_engine.ex
```diff
@@ -219,16 +219,41 @@ defmodule Surface.Compiler.EExEngine do
 
     dynamic_props_expr = handle_dynamic_props(dynamic_props)
 
-    {case_expression, slot_meta, slot_props} =
-      collect_slot_meta(component, templates, buffer, state)
+    if module.__use_context__? do
+      Module.put_attribute(meta.caller.module, :use_context?, true)
```
lnr0626, line 223: we should probably move this attribute setting into the compiler

lib/surface/components/form/input.ex
```diff
@@ -41,4 +40,23 @@ defmodule Surface.Components.Form.Input do
       @default_class get_config(:default_class) || get_config(unquote(__MODULE__), :default_class)
     end
   end
+
+  defmodule InputContext do
```
lnr0626, line 44: 👍

lib/surface/components/form/hidden_input.ex
```diff
@@ -21,13 +21,13 @@ defmodule Surface.Components.Form.HiddenInput do
...
 
   def render(assigns) do
-    form = get_form(assigns)
-    field = get_field(assigns)
     props = get_non_nil_props(assigns, [:value, :class])
     event_opts = get_events_to_opts(assigns)
 
     ~H"""
-    {{ hidden_input(form, field, props ++ @opts ++ event_opts) }}
+    <InputContext assigns={{ assigns }} :let={{ form: form, field: field }}>
```
lnr0626, line 28: why the full assigns instead of a prop for form and a prop for field?

lib/surface/compiler/eex_engine.ex
```diff
@@ -219,16 +219,41 @@ defmodule Surface.Compiler.EExEngine do
...
+            Map.merge(@__context__ || %{}, the_context)
+          end
+
+        true ->
+          quote generated: true do
+            @__context__ || %{}
+          end
+      end
+
+    {gets, props_expr} = Keyword.pop(props_expr, :get, [])
```
lnr0626, line 244: would this effect any component with a prop named `get`?

lib/surface/compiler/eex_engine.ex
msaraiva: Besides this one, there are other things related to contexts that we need to decide the best place to put them. With the current implementation, those things are spread along `Context`, `EExEngine`, `ContextPut` and `ContextGet`. Since we now have everything working, we can start thinking about refactoring the whole thing. Maybe for `v0.2.0`.

lib/surface/components/form/hidden_input.ex
msaraiva: Well, since we don't have to worry about change tracking for form components and this is kind of a "private" component, I think making the whole thing available would be more flexible so we can add new features easily in the future.

lib/surface/compiler/eex_engine.ex
msaraiva: Good catch! I'll restrict to `Context`.

msaraiva:
@lnr0626 after updating the context's  API to use variables and playing around with `use`, I had the feeling it would be more intuitive to just keep `get`. So I changed `set` to `put` but left `get` unchanged.

lib/surface/compiler.ex
```diff
@@ -676,7 +676,8 @@ defmodule Surface.Compiler do
             `slot #{slot_name}, props: [..., #{inspect(prop)}]`\
             """
 
-            IOHelper.compile_error(message, prop_meta.file, prop_meta.line)
+            line = prop_meta.line + prop_meta.line_offset - 1
```
msaraiva, line 679: @lnr0626 although this fixes the problem, I'm not sure it's the best way to handle it. Isn't `prop_meta.line` already supposed to be the line since the beginning of the file?

Currently, using `live_render` in tests hides similar issues as it creates a dynamic view in a single line. I don't remember exactly why it was implemented like that but it was probably because before the new compiler, we didn't have much control over the lines. I think now we'll be able to try a different approach. We can do that on a separate PR.

lnr0626:
There's definitely additional code we can remove - the first that comes to mind is the `validate_interpolation` code in `Surface.Compiler.Helpers` - that was intended to validate users were calling `@inner_content` correctly; however with the `render_live` change they no longer call it as a function. I'm sure there's other code, but nothing else jumps to mind right now

lib/surface/compiler.ex
msaraiva: Sorry for the lack of context. Without this change, the error points to the wrong line. Instead of line 2 in the template, it points to line 2 in the whole file:

image
 
As I said, I'm not sure this is the right way the fix it because we need to make sure the value of `prop_meta.line` is being calculated correctly. Otherwise, we might be just hiding an existing bug.

lib/surface/components/context.ex
```diff
@@ -6,134 +6,148 @@ defmodule Surface.Components.Context do
...
-    get={{ :field, scope: Field, as: :my_field }}>
-    <MyTextInput form={{ @form }} field={{ @my_field }} />
+    get={{ Form, form: form }}
+    get={{ Field, field: my_field }}>
+    <MyTextInput form={{ form }} field={{ my_field }} />
   </Context>
   ```
   """
-  property get, :context_get, accumulate: true, default: []
+  property get, :context_get, accumulate: true, to_assign: false, default: []
```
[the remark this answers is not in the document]
msaraiva: I don't think so. I'll remove it.

lib/surface/compiler.ex
lnr0626: good catch - there was a bug with let and debug where it wasn't actually calculating the line numbers correctly

msaraiva: It's just great you could find the root of the problem. Thanks!

I've been playing around with some examples, trying to simulate some common mistakes the user can make so we can make sure we have helpful error/warning messages for most situations.

PR #144, opened 2020-09-28
Select component

miguel-s:
A wrapper for `select/4`.

Related to #32.

### Example
```Elixir
<Select form="user" field="role" options={{ ["Admin": "admin", "User": "user"] }} />
```

### Example with context
```Elixir
<Form for={{ :form }}>
  <Field name={{ :role }}>
    <Select options={{ ["Admin": "admin", "User": "user"] }} />
  </Field>
</Form>
```

PR #145, opened 2020-09-29
MultipleSelect component

miguel-s:
A wrapper for `multiple_select/4`.

Related to #32.

### Example
```Elixir
<MultipleSelect form="user" field="roles" options={{ ["Admin": "admin", "User": "user"] }} />
```

### Example with context
```Elixir
<Form for={{ :form }}>
  <Field name={{ :roles }}>
    <MultipleSelect options={{ ["Admin": "admin", "User": "user"] }} />
  </Field>
</Form>
```

### DIFF
--- a/lib/surface/components/form/multiple_select.ex
+++ b/lib/surface/components/form/multiple_select.ex
@@ -23,7 +23,7 @@ defmodule Surface.Components.Form.MultipleSelect do
   property class, :css_class
 
   @doc "The options in the select"
-  property options, :keyword, default: []
+  property options, :any, default: []
 
   @doc "Options list"
   property opts, :keyword, default: []
--- a/lib/surface/components/form/multiple_select.ex
+++ b/lib/surface/components/form/multiple_select.ex
@@ -1,0 +1,45 @@
+defmodule Surface.Components.Form.MultipleSelect do
+  @moduledoc """
+  Defines a select.
+
+  Provides a wrapper for Phoenix.HTML.Form's `multiple_select/4` function.
+
+  All options passed via `opts` will be sent to `multiple_select/4`, `class` can
+  be set directly and will override anything in `opts`.
+  """
+
+  use Surface.Component
+
+  import Phoenix.HTML.Form, only: [multiple_select: 4]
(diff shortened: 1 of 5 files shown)


miguel-s:
Hi @msaraiva!

I'm running into a strange issue here and in my previous PR #144.

The property `options` is now defined with type `:keyword`, however `select/4` and `multiple_select/4` accept many kinds of Enumerables. So I tried changing the type of this field to `:any` and I'm getting this error:

`** (RuntimeError) invalid value for property "options". Expected a :any, got: [Admin: "admin", User: "user"].`

Do you know what could be causing this?

Thanks!

msaraiva:
Hi @miguel-s!

Can you try changing this code: https://github.com/msaraiva/surface/blob/710f8025d56fac7f2cf4782bfed736ac0e6e0f37/lib/surface/type_handler/default.ex#L28-L31

to:   
```elixir
  @impl true
  def expr_to_value([value], []) do
    {:ok, value}
  end

  def expr_to_value([], opts) do
    {:ok, opts}
  end
```

This way the default type handler can properly handle any value containing either a single clause or a keyword list (opts).

Let me know if that works :)

miguel-s:
Yep, that did the trick!
Now I just need to actually understand what happened ;)

msaraiva:
Basically, in order to support syntactic sugar, type handlers receive not one a single value but two different sets of values, clauses and opts. For instance, the `:css_class` type accepts the following syntax:

```jsx
<button class={{ :btn, :rounded, active: @active, disabled: @disabled }}>
```
which has two clauses: `[:btn, :rounded]` and two opts: `[active: @active, disabled: @disabled]`. 

Another example is the `:for` directive which can receive something like:

```jsx
<button :for={{ item <- @items, item != nil }}>
```

which contains two clauses and no opts.

In both cases, the expressions passed are not valid Elixir expressions in a sense that they cannot be directly assigned as a single value so they need to be translated by a type handler, which receives those values (clauses + opts), validate and convert them into a single internal representation.

msaraiva:
@miguel-s thanks!

PR #149, opened 2020-10-02
utilize slot props to define context variables

lnr0626:
After looking at your current PR, I thought we might be able to take advantage of the slop props to define context variables instead of injecting `var = value` statements. Some of the code in the context class would need to be cleaned up, but the basic idea does work.

WDYT?

### DIFF
--- a/lib/surface/compiler/eex_engine.ex
+++ b/lib/surface/compiler/eex_engine.ex
@@ -215,11 +215,7 @@ defmodule Surface.Compiler.EExEngine do
          state
        )
        when ast_type in [AST.Component, AST.SlotableComponent] do
-    {props, non_assign_props} =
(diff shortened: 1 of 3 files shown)


msaraiva:
@lnr0626 I'm not sure. Would that play nice with the recent changes I made? If you think it does, can you please rebase it? I'm calling it a night but I can take a look at it first thing in the morning. Thanks.

lnr0626:
just rebased against your branch, the changes to get/put made the code a bit simpler too and allowed me to get rid of the call to `Code.eval_quoted`

msaraiva:
I like it! It might be trickier to accept more complex patterns in the future as we'll not be able to individualize error messages per key (any unmatched pattern invalidates the whole expression). But since we may never accept that, let's go for it :)

PR #152, opened 2020-10-03
fix directives using Map.merge instead of Helpers.to_meta

lnr0626:
A few of the directives were just calling `Map.merge` instead of `Helpers.to_meta` which handles updating the line number based on the offset


msaraiva:
Awesome! 🎉

PR #155, opened 2020-10-06
DateTimeSelect component

miguel-s:
A wrapper for `datetime_select/3`.

Related to #32.

### Example
```Elixir
<DateTimeSelect form="user" field="born_at" />
```

### Example with context
```Elixir
<Form for={{ :user }}>
  <DateTimeSelect field={{ :born_at }} />
</Form>
```

### DIFF
--- a/lib/surface/components/form/datetime_select.ex
+++ b/lib/surface/components/form/datetime_select.ex
@@ -4,8 +4,8 @@ defmodule Surface.Components.Form.DateTimeSelect do
 
   Provides a wrapper for Phoenix.HTML.Form's `datetime_select/3` function.
 
-  All options passed via `opts` will be sent to `datetime_select/3`, `value` and
-  `class` can be set directly and will override anything in `opts`.
+  All options passed via `opts` will be sent to `datetime_select/3`, `value`
+  can be set directly and will override anything in `opts`.
 
 
   ## Examples
--- a/lib/surface/components/form/datetime_select.ex
+++ b/lib/surface/components/form/datetime_select.ex
@@ -14,7 +14,7 @@ defmodule Surface.Components.Form.DateTimeSelect do
   <DateTimeSelect form="user" field="born_at" />
 
   <Form for={{ :user }}>
-    <FileInput field={{ :born_at }} />
+    <DateTimeSelect field={{ :born_at }} />
   </Form>
   ```
   """
--- a/lib/surface/components/form/datetime_select.ex
+++ b/lib/surface/components/form/datetime_select.ex
@@ -1,0 +1,49 @@
+defmodule Surface.Components.Form.DateTimeSelect do
+  @moduledoc """
+  Generates select tags for datetime.
+
+  Provides a wrapper for Phoenix.HTML.Form's `datetime_select/3` function.
+
+  All options passed via `opts` will be sent to `datetime_select/3`, `value` and
+  `class` can be set directly and will override anything in `opts`.
+
+
+  ## Examples
+
+  ```
+  <DateTimeSelect form="user" field="born_at" />
+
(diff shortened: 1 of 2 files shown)


miguel-s:
Hi @msaraiva!

I'm running into some issues with the `value` prop in this component. 
When trying to set the `value` as either a map:

`<DateTimeSelect form="user" field="born_at" value={{ %{year: 2020, month: 10, day: 9, hour: 2, minute: 11, second: 13} }} />`

 or a tuple:

`<DateTimeSelect form="user" field="born_at" value={{ { {2020, 10, 9}, {2, 11, 13} } }} />`

 I get this error:

`** (RuntimeError) invalid value for attribute "value". Expected a type that implements the String.Chars protocol (e.g. string, boolean, integer, atom, ...), got: {{2020, 10, 9}, {2, 11, 13}}`

I've been trying to track down the cause, unsuccessfully so far... it might be related to `InputContext`, as the `value` prop there has the type `:string` instead of `:any`.

Do you have an idea what could be causing this?

---

Also in the tuple version I have to separate the curly braces

`value={{ { {2020, 10, 9}, {2, 11, 13} } }}` instead of `value={{ {{2020, 10, 9}, {2, 11, 13}} }}`

else I get this error:

`** (Surface.Compiler.ParseError) code:1: expected opening HTML tag`

Seems like the Surface parser doesn't like double curly braces in this situation.

---

You should be able to reproduce these error with the tests, but let me know if I can provide more detail.

Thanks!

msaraiva:
Hi @miguel-s!

> I've been trying to track down the cause, unsuccessfully so far... it might be related to `InputContext`, as the `value` prop there has the type `:string` instead of `:any`.
> 
> Do you have an idea what could be causing this?

The underlying logic of `get_non_nil_props/2` cannot handle all types. We need to fix that. In the meantime, you could try to skip `get_non_nil_props/2` and pass the value directly. Something like:

```elixir
  def render(assigns) do
    ~H"""
    <InputContext assigns={{ assigns }} :let={{ form: form, field: field }}>
      {{ datetime_select(form, field, [value: @value] ++ @opts) }}
    </InputContext>
    """
  end
``` 

> Seems like the Surface parser doesn't like double curly braces in this situation.

Yeah, currently the parser can't handle any double braces inside `{{...}}`.

miguel-s:
Your solution works and will update my PR!

Following your hint on `get_non_nil_props/2` I also spent some time reading and understanding the logic behind it. I got very close to a working solution by adding clauses to `attr_to_opts!/3` and `value_to_html/2` in the default type handler, however the changes in `value_to_html/2` felt quite hacky...

I believe a cleaner way would be to add a new callback called `value_to_opts/2`, which will be used by `attr_to_opts!/3 (similar to `value_to_html/2` with `attr_to_html!/3`)

Happy to give this a try if it sounds good to you :)

msaraiva:
> I believe a cleaner way would be to add a new callback called value_to_opts/2, which will be used by attr_to_opts!/3 (similar to value_to_html/2 with attr_to_html!/3)

Sounds good to me.

@lnr0626 WDTY?

msaraiva:
@miguel-s I believe this was the last wrapper component we needed for form controls. 

Thank you so much for working on this. ❤️

lnr0626:
> I believe a cleaner way would be to add a new callback called value_to_opts/2, which will be used by attr_to_opts!/3 (similar to value_to_html/2 with attr_to_html!/3)

That sounds like a good approach to me

msaraiva:
Alright! 

@miguel-s please go ahead and give it a try and let us know if you need any assistance.

miguel-s:
Great! Thanks for the feedback @msaraiva and @lnr0626.

PR #166, opened 2020-10-19
Generate require to unloaded components

msaraiva:


lib/surface/compiler/eex_engine.ex
```diff
@@ -722,10 +724,10 @@ defmodule Surface.Compiler.EExEngine do
     expr
   end
 
-  defp require_expr(module) do
+  defp require_expr(module, line \\ nil) do
     %AST.Expr{
       value:
-        quote generated: true do
+        quote generated: true, line: line do
```
msaraiva, line 730: Without this, the error points to the `~H"""`, not to where the component is used.

lnr0626 (approved):
This looks good to me, though I'm not quite following why this fixes the issue

lib/surface/compiler.ex
```diff
@@ -311,11 +311,12 @@ defmodule Surface.Compiler do
   defp convert_node_to_ast(:component, {name, attributes, children, node_meta}, compile_meta) do
     # TODO: validate live views vs live components ?
     meta = Helpers.to_meta(node_meta, compile_meta)
+    mod = Helpers.actual_component_module!(name, meta.caller)
+    meta = Map.merge(meta, %{module: mod, node_alias: name})
```
msaraiva, line 315: @lnr0626 this fixes the issue because now, no matter if the child module can be compiled or not, the `meta` var will always contain the component's module name, consequently, this [to_dynamic_nested_html](https://github.com/msaraiva/surface/pull/166/files#diff-96328b37fbe00256a44961d185c56d2b6812ccf64fd5108cbc8cf9b4a1b2f2ecR643) clause will match and the `require TheChildComponent` will exist, creating a compile-time dependency. Without that explicit dependency, the compiler has no idea that it needs to recompile the parent when the problem on the child is fixed.

PR #170, opened 2020-10-21
Remove <div> wrapper from <Field>

paulstatezny:
I'd like to propose that the `<div>` wrapper of the `<Field>` component be removed.

I ran into a use case using a `:first-child` CSS selector where this extra wrapper broke the visual layout I was going for.

Perhaps there's a reason I'm not seeing that would make this `<div>` desired or necessary? If not, I think this would make `<Field>` more flexible. Thanks for considering!

### DIFF
--- a/lib/surface/components/form/field.ex
+++ b/lib/surface/components/form/field.ex
@@ -12,9 +12,6 @@ defmodule Surface.Components.Form.Field do
   @doc "The field name"
   prop name, :atom, required: true
 
-  @doc "The CSS class for the generated `<div>` element"
-  prop class, :css_class
-
   @doc """
   The content for the field
   """
@@ -22,15 +19,9 @@ defmodule Surface.Components.Form.Field do
 
   def render(assigns) do
     ~H"""
-    <div class={{ class_value(@class) }}>
-      <Context put={{ __MODULE__, field: @name }}>
-        <slot/>
-      </Context>
-    </div>
+    <Context put={{ __MODULE__, field: @name }}>
+      <slot/>
+    </Context>
     """
-  end
-
-  defp class_value(class) do
-    class || get_config(:default_class) || ""
   end
 end
--- a/test/components/field_test.exs
+++ b/test/components/field_test.exs
@@ -5,34 +5,6 @@ defmodule Surface.Components.FieldTest do
 
   import ComponentTestHelper
 
-  test "creates a wrapping <div> for the field's content" do
-    code = """
-    <Field name="name">
-      Hi
-    </Field>
-    """
-
-    assert render_live(code) =~ """
-           <div>
-             Hi
-           </div>
-           """
-  end
-
-  test "property class" do
-    code = """
-    <Field name="name" class={{ :field }}>
-      Hi
-    </Field>
-    """
-
-    assert render_live(code) =~ """
-           <div class="field">
-             Hi
-           </div>
-           """
-  end
-
   test "sets the provided field into the context" do
     code = """
     <Field name="my_field">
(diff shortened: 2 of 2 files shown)


paulstatezny:
P.S. I think those tests I removed aren't applicable any more; didn't just remove them to make it pass CI. 😄

msaraiva:
Hi @paulstatezny!

I'd like to extend the discussion a bit on this one.

I agree that we should have a way to add a field into the context without making any assumption about the design, even if it's just adding a simple `<div>`. So let's see if we can find a better solution.

Most of the time, for each field in a form, we have additional elements, like a `label` and something to present the error. Usually, those elements are grouped together to make it easier to style them. That's why most of the CSS libs I know provide a class for it. For instance, Bootstrap has `form-group` and Bulma has `field`. Adding the `<div>` is handy because it makes the code cleaner as the user doesn't have to inject any extra markup in every single `<Field>`. In case you're not grouping them, can you provide an example on how you arrange them?

@lnr0626 do you think we should make the component flexible in a way that allows the user to choose which approach is desired? For instance, we could add a property `unwrap` to remove the outer `<div>` as we did for `Markdown`. The user could even define that globally to avoid setting this on every `<Field>` and the default could be to have the field unwrapped as it's design-agnostic.

lnr0626:
If we wanted to have both options, I'd rather have two components instead of adding options to this component - because elixir allows renaming aliases we can just tell people to `alias Surface.UnwrapedField/DivWrapedField, as: Field` (or w/e we'd end up deciding to call it). I do think we should find a way to support both - the div wrapper is really handy, and it's more flexible if we have an option to not include that wrapper.

I've only really used bootstrap and more recently tailwind (and some of the tailwind ui components), and those both wrap each input in a div. I'm also interested in seeing an example of laying out a field without the divs wrapping each input.

msaraiva:
I like the idea of having different components 👍

Assuming we confirm that in most scenarios the user wants the wrapper div. I think we should keep `<Field>` with the current behaviour and create a new component called `FieldContext`, which describes better what it does. It also makes it more intuitive as we already have a [InputContext](https://github.com/msaraiva/surface/blob/7c2a1ed5c46c790f057291f6476aff15d42e902c/lib/surface/components/form/input.ex#L44) that does something similar. This way `Field` could use the `FieldContext` just like the input controls use the `InputContext`.

paulstatezny:
@msaraiva @lnr0626 Thanks for your feedback! That totally makes sense.

## My Scenario (since you asked 😁)

So I was building out a component for building forms, taking a markup structure from [TailwindUI](https://tailwindui.com/). It's a grid-based layout where each field has the label in a left column, and inputs in a right column, but on mobile they stack vertically.

Screen Shot 2020-10-22 at 11 51 59 AM

The markup structure around a single field looks something like this: (Redacting attributes since you have to pay to get access 😅)

```html
<div>
  <label>
    LABEL HERE
  </label>
  <div>
    <div>
      <input>
    </div>
  </div>
</div>
```

You might say, the component that defines the outer `<div>` should define a `<Field>` instead. But in my case, this "row" _sometimes_ represents multiple fields, like in a row that contains multiple checkboxes, each backed by a different one. But we still want them grouped in this structure, under this "label".

```html
<div>
  <div>
    LABEL HERE <!-- not an actual <label> this time! -->
  </div>
  <div>
    <div>
      <input id="foo" type="checkbox">
      <label for="foo">Foo</label>
    </div>
    <div>
      <input id="bar" type="checkbox">
      <label for="bar">Bar</label>
    </div>
  </div>
</div>
```

---

I'm glad you agree that it's reasonable for Surface to support cases like this. 😅 Can I go ahead and make another PR implementing your `FieldContext` concept? Or do you need more time to mull it over?

msaraiva:
@paulstatezny thanks for the example.

> Can I go ahead and make another PR implementing your `FieldContext` concept?

Sure! I think it's the way to go as it keeps the most common behaviour as the default but still allows any other case where the user needs to roll its own custom markup.

paulstatezny:
Closing this in favor of https://github.com/msaraiva/surface/pull/172. Thanks!

PR #171, opened 2020-10-22
allow double braces within interpolation

### LINKED ISSUE #156: Parser cannot handle double braces inside interpolation
Currently the parser can't handle double braces inside interpolations `{{...}}`.
E.g. when passing nested tuples as prop value.

#### This works
```Elixir
<DateTimeSelect form="user" field="born_at" value={{ { {2020, 10, 9}, {2, 11, 13} } }} />
```

#### This throws `** (Surface.Compiler.ParseError) code:1: expected opening HTML tag`
```Elixir
<DateTimeSelect form="user" field="born_at" value={{ {{2020, 10, 9}, {2, 11, 13}} }} />
```

lnr0626:
this updates the parser to allow curly braces within interpolation
as long as all of the braces are matched, and ignore any curly braces
inside binaries and charlists.

Fixes #156

### DIFF
--- a/lib/surface/compiler/parser.ex
+++ b/lib/surface/compiler/parser.ex
@@ -136,11 +136,51 @@ defmodule Surface.Compiler.Parser do
   end
 
   ## Regular node
+  defparsecp(
+    :tuple,
+    string("{")
+    |> choice([
+      parsec(:tuple),
+      repeat(utf8_char(not: ?}))
+    ])
+    |> string("}")
+  )
+
+  defparsecp(
+    :binary,
+    string("\"")
+    |> repeat(
+      choice([
+        string("\\\""),
+        utf8_char(not: ?")
+      ])
+    )
+    |> string("\"")
+  )
+
+  defparsecp(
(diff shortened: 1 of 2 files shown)


msaraiva (commented):
That's awesome! I have a couple of questions:

1. I believe this will work for both, tuple and maps, right? Like `{{ %{a: %{b: 1}} }}`. If so, maybe we should add a test for it.
1. Does it also work without the space between the double braces? Like `{{{:a, :b}}}`. If so, I think we should also have a test for it. If not, we can merge this as it is and try to fix that later. I mean, it's already a huge improvement since things like `{{ {:a, {:b, :c}} }}` doesn't work today at all, even with the spaces.

lnr0626:
It should work with any set of matched curly braces - i'll add a test for those two cases to make sure :-)

lnr0626:
this wont work for attribute expressions yet

msaraiva:
Fantastic! ❤️ 

Ship it.

PR #172, opened 2020-10-22
Add <FieldContext>

paulstatezny:
This adds a way to put a field name into context like `Surface.Components.Form.Field` but without wrapping the contents in an extra `<div>`:

```html
<FieldContext name="last_name">
  <TextInput />
</FieldContext>
```

yields:

```html
<input ... />
```

instead of:

```html
<div>
  <input ... />
</div>
```

### DIFF
--- a/lib/surface/components/form/field_context.ex
+++ b/lib/surface/components/form/field_context.ex
@@ -1,0 +1,29 @@
+defmodule Surface.Components.Form.FieldContext do
+  @moduledoc """
+  Defines a context for a form field but without the `<div>` wrapper in `Field`.
+
+  Like the `Field` component, sets the provided field name into the context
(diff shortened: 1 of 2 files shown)


paulstatezny:
@msaraiva Is there anywhere else I should add documentation? I know the docs at http://surface-demo.msaraiva.io/ are in another repository.

msaraiva:
Hey @paulstatezny!

Thanks for the PR! ❤️

> Is there anywhere else I should add documentation? I know the docs at http://surface-demo.msaraiva.io/ are in another repository.

If you think the use of the component might not be trivial to the user and it would be nice to have a more elaborate example on http://surface-demo.msaraiva.io, you can just leave it here as a comment. I'll take a look at all PRs that added new components and search for examples before updating the website.

Thanks!

PR #174, opened 2020-10-23
Improve unloaded module error

### LINKED ISSUE #169: Improve message when a component cannot be loaded
Original discussion can be found in #167.

### Proposal

If the module is not aliased, we should add a hint:

```
warning: cannot render <Form> (module Form could not be loaded)

Hint: Make sure module `Form` can be successfully compiled. 

If the module is namespaced, you can use its full name. For instance:

  <MyProject.Components.Form>

or add a proper alias so you can use just `<Form>`:

  alias MyProject.Components.Form
```

If the module is already aliased, we don't need the hint but we have to show both, the alias and full module's name:

```
warning: cannot render <Form> (module MyProject.Components.Form could not be loaded)
```

msaraiva:
Closes #169.

  * Show full module name
  * Add hint for unaliased modules

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -169,6 +169,11 @@ defmodule Surface.Compiler do
         {:error, {message, line}, meta} ->
           IOHelper.warn(message, compile_meta.caller, fn _ -> line end)
           %AST.Error{message: message, meta: meta}
+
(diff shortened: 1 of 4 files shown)


lnr0626 (approved):
🔥🔥🔥

PR #183, opened 2020-10-29
Expose form instance as slot prop

### LINKED ISSUE #181: Expose form as a slot prop
We should expose the generated form instance so people can pass it to functions or components that are not context-aware. For instance:

```jsx
<Form for={{@changeset}} :let={{ form: f }}> 
  ...
  {{ MyAppWeb.ErrorHelpers.error_tag(f, :my_field) }}
</Form>
```

miguel-s:
Resolves #181.


msaraiva:
@miguel-s thanks! ❤️

PR #184, opened 2020-10-31
add wrapper components for if and for

lnr0626:


lib/surface/components/if.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Surface.Components.If do
+  @moduledoc """
+  Provides an alternative to the `:if` directive for wrapping multiple elements in an if expression.
+
+  ## Examples
+  ```
+  <If condition={{ @display_link }} />
```
msaraiva, line 7: One thing we could try is to add the concept of "root prop" which would be similar to react's spread attribute:

```jsx
<Component {...props} />
```
But in our case, it would be treated just like any other property. Something like:

```elixir
prop condition, :boolean, required: true, root: true
...
prop each, :generator, required: true, root: true
```

Usage:

```jsx
<If {{ whatever == :done }}>
...
</If>

<For {{ item <- @items }}>
...
</For>
```

 I'm not sure yet how hard it would be to adjust the parser to handle that but it seems like a very nice addition, especially when start adding other stuff like `Case` and `Cond`. Naming is hard and I believe we'll end up using generic names like `expression` or `condition`. No need to implement that now, of course, but let me know if you like the idea so we can create an issue for it.

test/components/for_test.exs
```diff
@@ -0,0 +1,26 @@
+defmodule Surface.Components.ForTest do
+  use ExUnit.Case, async: true
+
+  alias Surface.Components.For, warn: false
+
+  import ComponentTestHelper
+
+  describe "Without LiveView" do
```
msaraiva, line 8: I think we don't need that `describe` anymore. I used to use this in some old tests for reasons I don't even remember :)

msaraiva (commented):
I like it ❤️

One thing I noticed while using my `<#For>` was that when there was an error in the content, the compiler was pointing to the wrong line so maybe it's a good idea to add a test to check that.

lib/surface/components/if.ex
lnr0626: I like the idea!

PR #190, opened 2020-11-03
Move constructs to its own namespace

msaraiva:
* Move For and If to Surface.Constructs namespace
  * Create aliases to For and If

@lnr0626 I gave up moving `Context` too. The implementation is quite different and depends on code in the compiler/eex_engine so I left it out for now.

### DIFF
--- a/test/constructs/for_test.exs
+++ b/test/constructs/for_test.exs
@@ -1,8 +1,6 @@
 defmodule Surface.Components.ForTest do
   use ExUnit.Case, async: true
 
-  alias Surface.Components.For, warn: false
-
   import ComponentTestHelper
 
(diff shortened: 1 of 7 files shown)


test/constructs/for_test.exs
```diff
@@ -1,8 +1,6 @@
 defmodule Surface.Components.ForTest do
```
lnr0626, line 1: Should these be moved into `Surface.Constructs` as well?

lnr0626 (commented):
just the one question, other than that looks good to me

test/constructs/for_test.exs
msaraiva: Yup :)

Issue #192, opened 2020-11-03
Data assigns don't work with `assign_new/3`


miguel-s:
Hi!

It seems that declaring `data` assigns does not play well with LiveView's `assign_new/3`.

Surface automatically initializes all `data` assigns with their default value or nil, while `assign_new/3` only checks for the existence of the key in the socket assigns. This results in `assign_new/3` never actually assigning a value to the socket.

You can see a small example of this behaviour when trying to implement this LiveView guide with Surface: https://hexdocs.pm/phoenix_live_view/security-model.html#mounting-considerations

Using Surface the following part will not work and leave `current_user` with the value `nil` (or default value if specified):

```JS
socket = assign_new(socket, :current_user, fn -> Accounts.get_user!(user_id) end)
```

msaraiva:
I believe we have at least two options to fix this:

1. If the `data` assign is declared without a `default` option, we don't initialize it. This will break any existing code that assumes the data will be present.
1. We keep the current behaviour and add a new option to not initialize the value, e.g:
`data current_user, :string, init: false`

@lnr0626 any other ideas?

lnr0626:
Those are the first two thoughts that pop into my head. I'd kind of expected the behaviour to be closer to 1 to work well with `assign_new`

miguel-s:
This is probably a bad idea, but just to mention option 3: provide a `before_init` hook where it's possible to access the socket before Surface initializes data assigns.

msaraiva:
@miguel-s I wonder why you think it's a bad idea. I think option `#1` is probably our best shot as it keeps the default behaviour of live views and components and only when adding `default`, we do something different.

miguel-s:
@msaraiva agree with your assessment of option 1, and those are the same reasons why I think option 3 is not good (different behaviour to LiveView and adds an unexpected workaround)

PR #194, opened 2020-11-04
Only init data assigns with default value

miguel-s:
Related to #192

@msaraiva @lnr0626 decided to try out option 1, what do you think?

### DIFF
--- a/lib/surface/live_view.ex
+++ b/lib/surface/live_view.ex
@@ -63,9 +63,10 @@ defmodule Surface.LiveView do
 
   defp quoted_mount(env) do
     defaults =
-      for %{name: name, opts: opts} <- Module.get_attribute(env.module, :data) do
-        {name, Keyword.get(opts, :default)}
-      end
+      env.module
+      |> Module.get_attribute(:data)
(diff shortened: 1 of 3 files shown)


msaraiva:
@miguel-s I just realised we were working on the same issue at the same time 😬

@lnr0626 since you're reviewing the PRs, you can just pick one and merge 🙂

lib/surface/live_view.ex
```diff
@@ -63,9 +63,10 @@ defmodule Surface.LiveView do
 
   defp quoted_mount(env) do
     defaults =
-      for %{name: name, opts: opts} <- Module.get_attribute(env.module, :data) do
```
feliperenan, line 66: I guess you could keep `for` and just add the condition as second argument in order to avoid `filter` and `map`:

```Elixir
defaults =
  for %{name: name, opts: opts} <- Module.get_attribute(env.module, :data), Keyword.has_key?(opts, :default) do
    {name, Keyword.get(opts, :default)}
  end
  |> Macro.escape()
```

BTW, here is a nice blog post about for comprehensions: https://www.pompecki.com/post/for-comprehensions/

lnr0626:
Thanks for the PR @miguel-s!

Closed via #195

PR #197, opened 2020-11-06
Fix line for string with interpolation

msaraiva:
Errors on properties using string interpolation were pointing to the wrong line.

This PR fixes the issue and adds a proper test.

@lnr0626 before fixing the problem I added following test:

```elixir
test "raise error on the right line for string with interpolation" do
  code =
    quote do
        ~H"""
        <div>
        <StringProp
          label="Undefined func {{ func }}"
        />
        </div>
        """
    end
    
  assert_raise(CompileError, "code:3: undefined function func/0", fn ->
    render_live(code, %{})
  end)
end
```

Which didn't catch the error. It passed flawlessly. 

The test is misleading as it reports the line having the `~H"""` as the offset, which is ok for all cases where the metadata was properly initialized with `Helpers.to_meta`. However, if the metadata is not initialized like that, the tests will never fail. 

I believe we have most of our tests using ` render_live` like this so none of them will ever catch a similar bug.

It's the second time recently we have a similar bug due to a missing `Helpers.to_meta`. We need to either change `render_live` to report the error taking the beginning of the file as offset or somehow raise an error if the metadata has not been initialized. The former is easier to implement but it will make the tests look weird as we'll have to assert on a line that does not correspond the line we see in the test.

Let me know if you have any other idea.

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -474,6 +474,7 @@ defmodule Surface.Compiler do
     {originals, quoted_values} =
       Enum.reduce(values, {[], []}, fn
         {:attribute_expr, value, expr_meta}, {originals, quoted_values} ->
+          expr_meta = Helpers.to_meta(expr_meta, attr_meta)
           {["{{#{value}}}" | originals], [quote_embedded_expr(value, expr_meta) | quoted_values]}
 
         value, {originals, quoted_values} ->
--- a/test/properties_test.exs
+++ b/test/properties_test.exs
@@ -134,7 +134,7 @@ defmodule Surface.PropertiesTest do
         end)
 
       assert output =~ ~r/variable "func" does not exist/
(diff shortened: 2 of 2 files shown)


lnr0626 (approved):
lgtm

> somehow raise an error if the metadata has not been initialized

I think that would be the more robust approach if we can figure it out - we may be able to pattern match on the meta struct where we're using it?

msaraiva:
> we may be able to pattern match on the meta struct where we're using it?

Maybe. The problem at this point is to identify where we must accept only the converted metadata. Currently, we refer to both as just `metadata` or `meta` so we use the same naming variants, e.g. `*_meta` and in most cases, the used fields are also the same, e.g. `line`. I usually find myself wondering what kind of metadata I'm dealing with at the moment in all the different parts of the code, which are many. I believe we should probably start using different names for each one of them so we can easily spot what kind of data (meta) we expect.

PR #199, opened 2020-11-10
<ErrorTag> component

paulstatezny:
Hello! I think it'd be helpful to have an `<ErrorTag>` to go alongside the other components that wrap around `Phoenix.HTML` utilities.

`error_tag` isn't part of `Phoenix.HTML`, but rather something that `mix phx.new` generates into the project. I believe this is because:

1. It ties directly into `MyAppWeb.Gettext`
1. Possibly to allow the markup to be customized?

Still, I think it would useful to ship it with Surface. I solved point 1 with a configuration line. The component will error with instructions if it's missing:

```
** (RuntimeError) Please define a gettext module in config in order to use <ErrorTag>:

config :surface, gettext_module: MyAppWeb.Gettext
```

It would be great to use this as a launchpad for discussion. 🙂


lib/surface/components/form/error_tag.ex
```diff
@@ -0,0 +1,81 @@
...
+  Renders error messages if there are any about the given field.
+
+  ## Examples
+
+  ```
+  <ErrorTag field="password" />
+  ```
+  """
+
+  use Surface.Components.Form.Input
```
paulstatezny, line 15: I'm not 100% sure that we want this line. I think we do want `form`, `field`, and `class` props though. And `InputContext`.

msaraiva: Yup. We don't want that line. It should only have it in components that use `<input ...>`. In this case, it's better to define those properties explicitly.

paulstatezny: Perfect, thanks for the feedback. Fixed that.

paulstatezny:
@msaraiva I believe I've addressed all of the feedback and refactored according to our discussion. Thanks for your help! Let me know if there's anything I missed or any thoughts that come up.

lib/surface/components/form/error_tag.ex
```diff
@@ -0,0 +1,124 @@
...
+  """
+  prop phx_feedback_for, :string
+
+  def render(assigns) do
+    translate_error = assigns.translator || translator_from_config() || (&translate_error/1)
+
+    ~H"""
+    <InputContext assigns={{ assigns }} :let={{ form: form, field: field }}>
+      <span
+        :for={{ error <- Keyword.get_values(form.errors, field) }}
```
msaraiva, line 96: What about the case we discussed about having `<Form for={{ :an_atom }}>`. It will crash with a `function :an_atom.errors/0 is undefined`. We have to handle that, right?

paulstatezny: So there was something about the [prior to this PR] implementation that caused a crash.

But in this case, even if you do `for={{ :atom }}`, we'll still have a form struct. I believe it's `form.data` that is nil in this case. But we're not referencing `form.data.errors`. We're referencing `form.errors`, which will be identical to `form.data.errors` as long as `form.data` is a changeset with `changeset.action != nil`. Otherwise `form.errors` will be `[]`. There is a test in this PR that demonstrates that it doesn't crash.

test/components/form/error_tag_test.exs
```diff
@@ -0,0 +1,166 @@
...
+    code =
+      quote do
+        ~H"""
+        <Form for={{ :user }}>
+          <Field name="name">
+            <ErrorTag />
+          </Field>
+        </Form>
+        """
+      end
```
paulstatezny, line 98: @msaraiva This is the test that demonstrates that using an atom for `for` doesn't crash.

paulstatezny: Would you like me to make the description indicate that it demonstrates that?

test/components/form/error_tag_test.exs
```diff
@@ -0,0 +1,166 @@
...
+             "<span phx-feedback-for=\"test-id\">is already taken</span>"
+  end
+
+  test "prop class", %{changeset: changeset} do
+    assigns = %{changeset: changeset}
+
+    code =
+      quote do
+        ~H"""
+        <Form for={{@changeset}} opts={{ as: :user }}>
```
msaraiva, line 76: We should probably define explicit props for `:as` along with all options listed in https://hexdocs.pm/phoenix_html/Phoenix.HTML.Form.html#form_for/4-options.

This way it would be clear to the user what will become tag attributes and what are the real options (and what they mean). It's a bit more work for us but it gives a much better dev experience. We can try to follow this rule for all wrappers whenever is possible. 

No need to do it in this PR. I'll create an issue for it

test/components/form/error_tag_test.exs
msaraiva: No. It's fine.

test/components/form/error_tag_test.exs
paulstatezny: Love it. I actually got tripped up doing `<Form for={{ ... }} as={{ ... }}>` and was a bit confused when it still complained about not having an `as`.

I think with the `for` prop many will expect `as`.

💯 👍

lib/surface/components/form/error_tag.ex
```diff
@@ -0,0 +1,124 @@
...
+  def translate_error({msg, opts}) do
+    # Because the error messages we show in our forms and APIs
+    # are defined inside Ecto, we need to translate them dynamically.
+    Enum.reduce(opts, msg, fn {key, value}, acc ->
+      String.replace(acc, "%{#{key}}", to_string(value))
+    end)
+  end
+
+  defp translator_from_config do
+    if translator = get_config(:translator) do
```
msaraiva, line 119: One thing I've just noticed that I've been using the `default_` prefix on those kinds of configuration. For instance:

```elixir

  {Surface.Components.Markdown, default_class: "content markdown", default_opts: [smartypants: false]},
  # Bulma defaults
  {Surface.Components.Form.Input, default_class: "input"},
  {Surface.Components.Form.Field, default_class: "field"},
  {Surface.Components.Form.TextArea, default_class: "textarea"},
  {Surface.Components.Form.Label, default_class: "label"}
]
```

The intention was to be more explicit that those values are just defaults since the component author can do whatever he wants with any of those values, e.g. use it as a default, override an existing value, merge the value (like Markdown's opts), etc. 

However, I think this might be a silly concern and I we should probably remove the prefixes and keep everything with the same name. Thoughts?

test/components/form/error_tag_test.exs
msaraiva: Issue #200 created!

lib/surface/components/form/error_tag.ex
paulstatezny: > since the component author can do whatever he wants with any of those values

I think that's the rub. If the component uses it as a default that can be overridden with a prop, personally I think `default_` potentially adds clarity.

But if it merges the config value with prop value, I'm not sure `default_` is the most clear. I can't think of anything at the moment more clear in that instance than removing the prefix and documenting the merge functionality explicitly.

That said, I think `default_translator` would be an improvement here. Shall we move forward with that?

Related question: where is the standard place to document these configuration options? I put it in the `moduledoc` as well as the related `prop` because I couldn't find another relevant place.

lnr0626: I agree re default adding clarity, though I'm not quiet following the distinction between merging and overriding - would that only apply for a keyword list like opts?

msaraiva: > I think that's the rub. If the component uses it as a default that can be overridden with a prop, personally I think `default_` potentially adds clarity.
> 
> But if it merges the config value with prop value, I'm not sure `default_` is the most clear. I can't think of anything at the moment more clear in that instance than removing the prefix and documenting the merge functionality explicitly.

That's exactly why I changed my mind. I think we shouldn't try to have standard prefixes that are semantically consistent across all components. Since each component can do whatever it wants, it should be up to the author to properly document how the config is used.

> where is the standard place to document these configuration options? I put it in the `moduledoc` as well as the related `prop` because I couldn't find another relevant place.

Yeah, `moduledoc` is the right place for now. I thought about adding something like `config` to the API. Like:

```elixir
@doc "Class or classes to apply to each error tag <span>"
prop class, :css_class

@doc "The default class"
config default_class, :string
```

That would make configs first-class citizens but I'm not sure we should do this. Configs should not be abused as it makes the component appearance and behaviour much less intuitive as you change projects. By promoting them to the API, I fear people will start using configs everywhere.

paulstatezny: @msaraiva 100% agree about trying not to encourage usage of config. I see what you're saying.

@lnr0626 I agree, I think it'd typically only be mergeable with composite values like keyword lists.

paulstatezny: Ok, I changed the config key from `translator` to `default_translator`.

test/components/form/error_tag_test.exs
```diff
@@ -0,0 +1,186 @@
...
+      end
+
+    assert render_live(code, assigns) =~
+             "<span phx-feedback-for=\"user_name\">is already taken</span>"
+
+    assert render_live(code, assigns) =~
+             "<span phx-feedback-for=\"user_name\">another test error</span>"
+  end
+
+  test "no errors are shown if changeset.action is empty", %{changeset: changeset} do
```
paulstatezny, line 52: This functionality is probably more closely related to `Phoenix.HTML.Form` than `<ErrorTag>`. But I think it's worth documenting and testing so we can catch regressions.

lib/surface/components/form/error_tag.ex
msaraiva: Fair enough. Let's keep the `default_` prefix then.

msaraiva:
@paulstatezny thanks! ❤️

paulstatezny:
Gladly! Thanks for the help on this one @msaraiva 🤜🤛

Issue #162, opened 2020-10-14
Improving types support


paulstatezny:
First of all, thanks for Surface!

The docs currently say:

> **Note:** Currently, some of the types above work just as annotations and don't have any practical use aside from documentation. If the type you need is not in that list, you can safely use `:any` instead. However, some other types like `:css_class`, `:list` and `:event` are handled differently, i.e. there are extra rules and behaviours applied to them.

Is there a technical or design reason to not validate property types? Could Surface (for example) emit a warning if you supply an invalid type to a component? Or maybe have a "strict mode" config option where it crashes if you supply the wrong type?

I'm not sure if this is technically unrelated, but I would _love_ to be able to supply `values: [:foo, :bar]` and have Surface somehow yell at me if I pass in something other than `:foo` or `:bar`. 🙂

msaraiva:
Hi @paulstatezny!

> Is there a technical or design reason to not validate property types? Could Surface (for example) emit a warning if you supply an invalid type to a component? Or maybe have a "strict mode" config option where it crashes if you supply the wrong type?

We recently added the ability to do some validation at compile-time. Some of them are already validated. However, since Elixir is a dynamic language, there are many cases where there's nothing we can do. Here are some examples of types we can validate at compile-time:

  * The passed value is a literal - See `atom`'s [literal_to_ast_node/4](https://github.com/msaraiva/surface/blob/f1c75f92513b607c98ba578030a647e0f5d6d11c/lib/surface/type_handler/atom.ex#L6-L13).
  * The passed expression can be recognized as the expected type - See `explict_keyword`'s [expr_to_quoted/6](https://github.com/msaraiva/surface/blob/f1c75f92513b607c98ba578030a647e0f5d6d11c/lib/surface/type_handler/explict_keyword.ex#L11-L19).

Any other kind of expression passed can only be validated at runtime. See `styles`'s [expr_to_value/2](https://github.com/msaraiva/surface/blob/f1c75f92513b607c98ba578030a647e0f5d6d11c/lib/surface/type_handler/style.ex#L17-L31).

We've added compile-time validation to some of the types but most of them currently have only runtime validation. PRs are welcome :) 

There are other interesting cases that we could start validating at compile-time, e.g. the passed expression is a declared assign, e.g. `@some_assign`. In this case, we could introspect the component's metadata to retrieve the type of the assign and check if it's compatible with the property's type. We could even try to treat a few other simple expressions but the more we go in this direction, the more complex it will get to maintain all different types.

[...]

PR #204, opened 2020-11-17
Disable validation for required props if :props is passed

msaraiva:
Resolves https://github.com/msaraiva/surface/issues/202.


lnr0626 (approved):
this lgtm, we should revisit when we have a better idea of #162

PR #207, opened 2020-11-18
default_class config option for <ErrorTag>

paulstatezny:
Adds the ability to do this:

```elixir
  {Surface.Components.Form.ErrorTag, default_class: "invalid-feedback"}
]
```

As a result, all `<ErrorTag>` components render:

```html
<span class="invalid-feedback" ...>
```

---

The `class` prop overrides it.

```surface
<ErrorTag class="overriding" />
```

```html
<span class="overriding" ...>
```


msaraiva:
@paulstatezny thanks! ❤️

PR #208, opened 2020-11-19
Fix warning line for undefined slot

msaraiva:


lib/surface/compiler.ex
```diff
@@ -657,9 +657,11 @@ defmodule Surface.Compiler do
       IOHelper.warn(message, meta.caller, fn _ -> meta.line end)
     end
 
-    for slot_name <- names,
-        mod.__get_slot__(slot_name) == nil do
-      missing_slot(mod, slot_name, meta)
+    for {slot_name, template_instances} <- templates,
+        slot_name != :default,
```
msaraiva, line 661: @lnr0626 currently we don't validate the default slot, which makes its declaration optional. Do you think we should keep it that way? It seems to me that forcing the user to declare all slots, including the default one, is more consistent for the user and also simplifies the code as it avoids some conditionals like the one above. 

Personally, I don't mind declaring it. I'll leave it as it is for now but let me know if you have a strong opinion about it.

lnr0626: I don't really have a strong opinion - the only case I can think of where it's used but not declared currently is if it's used to provide some inner content and does not have any props (i.e. similar to Surface.Components.Label's default slot).

Or are you referring to the user providing the default slot, but that slot isn't actually defined/used within the component?

msaraiva: What I mean is whether we should force the user to declare:

```elixir
slot default
```

Currently, I believe we only validate that if the slot is either a named slot or if it defines slot props.

Defining a `<slot/>` automatically adds the `:default` slot to the list but I'm not sure we should keep doing this.

lnr0626: That makes sense - I'm on board with removing that

PR #210, opened 2020-11-20
Update LiveView Dependency

maennchen:
This will also allow a new rc Release on Hex.


msaraiva:
@maennchen thanks! I'll try it out for a couple of days and release `v0.1.0`.

PR #213, opened 2020-11-24
Fix examples in doc

Malian:
This pull request fixes non working example in the dc of the `For` construct component.


msaraiva:
@Malian thanks! ❤️

PR #215, opened 2020-11-27
Added props to Form input elements, fixes #200 

### LINKED ISSUE #200: Define explicit props for all options of the Form wrapper and friends
It would be great to the user if we defined explicit props for all options of a wrapper component.

For instance, for the `<Form>` component, we could have `:as` along with all options listed in https://hexdocs.pm/phoenix_html/Phoenix.HTML.Form.html#form_for/4-options.

This way it would be clear to the user what will become tag attributes and what are the real options (and what they mean). 

It's a bit more work for us but it gives a much better dev experience as the information becomes part of the public API can be used by the editor and any other tool to retrieve the list of props along with the docs (which can be just copied-and-pasted from the the original).

Since the interface of those wrappers is very stable, we don't expect many changes in that regard. We can try to follow this rule for all wrappers **whenever is possible**.

Zurga:
I added all the explicit props to the Form input elements, as requested in #200 . 
The only missing prop is the `value` for the checkbox, this was a bit hard to test to get working. Any input is helpful.

### DIFF
--- a/lib/surface/components/form/checkbox.ex
+++ b/lib/surface/components/form/checkbox.ex
@@ -20,13 +20,30 @@ defmodule Surface.Components.Form.Checkbox do
   import Phoenix.HTML.Form, only: [checkbox: 3]
   import Surface.Components.Form.Utils
 
+  @doc "The value to be sent when the checkbox is checked. Defaults to \"true\""
+  prop checked_value, :any, default: true
+
+  @doc "Controls if this function will generate a hidden input to submit the unchecked value or not, defaults to \"true\"."
+  prop hidden_input, :boolean, default: true
+
+  @doc "The value to be sent when the checkbox is unchecked, defaults to \"false\"."
+  prop unchecked_value, :any, default: false
+
   def render(assigns) do
-    props = get_non_nil_props(assigns, [:checked, class: get_config(:default_class)])
+    props =
+      get_non_nil_props(assigns, [
+        :checked_value,
+        :hidden_input,
+        :unchecked_value,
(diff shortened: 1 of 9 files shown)


msaraiva:
Hey @Zurga. Fantastic job! 

Regarding testing Checkbox's `value` property, something as simple as this should do the trick:

```elixir
  test "setting the value" do
    code =
      quote do
        ~H"""
        <Checkbox value={{ true }}/>
        """
      end

    assert render_live(code) =~ ~r/checked"/

    code =
      quote do
        ~H"""
        <Checkbox value={{ false }}/>
        """
      end

      refute render_live(code) =~ ~r/checked"/
  end
```

lib/surface/components/form/checkbox.ex
```diff
@@ -20,13 +20,30 @@ defmodule Surface.Components.Form.Checkbox do
...
+  prop unchecked_value, :any, default: false
+
   def render(assigns) do
-    props = get_non_nil_props(assigns, [:checked, class: get_config(:default_class)])
+    props =
+      get_non_nil_props(assigns, [
+        :checked_value,
+        :hidden_input,
+        :unchecked_value,
+        :checked,
```
msaraiva, line 38: I think we can remove `:checked` from that list since we'll be using `:value` instead.

Zurga: Thanks for the review! This has been updated in the forked branch.

msaraiva:
@Zurga great job! Thanks again ❤️

PR #217, opened 2020-11-29
Fix doc in the ErrorTag component

Malian:
This Pull Request fixes an typo in the doc of the ErrorTag component.


msaraiva:
@Malian thanks! ❤️

Malian:
@msaraiva I noticed that the same errors are present on the documentation website: http://surface-demo.msaraiva.io/builtincomponents/ErrorTag

Is this repository public?

msaraiva:
@Malian it's not public yet but since it introspects the docs automatically from the modules, it will be updated as soon as I update the Surface version there.

PR #221, opened 2020-12-01
Add DateSelectComponent

Malian:
This PR aims to add a `DateSelectComponent` that is a wrapper for Phoenix.HTML.Form's `date_select/3` function

- [x] Create component
- [x] Test
- [x] Doc
- [x] Parse "class" option of each select to allow the use of `"my-class": true` format
- [ ] ~Maybe update other select component to respect the new standard way to write Surface Form component.~

### DIFF
--- a/lib/surface/components/form/date_select.ex
+++ b/lib/surface/components/form/date_select.ex
@@ -1,0 +1,95 @@
+defmodule Surface.Components.Form.DateSelect do
+  @moduledoc """
+  Generates select tags for date.
+
+  Provides a wrapper for Phoenix.HTML.Form's `date_select/3` function.
+
+  All options passed via `opts` will be sent to `date_select/3`, `value`
+  can be set directly and will override anything in `opts`.
+
+
+  ## Examples
+
+  ```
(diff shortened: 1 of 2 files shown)


Malian:
Ready to review

I am not sure about the way I handle the "css_class" of underlying selects.

I propose to move the "update other select component to respect the new standard way to write Surface Form component" to an other pull request.

msaraiva (commented):
@Malian looks good. I left just a couple of nitpicks for your consideration.

By _"Maybe update other select component to respect the new standard way to write Surface Form component"_ do you mean only the `DateTimeSelect`? If so, yes. We should add `:default`, and all the `year`, `month`, `day`, etc. options to keep them consistent. But it can be certainly done in a separate PR.

Thanks!

Malian:
> We should add `:default`, and all the `year`, `month`, `day`, etc. options to keep them consistent. But it can be certainly done in a separate PR.

Yes, I was referring to `DateTimeSelect` and `TimeSelect`. I'll modify these components in another pull request.

msaraiva:
@Malian thanks! ❤️

PR #222, opened 2020-12-02
Standardize datetime select, date select and time select components

Malian:
This pull request:

- add the following props to the `DateTimeSelectComponent`: default, year, month, day, hour, minute, second and builder to
- add the following props to the `TimeSelectComponent`: default, hour, minute, second and builder to
- reintroduce `opts` props to the `DateSelectComponent`
- add tests to ensure options passed through `opts` are correctly handle by the components
- improve docs of components

### DIFF
--- a/lib/surface/components/form/datetime_select.ex
+++ b/lib/surface/components/form/datetime_select.ex
@@ -4,9 +4,9 @@ defmodule Surface.Components.Form.DateTimeSelect do
 
   Provides a wrapper for Phoenix.HTML.Form's `datetime_select/3` function.
 
-  All options passed via `opts` will be sent to `datetime_select/3`, `value`
+  All options passed via `opts` will be sent to `datetime_select/3`,
+  `value`, `default`, `year`, `month`, `day`, `hour`, `minute`, `second` and `builder`
   can be set directly and will override anything in `opts`.
-
 
   ## Examples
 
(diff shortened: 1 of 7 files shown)


lib/surface/components/form/datetime_select.ex
```diff
@@ -34,15 +34,62 @@ defmodule Surface.Components.Form.DateTimeSelect do
...
+  prop second, :keyword
+
+  @doc """
+  Specify how the select can be build. It must be a function that receives a builder
+  that should be invoked with the select name and a set of options.
+  """
+  prop builder, :fun
+
   @doc "Options list"
   prop opts, :keyword, default: []
```
msaraiva, line 65: This is not being used anymore. Which raises the question: does the underlying `datetime_select/3` and friends accept extra options that are forward to any generated tag as attributes (like in `<Select>`)?

If so, we'll need to keep it and forward them properly. Otherwise, we can remove this declaration.

msaraiva (commented):
@Malian Looks great! I have just a question about `opts`.

lib/surface/components/form/datetime_select.ex
lnr0626: The docs imply that these are the only available options to me (https://hexdocs.pm/phoenix_html/Phoenix.HTML.Form.html#datetime_local_input/3)

lnr0626 (approved):
This looks good to me, it does look like we're good to remove the opts prop as well

lib/surface/components/form/datetime_select.ex
Malian: Mhhh not sure we can remove it as the `opts` is the second argument of the datetime and time builder. You can set a custom "parent" `id` and a custom parent `name` by specify its in `opts`. See these functions [datetime_builder](https://github.com/phoenixframework/phoenix_html/blob/v2.14.2/lib/phoenix_html/form.ex#L1580-L1582) and [datetime_options](https://github.com/phoenixframework/phoenix_html/blob/v2.14.2/lib/phoenix_html/form.ex#L1615-L1616).

I suggest to keep it and add `++ opts` in the following code:

- DateTimeSelect
- DateSelect
- TimeSelect

and add more tests.

Malian: After a few minutes of reflection, I am in favour of adding two props. However, if we decide to do this, it must be done for all the other form fields components.

msaraiva: @Malian in case you confirm that all inputs, selects or whatever other form control can handle `id` and `name` properly, then yes. I think it would be better to promote them as props as you suggested. And since they all will have those properties, you could change: https://github.com/msaraiva/surface/blob/595fc6df20885f3c6a4a190a0a457148344729e5/lib/surface/components/form/utils.ex#L6-L7

to something like:

```elixir
    props = [:id, :name] ++ props
    quote do
      Enum.reduce(unquote(props), [], fn prop, acc ->
      ...
```

This way you don't have to change all `get_non_nil_props/2` calls.

For all the inputs, you can define the new properties in a single place at [Surface.Components.Form.Input](https://github.com/msaraiva/surface/blob/master/lib/surface/components/form/input.ex), but for the others, you'll have to add it manually.

If you prefer, you can add the new properties only to the components related to this PR and then create a separate PR later for the other components. WDYT?

Malian: @msaraiva Thanks for your answer. I will check if every form components can handle `id` and `name`.

In the meantime, I suggest to put the "opts" back into date, time and datetime component mentioned above and make a separate merge request to implement the `id` and `name` in all the form components.

msaraiva: > I will check if every form components can handle id and name.

I'm quite confident that they should since they all use the same underlying functions like `generic_input/4`, `select/4` and `datetime_builder/5`.

> In the meantime, I suggest to put the "opts" back into date, time and datetime component mentioned above and make a separate merge request to implement the `id` and `name` in all the form components.

Yeah, I believe we should keep `opts` even if we add `id` and `name`. This way we keep the component more flexible regarding future changes in the underlying functions. So, yes. I think you can keep `opts` and then later add `id` and `name` to all components.

Malian:
@msaraiva & @lnr0626 The pull request should be ready for a last review. I created an issue to add "id" and "name" props to all form components. I will work on this tomorrow.

msaraiva:
@Malian thanks! ❤️

PR #225, opened 2020-12-03
Improve validation of built-in assigns

msaraiva:
### Improvements

  - Declare all public built-in assigns (helps with validation and also makes them all available for autocomplete)
  - Raise error on duplicate built-in assign
  - Use a different error message when there's a duplicate built-in assigns
  - Consider only the built-in assings specific for each type (Component, LiveComponet and LiveView)


lnr0626 (approved):
LGTM

Issue #226, opened 2020-12-03
Add id and name props to all form components


Malian:
See discussion in #222 and [this](https://github.com/msaraiva/surface/pull/222#discussion_r535200846) discussion.

> I will check if every form components can handle id and name.

I'm quite confident that they should since they all use the same underlying functions like `generic_input/4`, `select/4` and `datetime_builder/5`.

> In the meantime, I suggest to put the "opts" back into date, time and datetime component mentioned above and make a separate merge request to implement the `id` and `name` in all the form components.

Yeah, I believe we should keep `opts` even if we add `id` and `name`. This way we keep the component more flexible regarding future changes in the underlying functions. So, yes. I think you can keep `opts` and then later add `id` and `name` to all components.

_Originally posted by @msaraiva in https://github.com/msaraiva/surface/pull/222#r535454291_

Malian:
@msaraiva  After few investigations, the `name` props will not be an issue. However, the `id` props makes the component stateful.

The doc says: 
> A stateful component is a component that receives an :id on live_component/3

I am not sure we could easily add an `id` props to all form component and keep them stateless.

msaraiva:
You're right. We'll need to change the way the property `id` of a `Surface.Component` is passed to the `live_component`. I'll take a look at it and let you know what I can do before we move on.

msaraiva:
> I am not sure we could easily add an `id` props to all form component and keep them stateless.

@Malian after #227 is merged, we can! 😉

msaraiva:
@Malian #227 has been merged. Please let me know if you run into any issue.

PR #227, opened 2020-12-04
Stateless components with id don't become stateful

msaraiva:
Currently, if you define a prop `id`  in a stateless `Surface.Component`, it becomes a stateful component since the id will automatically be passed to `live_component/3`. However, the user may just want to use the `id` to pass it to some HTML node, like:

```
<input id={{ @id }} ... />
```

This PR fixes this issue. 

Related to #226.

### DIFF
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -173,7 +173,12 @@ defmodule Surface do
         {name, Surface.TypeHandler.runtime_prop_value!(module, name, value, node_alias || module)}
       end)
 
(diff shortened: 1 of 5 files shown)


lnr0626 (approved):
This looks good to me

do we want to add any tests for change tracking with the id parameter?

msaraiva:
Not sure. Since this just guarantees that the component keeps stateless, I don't think we need it but maybe I misunderstood what you mean.

PR #229, opened 2020-12-04
Add id and name props to all form components

### LINKED ISSUE #226: Add id and name props to all form components

Malian:
Fixes #226

### DIFF
--- a/lib/surface/components/form/date_select.ex
+++ b/lib/surface/components/form/date_select.ex
@@ -29,6 +29,12 @@ defmodule Surface.Components.Form.DateSelect do
   @doc "The form identifier"
   prop form, :form
 
+  @doc "The id of a corresponding input field"
+  prop id, :string
+
+  @doc "The id of a corresponding input field"
+  prop name, :string
+
   @doc "The field name"
   prop field, :string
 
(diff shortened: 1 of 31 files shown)


msaraiva (commented):
Looks good so far 👍

Malian:
@msaraiva It should be ready for review.

msaraiva:
@Malian awesome! Thanks ❤️

PR #230, opened 2020-12-05
Allow to define the assign name for slot through the :as option

### LINKED ISSUE #209: Add option :as and validate any undeclared slot
Initial discussion https://github.com/msaraiva/surface/issues/206#issuecomment-730541166.

The main reason we internally use `:__default__` for the default slot is to keep the name `:default` available as it might be a common name for a property, however, this solution is inconsistent as it does not prevent other good names from falling into the same problem. 

### Proposal

Create an optional option `:as` for slots that should be used to name the related assigns. Example:

```elixir
# Defines a :default assign (not :__default__)
slot default

# Defines a :default_slot assign if you want to reuse :default as a property
slot default, as: :default_slot
```

This would make the behaviour of all slots, named or not, consistent.

Additionally, we should make declaring slots mandatory, even for `default` as proposed here https://github.com/msaraiva/surface/pull/208#discussion_r526926439.

Malian:
This pull request fixes #209 

Changes:

- [x] Add `:as` option to slot
- [x] Remove `__default__` name for `default` prop
- [x] Refactoring
- [ ] Doc

### DIFF
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -288,7 +288,6 @@ defmodule Surface do
   end
 
   @doc false
-  def slot_assigned?(assigns, :default), do: slot_assigned?(assigns, :__default__)
   def slot_assigned?(%{__surface__: %{provided_templates: slots}}, slot), do: slot in slots
   def slot_assigned?(_, _), do: false
 
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -181,14 +181,13 @@ defmodule Surface do
       |> Keyword.merge(static_props)
       |> rename_id_if_stateless(module.component_type())
 
-    renamed_slot_props =
+    slot_assigns =
       module
-      |> mapping_to_rename_slots_with_as_option()
-      |> rename_slots_props(slot_props)
+      |> map_slots_to_assigns(slot_props)
 
     Map.new(
       props ++
-        renamed_slot_props ++
+        slot_assigns ++
         [
           __surface__: %{
             slots: Map.new(slots),
@@ -199,13 +198,12 @@ defmodule Surface do
     )
   end
 
-  defp mapping_to_rename_slots_with_as_option(module) do
-    module.__slots__()
-    |> Enum.map(fn %{name: name, opts: opts} -> {name, Keyword.get(opts, :as)} end)
-    |> Enum.filter(fn value -> not match?({_, nil}, value) end)
-  end
+  defp map_slots_to_assigns(module, slot_props) do
+    mapping =
+      module.__slots__()
+      |> Enum.map(fn %{name: name, opts: opts} -> {name, Keyword.get(opts, :as)} end)
+      |> Enum.filter(fn value -> not match?({_, nil}, value) end)
 
-  defp rename_slots_props(mapping, slot_props) do
     slot_props
     |> Enum.map(fn {name, info} -> {Keyword.get(mapping, name, name), info} end)
   end
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -181,9 +181,27 @@ defmodule Surface do
       |> Keyword.merge(static_props)
       |> rename_id_if_stateless(module.component_type())
 
+    renaming_slot_mapping =
+      module.__slots__()
+      |> Enum.map(fn %{name: name, opts: opts} ->
+        if renamed = Keyword.get(opts, :as) do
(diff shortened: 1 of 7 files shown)


Malian:
@msaraiva Ready for review. We should update the doc accordingly when the documentation repository will be available.

Malian:
Here are some of the decisions I made:

- Only the `assigns` keys are modified. The `provided_templates` and `slots` keys under `__surface__` remain the same.
- The `:as` option only accepts `atom`

lib/surface.ex
```diff
@@ -181,9 +181,14 @@ defmodule Surface do
       |> Keyword.merge(static_props)
       |> rename_id_if_stateless(module.component_type())
 
+    renamed_slot_props =
```
msaraiva, line 184: To avoid confusion, maybe we should call this `slot_assigns` this way we have a clear separation between slots/props (inputs) and assigns (internal data). After all, we don't actually rename the props, we just map them to assigns. WDYT?

lib/surface.ex
```diff
@@ -181,9 +181,14 @@ defmodule Surface do
...
+      module
+      |> mapping_to_rename_slots_with_as_option()
```
msaraiva, line 186: And here also. Maybe rename it to something like `map_slots_to_assigns`. The important thing is that it should always be clear if we're dealing with slots (external) or assigns (internal) when naming variables or functions. There should be places where we don't follow this rules, especially regarding props. In the near future we'll probably need to have option `:as` for properties too as they also represent external inputs. So since we we'll have two possible different sets of names around (slots or assigns), I think is a good idea to start following this golden rule.

test/api_test.exs
```diff
@@ -216,9 +311,20 @@ defmodule Surface.APITest do
...
+    test "validate slot as option" do
+      code = "slot cols, as: [:info, {:a, :b}]"
+
+      message =
+        ~r/invalid value for option :as in slot. Expected an atom, got: \[:info, {:a, :b}\]/
+
+      assert_raise(CompileError, message, fn ->
+        eval(code)
+      end)
+    end
```
msaraiva, line 323: Do you think we need both, this one and this one https://github.com/msaraiva/surface/pull/230/files#diff-6edab3c057af39388f5ff5b56b1d815b3e1935acb8ae7782a7c3be52fd96209fR53-R59 ?

test/slot_test.exs
```diff
@@ -425,6 +459,68 @@ defmodule Surface.SlotTest do
     )
   end
 
+  test "rename slot with :as do not overide other assigns with same name" do
```
msaraiva, line 462:
```suggestion
  test "rename slot with :as do not override other assigns with same name" do
```

msaraiva (commented):
@Malian fantastic work! 🎉

I left a couple of comments for your consideration.

lib/surface.ex
Malian: I agree.

lib/surface.ex
Malian: Naming is hard and you are definitely better than me :smile: I updated the code accordingly to your comments.

I started to add the `:as` option for properties but after few minutes I forgot the usefulness of doing this. Why would we like to have something like this?

```elixir
defmodule Component

  prop title, :string, as: :default_title

  render(assigns) do
    ~H"""
       <div>
          {{ @default_tilte }}
       </div>  
    """
  end
end
```

Adding a `:as` option to prop will from my point of view solve the same issue we had with `slot`. As we already offer a proper way to handle this, I am not sure we should offer a second way of fixing the same issue. It may be confusing and add complexity. For example, the user could write something like that:
```elixir
defmodule Component

  slot title, as: :default_title
  prop title, :string, as: :default_title

  render(assigns) do
    ~H"""
       <div>
          <slot name="title" />
          {{ @default_tilte }}
       </div>  
    """
  end
end
```

test/api_test.exs
Malian: No we do not need both. Good catch!

lib/surface.ex
msaraiva: @Malian you're right. Maybe we won't need `:as` for `prop` anymore. Let's move on without it and if it turns out to be useful in the future, we can discuss again.

msaraiva:
@Malian thanks! ❤️

Issue #109, opened 2020-08-11
`Surface.component` macro


lnr0626:
With #15 wrapped up, I think we have most of the things we need to support a `Surface.component` macro that could be used inside LEEx. The generated code should look somewhat similar to https://github.com/msaraiva/surface/blob/master/lib/surface/compiler/eex_engine.ex#L212

I think it would be worthwhile to think through the usage a bit though. The simplest approach would be to have something like:

```elixir
<%= Surface.component(@socket, MyComponent, [prop1: @value, prop2: @value], """
<!-- contents for MyComponent -->
""" %>
```

which would be equivalent to:
```html
<MyComponent prop1={{ @value }} prop2={{ @value }}>
    <!-- contents for MyComponent -->
</MyComponent>
```

However, something that could be used more like the following would probably be more natural within an LEEx template:
```elixir
<%= Surface.component(@socket, MyComponent, [prop1: @value, prop2: @value]) do %>
<!-- contents for MyComponent -->
<% end %>
```

msaraiva:
Yeah, for the default slot it would be as natural as using `live_component`, however, handling named slots will be tricky.

msaraiva:
This feature will also allow users to use the existing `render_component/3` macro in tests as described [here](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#module-testing-components).

Currently, users need to either define a parent LiveView in their tests as in: https://github.com/msaraiva/surface/blob/011ff48401f6b3659f03fbd6a92e7028b0173547/test/live_component_test.exs#L168

or use our [render_live/2](https://github.com/msaraiva/surface/blob/011ff48401f6b3659f03fbd6a92e7028b0173547/test/test_helper.exs#L56) macro (which creates one on the fly) as in: https://github.com/msaraiva/surface/blob/011ff48401f6b3659f03fbd6a92e7028b0173547/test/live_component_test.exs#L123

The latter requires users to copy & paste the helper macro from our `test_helper.exs` which is suboptimal.

With `Surface.component` ready, we could define our own `render_surface_component/3` wrapping `render_component/3` adding support for slots and contexts.

aleDsz:
Using this approach, I'm creating and component library for a design system using Surface to provide a more easy way to develop web application with Phoenix LiveView.

So, I'm trying to make a way to test component renderization using `render_component/3` or `render_live/2` but not what I'm expecting to see.

I tried to use the same tests that you have on Surface but It's not what I want.

[...]

PR #231, opened 2020-12-05
Add render_surface/1 and compile_surface/2 test helpers

msaraiva:
This PR addresses something that has been bothering me for a while which is the amount of tests we have that depends on `render_live/2`. The problem with `render_live/2` is that it creates a module (LiveView) on the fly for each test, consequently, it's very slow and it's making the test suite take much longer than is should. Just to give you an idea of the problem, currently, the tests for the form controls alone are consuming 60% of the time when running the whole suite.

In general, from now on, `render_live/2` should be used sparingly and only for tests that requires either a parent LiveView or some complex interaction between components, e.g. testing events, changing state, etc. Unfortunately, we couldn't use phoenix's built-in `render_component/3` because surface components require some additional setup.

This PR introduces a new `render_surface_component/3` that should be used for most of the cases we have when testing stateless components (like form controls) as well as when testing live component's rendering (not events and their side effects).

After this PR is merged, I'll start migrating most of the tests to use the new function. Since we have lots of tests in this situation, any help with that task will be appreciated. I'll also review what we have in `test_helper.exs` and maybe remove some functions or move them to the new `Surface.LiveViewTest` module. This way those functions can be used by other users in their own tests.

@Malian, @miguel-s and @alexandrubagu since you implemented most of the existing form controls, I'd love to hear your thoughts on this one. Does any of you see any problem with this new approach?

### DIFF
--- a/lib/surface/live_view_test.ex
+++ b/lib/surface/live_view_test.ex
@@ -21,6 +21,7 @@ defmodule Surface.LiveViewTest do
 
   defmacro __using__(_opts) do
     quote do
+      import Phoenix.ConnTest
       import Phoenix.LiveViewTest
       import Phoenix.LiveView.Helpers, only: [live_component: 3, live_component: 4]
       import Surface, only: [sigil_H: 2]
@@ -35,7 +36,7 @@ defmodule Surface.LiveViewTest do
   that don't require a parent live view during the tests.
 
   For tests depending on the existence of a parent live view, e.g. testing events on live
-  components and its side-effects, you need to use either `render_live/2` or
+  components and its side-effects, you need to use either `Phoenix.LiveViewTest.live/2` or
   `Phoenix.LiveViewTest.live_isolated/3`.
 
   ## Example
--- a/lib/surface/live_view_test.ex
+++ b/lib/surface/live_view_test.ex
@@ -3,87 +3,120 @@ defmodule Surface.LiveViewTest do
   Conveniences for testing Surface components.
   """
 
-  alias Surface.TypeHandler
+  alias Phoenix.LiveView.{Diff, Socket}
+
+  defmodule BlockWrapper do
+    @moduledoc false
+
+    use Surface.Component
+
+    slot default, required: true
+
+    def render(assigns) do
+      ~H"""
+      <slot/>
+      """
+    end
+  end
 
   @doc """
-  Helper function to test stateless components or regular rendering of a live component.
+  Helper function to test the regular rendering of components.
 
-  The values passed in `assigns` must be in the runtime format. For instance, a property
-  of type `:css_class` must be passed as list (e.g. `["btn", "active"]`). A property of
-  type `:event` must be passed as `%{name: event_name, target: :live_view}`. This function
-  accepts a block that will be used to fill in the default slot.
+  For tests depending on the existence of a parent live view, e.g. testing events on live
+  components and its side-effects, you need to use either `render_live/2` or
+  `Phoenix.LiveViewTest.live_isolated/3`.
 
   ## Example
 
       html =
-        render_surface_component(Link, to: "/users/1") do
+        render_surface do
           ~H"\""
-          <span>user</span>
+          <Link label="user" to="/users/1" />
           "\""
         end
 
       assert html =~ "\""
-            <a href="/users/1"><span>user</span></a>
+            <a href="/users/1">user</a>
             "\""
 
-  ## Limitations
-
-  Currently, this function cannot be used to test:
-
-    * Slot props
-    * Named slots
-    * Contexts
-    * Events on live components
-
-  If your test depends on any of the features above, you need to use either
-  `render_live/2` or `Phoenix.LiveViewTest.live_isolated/3`.
-
   """
-  defmacro render_surface_component(component, assigns, opts \\ []) do
-    block = Keyword.get(opts, :do)
-
-    init_inner_block =
-      if block do
-        quote do
-          var!(assigns) = %{}
-          inner_block = unquote(block)
-        end
-      else
-        quote do
-          inner_block = nil
-        end
+  defmacro render_surface(do: do_block) do
+    render_component_call =
+      quote do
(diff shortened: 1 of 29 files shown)


msaraiva:
Related to #109.

lib/surface/live_view_test.ex
```diff
@@ -0,0 +1,101 @@
...
+      |> Surface.default_props()
+      |> Keyword.merge(assigns)
+      |> to_runtime_values(component)
+      |> Surface.rename_id_if_stateless(component.component_type())
+      |> Map.new()
+
+    if inner_block do
+      props
+      |> Map.put(:inner_block, fn _, _ -> inner_block end)
+      |> Map.put(:__surface__, %{provided_templates: [:__default__]})
```
Malian, line 82: This will have to be changed if we decide to merge #230 before this one.

Malian (commented):
What I like with the current implementation in tests is that it clearly shows how to use Surface. With the new approach, we lose the "tests as documentation" behaviour. My fear is that a newcomer to Surface will not be able to just look at the test to understand how to use Surface.

Except for this point, the tests are more concise and faster, which is a definite advantage.

miguel-s:
I like that `render_surface_component/3` mirrors LiveView's `render_component/3` and so does not require to re-learn how to test components (initially I was somewhat confused with `render_live/2`).

Obviously improvements in speed are always welcome!

@msaraiva will `render_surface_component/3` also be the recommended way of testing surface components for users of the library?

msaraiva:
> What I like with the current implementation in tests is that it clearly shows how to use Surface. With the new approach, we lose the "tests as documentation" behaviour.

@Malian I agree. It would still be very nice to have something like:

```elixir
  html = render_surface(...) do
    ~H"""
    <Link label="user" to="/users/1" />
    """
  end
```

But without the need to create a module. I'm doing some experiments and it's feasible. I'll have to submit a small PR to Phoenix but it's possible to do a temporary solution until it gets merged.

> will `render_surface_component/3` also be the recommended way of testing surface components for users of the library?

Well it think we could provide both `render_surface_component/3` and `render_surface/1`. I'd prefer the latter for the same reason @Malian already mentioned. 

I'll get back to you as soon as I have `render_surface/1` ready. Then we can decide which one should be the "recommended way" of testing surface components.

Thanks for the inputs!

msaraiva:
Hi folks!

I replaced `render_surface_component /3` with a new `render_surface/1` macro which is not only way way faster than `render_live/2` (since it doesn't create modules) but also allows us to pass Surface's code, consequently it and has no limitation regarding any type of slot or contexts. It also does not require the `warn: false` when defining aliases anymore.

### Example

```elixir
html =
  render_surface do
    ~H"""
    <Link label="user" to="/users/1" />
    """
  end

assert html =~ """
       <a href="/users/1">user</a>
       """
```

This should be the recommended way to test any component except for cases that depend on the existence of a parent live view, e.g. testing events on live components or live views and their side-effects or when you need to test compile-time errors/warnings.

Feedback is welcome!

Malian:
I tested on my pull request #230 and everything seems to work fine. My first findings are that the tests are much more clearer and so are the intentions. Good job!

I suggest to merge your proposal before #230 to reduce your workload. I'll adapt my pull request. I am also available to update the tests as soon as this pull request has been merged.

miguel-s:
Hi!

I tried out `render_surface/1` in my project and it makes for a very nice testing experience. Also the speed improvemens are quite noticeable.

However, I'm not too sure about the setup. to get it to work I added the following imports to my test file (copied from `Surface.ConnCase`):

```Elixir
import Phoenix.LiveViewTest
import Phoenix.LiveView.Helpers, only: [live_component: 3]
import Surface, only: [sigil_H: 2]
import Surface.LiveViewTest
```

Did I miss something or is this the intended way? If so should we add a `use` macro to `Phoenix.LiveViewTest` to clean this up?

Awesome work!

lnr0626 (approved):
This is really cool!

msaraiva:
Sorry for the late response. I was offline for most of the day today so I didn't have much time to implement all changes I wanted but was able to commit at least the minimum we need for the new testing API.

> I'm not too sure about the setup. To get it to work I added the following imports to my test file (copied from `Surface.ConnCase`)... Should we add a `use` macro to `Surface.LiveViewTest` to clean this up?

@miguel-s I did that. The user should do this now:

```
      import Phoenix.ConnTest
      use Surface.LiveViewTest

      # The default endpoint for testing
      @endpoint TheUserEndpoint
```

Or create a `ConnCase` as we did.

All the functions/macros we need for tests are now in `Surface.LiveViewTest`. As soon as we get rid of all `render_live` calls, we can clean up the `test_helper.exs` file. 

The basic rules to use the new API are:

1. Use `render_surface` for all tests of stateless components or when we want to test the regular rendering of a live component. This should cover 80-90% of our tests.
2. Use `compile_surface` to test compile-time errors/warnings
3. Use `compile_surface` + `Phoenix.LiveViewTest.live_isolated/3` to all other cases where a test-specific LiveView is required, i.e, you'll not share the LV between tests. Otherwise, it's better to define the module inline and use it in various tests, if possible.

The reason I didn't create a separate function for `#3` is that I want to be explicit to the user that a the code will be compiled so there's an extra cost.

msaraiva:
@Malian feel free to move on with #230. We should have a much better experience when writing tests from now on. There's still a lot of tests to migrate but as you can see most of them are quite straightforward. The ones I already migrated along with the other changes have already cut the test suite's time in 50% on my machine so it's definitely a big deal.

In case anyone wants to start migrating tests to the new API, please let us all know so we don't waste time working on the same files. I'll start working on these 3 now:

  * `test/directives_test.exs`
  * `test/components/live_redirect_test.exs`
  * `test/components/live_patch_test.exs`

miguel-s:
I finished with `test/components/form/checkbox_test.exs` to try out the new API.
Tomorrow I can work my way down the `test/components/form` folder.

PR #233, opened 2020-12-07
Migrate form component tests

miguel-s:
Related to #231

### DIFF
--- a/test/components/form/date_select_test.exs
+++ b/test/components/form/date_select_test.exs
@@ -1,6 +1,5 @@
 defmodule Surface.Components.Form.DateSelectTest do
   use Surface.ConnCase, async: true
-  import Phoenix.HTML, only: [html_escape: 1]
 
   alias Surface.Components.Form
   alias Surface.Components.Form.DateSelect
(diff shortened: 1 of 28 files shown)


test/components/form/date_select_test.exs
```diff
@@ -1,106 +1,94 @@
 defmodule Surface.Components.Form.DateSelectTest do
-  use ExUnit.Case, async: true
+  use Surface.ConnCase, async: true
+  import Phoenix.HTML, only: [html_escape: 1]
```
miguel-s, line 3: @msaraiva I had to add this import to keep the tests working, as we lose the `import Phoenix.HTML` from `render_live`. Is this ok?

msaraiva: @miguel-s it's totally fine, however, we don't need that anymore. This has been fixed in https://github.com/msaraiva/surface/commit/51878cd551ce294641eb3120b1f9f574cf5a0a35 so you can safely  remove the `html_escape` call here https://github.com/msaraiva/surface/blob/8c1def38c0a29693891f34d2b9997ec10f3f30d7/test/components/form/date_select_test.exs#L116.

msaraiva:
@miguel-s thank you so much! The test suite is soooooo much faster now.

❤️ ❤️ ❤️ ❤️ ❤️

PR #234, opened 2020-12-07
Replace render_live and render_static with render_surface

msaraiva:
This is the last function planned for the new test API and replaces the existing `render_static`. It should be used for any surface code containing only HTML markup + interpolation (no components). Here's the updated list, from the fastest to slowest:

  1. ~`render_surface_markup/1` - Test Surface code containing only HTML markup + interpolation (no components).~

  1. `render_surface/1`-  Test surface markup code, stateless components or the regular rendering of a live component.

  1. `compile_surface/2` - Test compile-time errors/warnings. 

  1. `compile_surface/2` (or inline module) + `Phoenix.LiveViewTest.live_isolated/3` - Test all other cases where a test-specific LiveView is required, i.e, you'll not share the LV between tests. Otherwise, it's better to define the module inline and use it in various tests, if possible.

### DIFF
--- a/lib/surface/live_view_test.ex
+++ b/lib/surface/live_view_test.ex
@@ -30,6 +30,55 @@ defmodule Surface.LiveViewTest do
   end
 
   @doc """
+  Render Surface code containing only HTML markup + interpolation.
+
+  This function is the preferred way to test Surface code that doesn't include components
+  as it's much faster than `render_surface/1`.
+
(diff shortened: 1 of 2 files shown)


lnr0626 (approved):

msaraiva:
It turns out `render_surface_markup ` is just slightly faster than our new `render_surface ` so I decided to remove it as I don't see it justifies having another function in the API for such a minimal gain. That means we can safely replace `render_static` with `render_surface`.

PR #235, opened 2020-12-08
Migrate component_test file, and construct for and if test files

Malian:
As @miguel-s is working on form components in #233 I tackle components outside that scope.

@msaraiva I had to alias `If` and `For` constructs to be able to use `render_surface` macro in tests. As these are also components I am not sure it's the right way to do or if we should modify something in the `render_surface` macro to automatically load these constructs.


msaraiva:
> I am not sure it's the right way to do

@Malian that's the right way to do it :)

Thanks! ❤️

PR #237, opened 2020-12-08
Raises compile error if slots are not declared

### LINKED ISSUE #236: Make declaring slots mandatory
Make declaring slots mandatory, even for `default` as proposed here https://github.com/msaraiva/surface/pull/208#discussion_r526865090.

Related code:

https://github.com/msaraiva/surface/blob/f373b0fbd193e9e60c95cbc42354e56346b031b6/lib/surface/compiler.ex#L661

and

https://github.com/msaraiva/surface/blob/f373b0fbd193e9e60c95cbc42354e56346b031b6/lib/surface/api.ex#L170-L175

There might be other places where conditionals like `slot_name != :default` have been used. We'll probably need to remove them too.

Malian:
This pull request fixes #236 by making the declaring slots mandatory.

- [x] Raises error if slots are not declared in the parent component
- [x] Raises error if slots are not declared in current component with slot macro
- [x] Write better error messages:
  - [x] Some slots are defined, list them
  - [x] No slot is defined, show how to add a slot with the slot macro
  - [x] Short syntax used without declaring the default slot
  - [x] Slotable component
- [x] Refactoring
- [x] Look carefully for dead code

The following codes are not allowed any more:

### Forgetting to declare a named slot
```elixir
    defmodule MyComponent do
      use Surface.Component

      slot header
      slot default

      def render(assigns) do
        ~H"""
          <div>
            <slot name="header"/>
            <slot />
            <slot name="footer" />
          </div>
        """
      end
    end
```
### Forgetting to declare the `default` slot
```elixir
    defmodule MyComponent do
      use Surface.Component

      slot default

      def render(assigns) do
        ~H"""
          <div>
            <slot />
          </div>
        """
      end
    end
```
### Using a named slot that are not defined in the parent component
```elixir
    defmodule MyComponent do
      use Surface.Component

      slot default

      def render(assigns) do
        ~H"""
          <div>
            <slot />
          </div>
        """
      end
    end
    defmodule MyView do
      use Surface.LiveView

      alias MyComponent

      def render(assigns) do
        ~H"""
          <div>
             <MyComponent>
                <template name="header">
                   My Header
                </template>
             </MyComponent>
          </div>
        """
      end
    end
```
### Using the default slot without declaring it in the parent component
```elixir
    defmodule MyComponent do
      use Surface.Component

      slot default

      def render(assigns) do
        ~H"""
          <div>
            <slot />
          </div>
        """
      end
    end
    defmodule MyView do
      use Surface.LiveView

      alias MyComponent

      def render(assigns) do
        ~H"""
          <div>
             <MyComponent>
                My Content
             </MyComponent>
          </div>
        """
      end
    end
```

All these cases will raise a CompileError with proper messages.

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -748,9 +748,11 @@ defmodule Surface.Compiler do
       if component_slotable?(template_meta.module) do
         """
         The slotable component <#{template_meta.module}> as the `:slot` option set to \
-        `#{slot_name}`.\n
-        That slot name is not declared in parent component <#{parent_meta.node_alias}>.\n
-        Please declare the slot in the parent component or rename the value in the `:slot` option.
+        `#{slot_name}`.
+
+        That slot name is not declared in parent component <#{parent_meta.node_alias}>.
+
+        Please declare the slot in the parent component or rename the value in the `:slot` option.\
         """
       else
         """
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -444,12 +444,8 @@ defmodule Surface.Compiler do
 
   defp has_attribute?([], _), do: false
 
-  defp has_attribute?(attributes, attr_name) do
-    Enum.find_value(attributes, false, fn
-      {^attr_name, _, _} -> true
-      _ -> nil
-    end)
-  end
+  defp has_attribute?(attributes, attr_name),
+    do: Enum.any?(attributes, &match?({^attr_name, _, _}, &1))
 
   defp attribute_value_as_ast(attributes, attr_name, default, meta) do
     Enum.find_value(attributes, default, fn
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -727,8 +727,9 @@ defmodule Surface.Compiler do
          true = _short_syntax?
        ) do
     message = """
-    no slot `#{slot_name}` defined in the component `#{inspect(module)}`\
-    \n\nPlease declare the default slot using `slot default` in order to use the `<slot />` notation.
+    no slot `#{slot_name}` defined in the component `#{inspect(module)}`
+
+    Please declare the default slot using `slot default` in order to use the `<slot />` notation.
     """
 
     IOHelper.compile_error(message, meta.file, meta.line)
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -655,10 +655,7 @@ defmodule Surface.Compiler do
   end
 
   defp validate_templates(mod, templates, meta) do
-    names =
-      templates
-      |> Map.keys()
-      |> Enum.reject(fn name -> name == :default end)
+    names = templates |> Map.keys()
 
     if !function_exported?(mod, :__slots__, 0) and not Enum.empty?(names) do
       message = """
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -744,8 +744,22 @@ defmodule Surface.Compiler do
 
     existing_slots_message = existing_slots_message(parent_slots)
 
+    header_message =
+      if component_slotable?(template_meta.module) do
+        """
+        The slotable component <#{template_meta.module}> as the `:slot` option set to \
+        `#{slot_name}`.\n
+        That slot name is not declared in parent component <#{parent_meta.node_alias}>.\n
+        Please declare the slot in the parent component or rename the value in the `:slot` option.
+        """
+      else
+        """
+        no slot "#{slot_name}" defined in parent component <#{parent_meta.node_alias}>\
+        """
+      end
+
     message = """
-    no slot "#{slot_name}" defined in parent component <#{parent_meta.node_alias}>\
+    #{header_message}\
     #{similar_slot_message}\
     #{existing_slots_message}
     """
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -252,27 +252,22 @@ defmodule Surface.Compiler do
     index =
       attribute_value_as_ast(attributes, "index", %Surface.AST.Literal{value: 0}, compile_meta)
 
-    with true <- not is_nil(name) and is_atom(name),
-         {:ok, directives, _attrs} <-
-           collect_directives(@slot_directive_handlers, attributes, meta) do
-      # TODO: REFACTOR
-      if name in defined_slot_names do
-        {:ok,
-         %AST.Slot{
-           name: name,
-           index: index,
-           directives: directives,
-           default: to_ast(children, compile_meta),
-           props: [],
-           meta: meta
-         }}
-      else
-        short_slot_syntax? = false == attribute_value(attributes, "name", false)
-
-        missing_slot(meta.caller.module, name, meta, defined_slot_names, short_slot_syntax?)
-      end
+    with {:ok, directives, _attrs} <-
+           collect_directives(@slot_directive_handlers, attributes, meta),
+         true <- name in defined_slot_names do
+      {:ok,
+       %AST.Slot{
+         name: name,
+         index: index,
+         directives: directives,
+         default: to_ast(children, compile_meta),
+         props: [],
+         meta: meta
+       }}
     else
-      _ -> {:error, {"failed to parse slot", meta.line}, meta}
+      _ ->
+        short_slot_syntax? = not has_attribute?(attributes, "name")
+        missing_slot(meta.caller.module, name, meta, defined_slot_names, short_slot_syntax?)
     end
   end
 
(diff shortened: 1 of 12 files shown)


test/context_test.exs
```diff
@@ -240,7 +242,7 @@ defmodule ContextTest do
           ~H"""
           <Context
             put={{ ContextTest.Outer, 123 }}>
-            <slot/>
```
Malian, line 243: This slot does not make sense as the code is compiled in a live view and live view haven't slots.

test/constructs/for_test.exs
```diff
@@ -55,7 +55,7 @@ defmodule Surface.Constructs.ForTest do
       quote do
         ~H"""
         <For each={{ fruit <- ["apples", "bananas", "oranges"] }}>
-          <ListProp prop="some string">The inner content</ListProp>
+          <ListProp prop="some string" />
```
Malian, line 58: `ListProp` does not have a `default` slot so I decided to remove the inner content. An alternative could be to update the `ListProp` component and add the `default` slot.

test/slot_test.exs
```diff
@@ -63,20 +63,6 @@ defmodule Surface.SlotTest do
     end
   end
 
-  defmodule OuterWithoutDeclaringSlots do
```
Malian, line 66: Moved to a "compile error" test

test/slot_test.exs
```diff
@@ -620,12 +581,7 @@ defmodule Surface.SlotTest do
       end
 
     message = """
-    code:1: there's no `default` slot defined in `Surface.SlotTest.OuterWithoutDefaultSlot`.
```
Malian, line 623: Unfortunately we loose this proper error message as the compiler will detect that the `default` slot is not defined before looking at the `:let` directive.

test/transform_test.exs
```diff
@@ -51,7 +57,7 @@ defmodule Surface.TransformTest do
     @impl true
     def render(assigns) do
       ~H"""
-      <div><slot /></div>
```
Malian, line 54: Slot are not allowed in LiveView

lib/surface/compiler.ex
```diff
@@ -250,17 +255,22 @@ defmodule Surface.Compiler do
...
+         %AST.Slot{
+           name: name,
+           index: index,
+           directives: directives,
+           default: to_ast(children, compile_meta),
+           props: [],
+           meta: meta
+         }}
+      else
+        short_slot_syntax? = false == attribute_value(attributes, "name", false)
```
Malian, line 270: To ugly. I should find a better way to do this check.

test/context_test.exs
msaraiva: @Malian maybe we should consider adding a `compile_surface_component` to the test API so we could generate a component instead of a LV. We only need to create a LV on-the-fly when we need to test events and side-effects. To test compilation errors in components, generating a component would be enough. I can see we currently use `Code.eval_string(...)` to do that in many places. Le me know what you think.

lib/surface/compiler.ex
```diff
@@ -241,17 +241,20 @@ defmodule Surface.Compiler do
...
+      |> Surface.API.get_slots()
+      |> Enum.map(& &1.name)
+
     # TODO: Validate attributes with custom messages
     name = attribute_value(attributes, "name", :default)
 
     index =
       attribute_value_as_ast(attributes, "index", %Surface.AST.Literal{value: 0}, compile_meta)
 
-    with true <- not is_nil(name) and is_atom(name),
```
Malian, line 250: I removed this condition because the value of the name will always be fixed and will always be an atom with at least the a default value `:default` regarding the following function.

https://github.com/msaraiva/surface/blob/f373b0fbd193e9e60c95cbc42354e56346b031b6/lib/surface/compiler.ex#L425-L431

As the next line will also always return `:ok` I simplified the code. 

Feel free to tell me if I should revert this part.

lib/surface/compiler.ex
```diff
@@ -262,7 +265,9 @@ defmodule Surface.Compiler do
          meta: meta
        }}
     else
-      _ -> {:error, {"failed to parse slot", meta.line}, meta}
+      _ ->
+        short_slot_syntax? = not has_attribute?(attributes, "name")
+        missing_slot(meta.caller.module, name, meta, defined_slot_names, short_slot_syntax?)
```
msaraiva, line 270: Maybe we should call this function `raise_missing_slot` or even `raise_missing_slot_error` to make it clear what they do. The same goes to all similar functions. WDYT?

lib/surface/compiler.ex
```diff
@@ -718,33 +716,80 @@ defmodule Surface.Compiler do
     :ok
   end
 
-  defp missing_slot(mod, slot_name, template_meta, parent_meta) do
+  defp missing_slot(module, slot_name, meta, _defined_slot_names, true) do
+    message = """
+    no slot `#{slot_name}` defined in the component `#{module}`\
```
msaraiva, line 721:
```suggestion
    no slot `#{slot_name}` defined in the component `#{inspect(module)}`\
```

Let's use `inspect` here to avoid messages showing the module with the `Elixir.` prefix. The same goes for all other messages.

test/slot_test.exs
```diff
@@ -871,19 +832,82 @@ defmodule Surface.SlotSyncTest do
...
+
+    Available slots: "footer", "header" and "default"
+    """
+
+    assert_raise(CompileError, message, fn ->
+      compile_surface(code)
+    end)
+  end
+
+  test "raises compile error on component that uses slots without declaring its" do
```
msaraiva, line 848:
```suggestion
  test "raises compile error on component that uses undeclared slots" do
```

lib/surface/compiler.ex
```diff
@@ -430,6 +435,15 @@ defmodule Surface.Compiler do
     end)
   end
 
+  defp has_attribute?([], _), do: false
+
+  defp has_attribute?(attributes, attr_name) do
+    Enum.find_value(attributes, false, fn
+      {^attr_name, _, _} -> true
+      _ -> nil
```
msaraiva, line 443: Since this function ends with `?`, it's probably a good idea to guarantee it returns a boolean.

```suggestion
      _ -> false
```

lib/surface/compiler.ex
```diff
@@ -718,33 +716,80 @@ defmodule Surface.Compiler do
...
+
+    existing_slots_message = existing_slots_message(defined_slot_names)
+
+    message = """
+    no slot `#{slot_name}` defined in the component `#{module}`\
+    #{similar_slot_message}\
+    #{existing_slots_message}\
+    \n\nHint: You can define slots using the `slot` macro. \
+    \nFor instance: `slot #{slot_name}`\
+    """
```
msaraiva, line 739: I think it's easier to read without those `\n`s :)

```suggestion
    message = """
    no slot `#{slot_name}` defined in the component `#{inspect(module)}`\
    #{similar_slot_message}\
    #{existing_slots_message}
    
    Hint: You can define slots using the `slot` macro.
    
    For instance: `slot #{slot_name}`\
    """
```

msaraiva (commented):
@Malian fantastic work!

I left a few comments for your consideration.

Thank you so much for working on this. ❤️

lib/surface/compiler.ex
lnr0626: since it's going to raise, should we also append a `!` to the name?

lib/surface/compiler.ex
```diff
@@ -636,10 +650,7 @@ defmodule Surface.Compiler do
   end
 
   defp validate_templates(mod, templates, meta) do
-    names =
-      templates
-      |> Map.keys()
-      |> Enum.reject(fn name -> name == :default end)
+    names = templates |> Map.keys()
```
lnr0626, line 653:
```suggestion
    names = Map.keys(templates)
```

lnr0626 (commented):
just a couple minor style things :heart:

lib/surface/compiler.ex
Malian: It always returns a boolean due to the `false` argument in `find_value`. Replacing `nil` by `false` and remove the second attribute of `find_value` will return `nil`. Replacing `nil` by `false` and keeping `false` as a default value will return `false` because of the default value and not because of `_ -> false`. This being said, it gains in clarity.

msaraiva: You're absolutely right. I missed the default value. Sorry. Just leave as it is or, if you want to make it more explicit, you can use:

```elixir
Enum.any?(attributes, &match?({^attr_name, _, _}, &1))
```

It's less overhead in my mind as reading it describes exactly what it does but I think we're stepping on the field of personal taste now.  So just ignore me, if you want. 🙂

Malian: Thanks for your proposition! I The intent is clearer than mine without the ambiguity I introduced ;)

Malian:
@msaraiva & @lnr0626 Thanks you for your comments.

msaraiva:
@Malian thanks! ❤️ ❤️ ❤️ ❤️ ❤️

PR #242, opened 2020-12-11
Remove default value from Form method prop

miguel-s:
By having Surface set a default value for the `method` prop in the `Form` component we are forcing this value to always be "post" unless explicitly defined.

This causes issues with Phoenix's `form_for` e.g. automatically setting the method to "put" depending on the changeset.

Phoenix's `form_for` already has a default for it's `method` option, so I don't expect issues removing it from Surface.

### DIFF
--- a/test/components/form_test.exs
+++ b/test/components/form_test.exs
@@ -149,7 +149,6 @@ defmodule Surface.Components.FormTest do
 
     {:ok, _view, html} = live_isolated(conn, ViewWithForm, session: assigns)
 
-    # assert html =~ ~s(<input name="_method" type="hidden" value="put" />)
     assert html =~ ~s(><input name="_method" type="hidden" value="put"/>)
   end
 end
--- a/test/components/form_test.exs
+++ b/test/components/form_test.exs
@@ -4,6 +4,14 @@ defmodule Surface.Components.FormTest do
   alias Surface.Components.Form
   alias Surface.Components.Form.TextInput
 
+  defmodule User do
+    use Ecto.Schema
+
+    schema "user" do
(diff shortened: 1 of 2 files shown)


test/components/form_test.exs
```diff
@@ -123,4 +123,33 @@ defmodule Surface.Components.FormTest do
...
+
+  test "form generates method input for changeset", %{conn: conn} do
+    assigns = %{
+      "changeset" =>
+        Ecto.Changeset.cast(
+          {%{}, %{name: :string}},
+          %{name: "myname"},
+          [:name]
+        )
+        |> Map.put(:data, %{__meta__: %{state: :loaded}})
```
miguel-s, line 147: @msaraiva This does not feel right, but couldn't figure out how to create a "loaded" changeset in this test. Any ideas?

msaraiva: Hey @miguel-s!

You could define a schema and use `Ecto.put_meta/2`:

```elixir
  defmodule User do
    use Ecto.Schema

    schema "user" do
      field :name, :string
    end
  end

  test "form generates method input for changeset", %{conn: conn} do
    changeset =
      %User{}
      |> Ecto.put_meta(state: :loaded)
      |> Ecto.Changeset.cast(%{name: "myname"}, [:name])

    assigns = %{"changeset" => changeset}

    {:ok, _view, html} = live_isolated(conn, ViewWithForm, session: assigns)

    assert html =~ ~s(><input name="_method" type="hidden" value="put"/>)
  end
```

Can't see any other way.

msaraiva:
@miguel-s thanks! ❤️

Issue #243, opened 2020-12-14
Use `hidden` attribute for :show directive


maennchen:
Currently the `:show` directive emits inline styles to hide an element. This does not work well if the page that uses it uses a strict CSP.

Using the `hidden` attribute, the same can be achieved using no inline styles. The attribute is accepted very well by current browsers:

https://caniuse.com/hidden

PR #244, opened 2020-12-14
:show directive via hidden attribute

maennchen:
Fix for #243

### DIFF
--- a/lib/surface/directive/show.ex
+++ b/lib/surface/directive/show.ex
@@ -27,37 +27,50 @@ defmodule Surface.Directive.Show do
   def extract(_, _), do: []
 
   def process(%AST.Directive{value: %AST.Literal{value: value}}, node) do
-    if value do
-      node
-    else
-      add_value_to_attribute(node, :style, value)
-    end
+    add_value_to_attribute(node, :style, %AST.AttributeExpr{
+      value:
+        quote generated: true do
+          unquote(value)
+        end,
+      original: to_string(value),
+      meta: true
+    })
   end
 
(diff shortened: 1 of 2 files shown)


lib/surface/directive/show.ex
```diff
@@ -27,37 +27,52 @@ defmodule Surface.Directive.Show do
...
+              quote generated: true do
+                not unquote(value)
+              end
+        },
+        meta: show_meta
+      }
+      | non_style
+    ]
+
+    # IO.inspect attributes
```
Malian, line 74: Slight oversight 😉

maennchen: Fixed :)

msaraiva:
Hey @maennchen!

Sorry for the late reply. I've been mostly on holidays for the last few days so I might be offline for a day or two.

About the PR. I like it! One thing I think we should do now is to change the name of the directive from `:show` to `:visible` as the latter sounds more intuitive as the opposite  of `hidden`. Let me know what you think and also if you want to do this in the context of this PR, otherwise I can merge it as it is and then work on the renaming on a separate PR.

Thanks a lot for you contribution.

maennchen:
@msaraiva I actually like the name `show`. There's quite a few projects naming it the same way:

- Vue: https://vuejs.org/v2/guide/conditional.html#v-show
- Angular: https://docs.angularjs.org/api/ng/directive/ngShow
- etc.

I would therefore recommend to keep the name since it's a well understood name for people building interactive web applications.

msaraiva:
@maennchen thanks! ❤️

PR #249, opened 2020-12-28
Fix arity of mount callback in docs

tschmidleithner:
See also https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#c:mount/3


msaraiva:
@tschmidleithner thanks! ❤️

PR #263, opened 2021-02-06
Don't validate undefined assign outside render

### LINKED ISSUE #261: Don't validate undefined assign outside `render/1`
The following gives a `undefined assign `@message` warning:

```elixir
  defp message(assigns) do
    ~H"""
    <div>
      {{ @message }}
    </div>
    """
  end
```

The warning should appear only if the caller module is a Surface Component/Liveview and the caller function is `render/1`.

Malian:
Fixes #261

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -133,7 +133,10 @@ defmodule Surface.Compiler do
     end
   end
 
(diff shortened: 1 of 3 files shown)


msaraiva:
@Malian thanks! ❤️

Issue #255, opened 2021-01-19
New type for html attributes


lnr0626:
the `:attrs` directive has some differences compared to passing attributes to `content_tag` based functions in phoenix ((e.g. `link` would convert `phx_change` to `phx-change`, and `data: [thing: "value"]` to `data-thing="value"`))

Adding a new type handler for html attributes to handle that behavior would address that inconsistency

msaraiva:
You're right. We should try to follow the same rules.

msaraiva:
@lnr0626 after this [PR](https://github.com/phoenixframework/phoenix_html/pull/331) gets merged and we have a new version of `phoenix_html` released, we should change the implementation of `:attrs` and any other stuff we do in that regard to use the new `Tag.attributes_escape/1`. We'll probably need to preprocess a couple of our existing types before calling it but at least everything will work more consistently.

PR #264, opened 2021-02-06
Allow component children in link and label

### LINKED ISSUE #262: Allow <Link> and <Label> to accept components as children
As reported in `#168` and `#250`, Phoenix's `link` and `label` functions depend on `content_tag`, i.e. they don't allow including components in their inner block. If you try to do that, you get  the following error:

```
** (exit) an exception was raised:
    ** (ArgumentError) cannot convert component Surface.Components.Form.Checkbox with id nil to HTML.
A component must always be returned directly as part of a LiveView template.
For example, this is not allowed:
    <%= content_tag :div do %>
      <%= live_component @socket, SomeComponent %>
    <% end %>
...
```

We should roll our own implementation without depending on `link` and `label` just like we did for `LivePatch` and `LiveRedirect`.

miguel-s:
Fixes #262.

In order to pass all tests for the `Link` component I had to jump through a few hoops and do a "mini" implementation of #255 (e.g. translating the `data` option to `"data-x"` tags). As it stands, I only implemented the cases needed to pass the tests, but not general cases like changing names with "_" to "-".

In other components like `LivePatch` we don't have any translation and rely on the user doing it manually e.g. `"data-confirm": "Really?"` instead of `data: [confirm: "Really?"]`. However, I did not want to change the tests and remove functionality of `Link` to match `LivePatch` without discussing it firsts.

What do you think?

### DIFF
--- a/lib/surface/components/link.ex
+++ b/lib/surface/components/link.ex
@@ -1,11 +1,12 @@
 defmodule Surface.Components.Link do
   @moduledoc """
-  Defines a hyperlink.
+  Generates a link to the given URL.
 
-  Provides a wrapper for Phoenix.HTML.Link's `link/2` function.
+  Provides similar capabilities to Phoenix's built-in `link/2`
+  function.
 
-  All options passed via `opts` will be sent to `link/2`, `label` and `class` can
-  be set directly and will override anything in `opts`.
+  Options `label` and `class` can be set directly and will override anything in `opts`.
+  All other options are forwarded to the underlying <a> tag.
 
   ## Examples
   ```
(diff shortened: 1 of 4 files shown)


lib/surface/components/link.ex
```diff
@@ -27,37 +28,68 @@ defmodule Surface.Components.Link do
...
   use Surface.Component
 
-  import Phoenix.HTML.Link, only: [link: 2]
-
-  @doc "Place to link to"
+  @doc "The page to link to"
   prop to, :string, required: true
 
+  @doc "The method to use with the link"
+  prop method, :atom, default: :get
```
msaraiva, line 35: Does this bring https://github.com/msaraiva/surface/pull/242 back?

lib/surface/components/link.ex
```diff
@@ -27,37 +28,68 @@ defmodule Surface.Components.Link do
...
+        :csrf_token -> {:"data-csrf", value}
+        :phx_click -> {:"phx-click", value}
+        :phx_target -> {:"phx-target", value}
+        :method -> method_to_attrs(value, assigns.to)
+        :data -> data_to_attrs(value)
+        _ -> {key, value}
+      end
+    end
+    |> List.flatten()
+  end
```
msaraiva, line 81: I think this is ok for now. I'm working on a PR for `phoenix_html` that will expose a function to generate attributes consistently. As soon as it gets merged, we'll replace the `:attrs` implementation to use it, which will also fix #255.

msaraiva (commented):
@miguel-s Thanks for the PR! this looks great.

lib/surface/components/link.ex
miguel-s: Couldn't reproduce #242

If the value of `method` is `:get` (default or explicitly set) it won't gat added as an attribute to the \<a> tag, see line  85. This was inspired by Phoenix's `link/2`.

msaraiva:
@miguel-s thanks! ❤️

Issue #259, opened 2021-01-28
Replace <template> and <slot> with <Template> and <Slot> to avoid conflicts


lnr0626:
This is related to what i mentioned in https://github.com/msaraiva/surface/issues/107#issuecomment-741912493

I recently found the need for an actual `<template>...</template>` node for one of my hooks, and had to do:
```
<#Raw><template></#Raw>
...
<#Raw></template></#Raw>
```

while this does work, it's not ideal - should we switch from `<template>` and `<slot>` nodes to `<Template>` and `<Slot>` components (with some time supporting both to make it a bit easier to do the switch over)? Or is there another way to allow template nodes that are treated as standard html template elements?

msaraiva:
I think we should switch to `<Template>` and `<Slot>` as soon as possible since we've seen more and more people facing this issue, like is was mentioned here https://elixirforum.com/t/surface-a-component-based-library-for-phoenix-liveview/27671/84. Maybe we don't even need to implement them as independent components for now and just keep the current hardcoded implementation.

> with some time supporting both to make it a bit easier to do the switch over

I don't think we should worry about it. If people are using `<#Raw><template>` to overcome that, it should be straightforward to replace it after we release a new version.

lnr0626:
> Maybe we don't even need to implement them as independent components for now and just keep the current hardcoded implementation.

+1 to that

regarding making it easier to switch, i was more referring to switching from `<template>` and `<slot>` to `<Template>` and `<Slot>` - i.e. release a new patch version that has support for both, then in the next minor remove support for `<template>` and `<slot>` (supporting both for a short time should be trivial if we're keeping the current hard coded implementation, and can log a warning when we encounter the lower cased version)

msaraiva:
> supporting both for a short time should be trivial if we're keeping the current hard coded implementation, and can log a warning when we encounter the lower cased version

On. I see. That sounds perfect indeed!

msaraiva:
Closing this in favour of https://github.com/surface-ui/surface/issues/340.

PR #268, opened 2021-02-13
Allow capitalized <Slot> and <Template> tags

cschmatzler:
As discussed in #259, this keeps everything hard coded and leaves available the downcased tags.

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -188,9 +188,11 @@ defmodule Surface.Compiler do
   end
 
   defp node_type({"#" <> _, _, _, _}), do: :macro_component
-  defp node_type({<<first, _::binary>>, _, _, _}) when first in ?A..?Z, do: :component
   defp node_type({"template", _, _, _}), do: :template
+  defp node_type({"Template", _, _, _}), do: :template
   defp node_type({"slot", _, _, _}), do: :slot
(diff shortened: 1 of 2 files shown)


Malian:
Thanks! I suggest to deprecate the usage of `<slot>` and `<template>` with a proper warning message. In a future release we will be able to remove these tag properly being sure to have warned the user beforehand. What do you think?

msaraiva:
I believe we can finally move on with this. Surface `v0.4` will introduce a few syntax changes. One of them is the use of `#` for all built-in constructs. We'll include `<template>` and `<slots>` in that list. So instead of `<Template>` and `<Slot>` we'll have `<#template>` and `<#slot>`.

We should keep `<template>` and `<slot>` working on `v0.4` but emit a deprecation warning instructing the user to manually update the code or run a built-in converter that will be shipped with Surface.

@cschmatzler please let us know if you still want to adjust the PR for those changes. Otherwise, we can implement them in the scope of the other upcoming changes.

Cheers.

msaraiva:
The full list of changes planned to `v0.4` and `v0.5` are listed in https://github.com/surface-ui/surface/issues/340 along with the link to the original proposal.

msaraiva:
Resolved in #https://github.com/surface-ui/surface/commit/42813934bd198d9222d25ebca8705ee3ac8cce67.

PR #271, opened 2021-02-24
Add support for auto-registrable/live-reloadable hooks

### LINKED ISSUE #269: Support colocated .js files for self-registrable/live-reloadable  hooks
Hi folks!

Does anyone out there ever wondered if we could have a more intuitive way to define **JS hooks for components**?

Yesterday I finally had the change to put something together. 

What have I managed to do, so far?

  1. Automatically load component hooks exported by a colocated `.js` file
  2. Automatically namespace imported hooks to avoid conflicts
  3. Automatically register loaded hooks to the socket
  4. Safely import hooks defined by other components
  4. Support live reload for component hooks

What can we still do?

  1. Static checking for missing hook files
  2. Static checking for unused hook files
  3. Static checking for undefined hooks in a file

As I'm still playing with it, I'd love to hear your thoughts on the idea, especially about limitations. I'm sure we'll not be able to cover all cases but I tested this approach with all components I had that require custom JS and it worked perfectly!

Here's the branch if you want to try it out https://github.com/msaraiva/surface/tree/ms-component-assets

## Setup

Update `mix.exs` to point to the branch above and set up the Surface compiler:

```elixir
defp deps do
  [
    ...,
    {:surface, git: "git@github.com:msaraiva/surface.git", branch: "ms-component-assets", override: true}
  ]
end

def project do
  [
    ...,
    compilers: [:surface, :phoenix] ++ Mix.compilers()
  ]
end
```

Update `.gitignore`:

```
# Ignore auto-generated hook files
/assets/js/hooks/
```

Update `assets/js/app.js`:

```js
import hooks from "./hooks"

// You can merge it with other non-surface hooks, if you need
let Hooks = {...hooks}
```

Update the Endpoint options in `config/dev.exs` for live reload of JS hooks:

```elixir
config :my_app, MyAppWeb.Endpoint,
  ...
  reloadable_compilers: [:phoenix, :elixir, :surface],
  live_reload: [
    patterns: [
      ~r"lib/my_app_web/(live|components)/.*(ex|js)$",
      ...
    ]
```

## Usage

Create a colocated `.js` file with the same base name of the component.

Example:

```
card.ex
card.js
```

Export the hooks you need in `card.js`:

```js
let Card = {
  mounted(){
    console.log("Card mounted")
  }
};

let OtherHook = {
  mounted(){
    console.log("OtherHook mounted")
  }
};

export {Card, OtherHook};
```

Use the `:hook` directive to bind whatever hook you need:

```vue
<div :hook="Card">
  ...
</div>
```

You can also access hooks defined by other components using option `:from`:

```vue
<div :hook={{ "Card", from: MyComponents.CardList }}>
  ...
</div>
```

That should be all! 

Let me know if you run into any issue. I want to have as much feedback as possible to make sure the idea is solid before moving on with this.

Cheers.

msaraiva:
Resolves #269.


lnr0626 (approved):
this is really cool; only question I have is how does this interact with dependencies (i.e. if a hook requires some lib available, does the user just need to install that themselves via npm/yarn?)

msaraiva:
Yup. Just add the dependency and import it as usual.

Issue #272, opened 2021-02-25
<Link> no longer accepts "link with scheme"?


neilberkman:
After upgrading to 0.3.0 a number of tests in my application started failing due to `<Link>` tags with `to` attributes with tuples of the form `{:scheme, :link}`. `Phoenix.HTML.Link` accepts this format:

https://github.com/phoenixframework/phoenix_html/blob/2d09867665295ac124b8b4913f14b6b80b0cb952/test/phoenix_html/link_test.exs#L56-L57

I'm using it for "deep links" to apps on mobile.

msaraiva:
@miguel-s I believe we'll need to partially copy this `valid_destination!` implementation out of the `link`code. Correct?

https://github.com/phoenixframework/phoenix_html/blob/2d09867665295ac124b8b4913f14b6b80b0cb952/lib/phoenix_html/link.ex#L147

miguel-s:
Thanks for pointing me to the relevant places in the Phoenix codebase @neilberkman @msaraiva.

PR #273 should fix this particular issue, but I'll check if we missed anything else to keep feature parity with Phoenix.

msaraiva:
Hey @miguel-s! That was fast :)

Thank you! ❤️

msaraiva:
@neilberkman thanks for reporting this issue.

PR #274, opened 2021-02-28
Improve Link component compatibility with Phoenix link/2

miguel-s:
Follow-up to #272 and #273.

Surface's `<Link>` component should now behave as Phoenix's `link/2`.
I replicated most of their test cases and fixed a couple of issues that emerged (mainly around how we handle csrf tokens).


msaraiva:
@miguel-s awesome! ❤️

PR #277, opened 2021-03-01
Allow question mark as a valid char in tag attribute name

### LINKED ISSUE #275: Using question mark in prop names at the end
Hello, great work on Surface!

I've been taking a look at it and I think it's great
(might be a bit biased coming from a Vue background :)

But, I did stumble upon an issue regarding prop definition.

When I define a prop which name includes `?` at the end (e.g. `prop raised?, :boolean`)
it allows it. But, when trying to pass it down like this:

```mustache
<Button id="home_page_button3" raised?={{ true }}>
  Open it!
</Button>
```

You'll see this error:

```
(Surface.Compiler.ParseError) expected opening HTML tag

Stacktrace:
  │ (surface 0.3.0) lib/surface/compiler.ex:110: Surface.Compiler.compile/5
  │ (surface 0.3.0) expanding macro: Surface.sigil_H/2
```

Removing the `?` from the prop name works as expected.

I think Elixir allowing us to use the question mark in names is really cool.
And I'd like to take advantage of it if possible at all.

Do you know if it is feasible and what kind of effort is required for it?

Cheers!

Malian:
Fix #275 

I followed the rules for attribute name here: https://www.w3.org/TR/2008/REC-xml-20081126/#NT-Name

It's maybe a little bit too complicate. As an alternative, we could simply add `??` in the current implementation.

```diff
- attr_name = ascii_string([?a..?z, ?0..?9, ?A..?Z, ?-, ?., ?_, ?:, ?@], min: 1)
+ attr_name = ascii_string([?a..?z, ?0..?9, ?A..?Z, ?-, ?., ?_, ?:, ?@, ??], min: 1)
```

### DIFF
--- a/lib/surface/compiler/parser.ex
+++ b/lib/surface/compiler/parser.ex
@@ -106,7 +106,12 @@ defmodule Surface.Compiler.Parser do
     |> wrap()
     |> post_traverse(:attribute_value)
 
-  attr_name = ascii_string([?a..?z, ?0..?9, ?A..?Z, ?-, ?., ?_, ?:, ?@], min: 1)
+  # Follow the rules of XML Names and Tokens plus "@" as a first char
+  # https://www.w3.org/TR/2008/REC-xml-20081126/#NT-Name
+  attr_start_char = ascii_string([?a..?z, ?A..?Z, ?_, ?:, ?@], 1)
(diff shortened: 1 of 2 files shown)


msaraiva:
❤️

PR #278, opened 2021-03-01
Support full list of void elements from HTTP specification

paulstatezny:
For some context:

- This started with [this PR on surface_formatter](https://github.com/surface-ui/surface_formatter/pull/12).
- [Here's the list of void elements from the HTTP spec that I used](https://www.w3.org/TR/2011/WD-html-markup-20110113/syntax.html#void-element).

From @msaraiva:

> I think we should update our list on the parser/compiler to have all of them, including the possibly deprecated ones.

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -51,15 +51,18 @@ defmodule Surface.Compiler do
     "base",
     "br",
     "col",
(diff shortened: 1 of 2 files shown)


Malian:
@paulstatezny thanks! ❤️

PR #282, opened 2021-03-05
Warn if prop is required and has default value

### LINKED ISSUE #281: Warn if prop is required and has default value
Reading the documentation, I was surprised that several examples show a `prop` that is `required` AND has a default value. The example in the README, for example, shows:

```elixir
prop name, :string, required: true, default: ""
```

This feels strange to me: what is the point of providing a default value if the prop **must** be defined anyway?

On the other hand, it seems pretty happy with a `prop` that is **not** required but has no default. In that case it seems the default "default" is `nil`, which makes sense but I haven't seen it mentioned in the doc.

My impression is that both options are antagonist. I would expect to have a warning when:
- `required` is `true` and there is a default ("default value will always be ignored")
- `required` is `false` and there is no default

What do you think? Did I miss something?

Malian:
Fixes #281 

Thanks @colinsmetz! This change caught 3 misuses of a prop with a default value and the required opt set at true.

### DIFF
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -260,6 +260,7 @@ defmodule Surface.API do
          :ok <- validate_opts_keys(func, name, type, opts),
          :ok <- validate_opts(func, type, opts),
          :ok <- validate_required_opts(func, type, opts) do
+      maybe_warn_required_prop_with_default_value(func, type, opts, caller)
       :ok
     else
       {:error, message} ->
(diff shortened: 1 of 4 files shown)


lib/surface/api.ex
```diff
@@ -260,6 +260,7 @@ defmodule Surface.API do
          :ok <- validate_opts_keys(func, name, type, opts),
          :ok <- validate_opts(func, type, opts),
          :ok <- validate_required_opts(func, type, opts) do
+      maybe_warn_required_prop_with_default_value(func, type, opts, caller)
```
lnr0626, line 263: how about `maybe_warn_mutually_exclusive_opts(...)` instead as we're running everything through instead of just props?

lnr0626 (commented):
this lgtm, just one minor naming suggestion

msaraiva:
❤️

Issue #265, opened 2021-02-07
Render function not run when component declared as a "slot"


ewildgoose:
I wanted to make a card component with optional <Header> and <Footer> parts. So starting with the example in the docs we see that the wrapper HTML is always defined and the slot is included in the middle. 

The obvious solution seems to be to create Header/Footer components which will render the appropriate HTML. However, using pure components means that the user needs to know precisely how to insert them inside the Card component. slots seem to solve this

So I tried defining a component like this:

```
defmodule ExampleWeb.Component.Card.Footer do
  @moduledoc false

  use Surface.Component, slot: "footer"

  prop class, :css_class, required: false, default: "bg-gray-50 px-5 py-3 sm:px-5 text-lg text-gray-500"
  slot default

  def render(assigns) do
    ~H"""
      <div class={{ @class }}>
        <slot name="default"/>
      </div>
    """
  end
end
```

However, the action of declaring that it's a "slot:" seems to stop render being called when it's inserted into the component upstream?

So I would desire that

```
  <Card>
    stuff
    <Footer class="some classes">
      a footer
    </Footer>
  </Card>
```

Will both render the Footer component *and* insert it into the Card component where I use the given slot. This seems useful more widely as well I think? eg your Grid example could acquire a default render function for the Column?

However, the original use case was to have a component with an optional slot where the content of the slot is run through a template function, perhaps there is another way to handle this?

msaraiva:
Hi @ewildgoose!

There's a similar issue here https://github.com/msaraiva/surface/issues/212 where we've discussed the reasons why slotable components are currently renderless and also presented some possible alternatives. We will probably create a separate type of component (maybe `Surface.Template`) to avoid confusion. Please let me know in case you need any extra information or have any suggestion.

ewildgoose:
Hmm, ok, so I think I see why this is confusing. However, I think the confusion comes (to a new user like me) because we are both using the renderless component for config only in the case of the "Grid.Columns" example, but with content in the slot in the case of the Card example

Having experimented a little more, I'm still not sure that I see the issue? Apologies because I'm probably at least 2 steps behind you.

So sticking with the card example above, I notice that trying to define two "Footer" slots simply keeps the first one:
```
<Footer>
  First footer
</Footer>
<Footer>
  I never see this footer
</Footer>

...then render in the card component with 
  <slot name="footer"/>
```

So it *feels* to be in this case as though the Footer component is behaving both as a renderless and a render-full component.

[...]

Issue #212, opened 2020-11-21
Investigate a :slot directive to avoid template tag


Malian:
I am playing with Surface and slots for few days now and I encountered many times the following type of repetitive code where I set several slots.

```html
      <Card>
        <template slot="header">
          <CardHeader title="A simple card header" />
        </template>

        This example demonstrates how to create components with multiple slots.
        It defines a <strong>default</strong> slot to hold the card's content
        and two <strong>named slots</strong>: header and footer.

        <template slot="footer">
          <CardFooter items={{ [{"link 1", "to1"}, {"link2", "to2"}] }} />
        </template>
      </Card>
```

I tried to fix this with "Typed Slotables" but it seems that the "Slotable" components are simple container without a `render(assigns)` function.

I am wondering if we could add a  `:slot` directive to a component to avoid the `template` tag and reduce boilerplate code.

```html
      <Card>
        <CardHeader title="A simple card header" :slot="header" />

        This example demonstrates how to create components with multiple slots.
        It defines a <strong>default</strong> slot to hold the card's content
        and two <strong>named slots</strong>: header and footer.

        <CardFooter items={{ [{"link 1", "to1"}, {"link2", "to2"}] }} :slot="footer" />
      </Card>
```

Am not sure if it's feasible neither useful, so I am asking your point of view about the proposition or any other alternative.

msaraiva:
Hi @Malian!

Although a `:slot` directive might turn out to be useful for some rare scenarios, I don't think you need it for this specific case. Could you share the code of `CardHeader` and `CardFooter`? If not, can you tell me why the code you have there is not in the `Card` component?

Malian:
The card was an example but I have a concrete case below.

First I have a "PageTitle" component that accepts a "page_title" prop and a "right" slot.
```elixir
defmodule AppWeb.Admin.Layout.PageTitle do
  use AppWeb, :surface_component

  prop page_title, :string, required: true

  @doc "A slot that is pushed on the right of the title"
  slot actions

  def render(assigns) do
    ~H"""
      <div class="border-b border-gray-200 px-4 py-4 flex items-center justify-between sm:px-6 lg:px-8">
        <div class="flex-1 min-w-0 py-2">
          <h1 class="text-lg font-medium leading-6 text-gray-900 sm:truncate">
            {{ @page_title }}
          </h1>
        </div>
        <div class="flex" :if={{ slot_assigned?(:actions) }}>
          <slot name="actions"/>
        </div>
      </div>
[...]

Issue #189, opened 2020-11-03
Raise error/warning when using :for/:if on a renderless component


msaraiva:
We should raise a compile-time error if the user tries to use `:for` or any incompatible directive on a renderless component. 

It should accept the same directives as `<template>`, which currently is just `:let`:https://github.com/msaraiva/surface/blob/b745e1322ce7b20fcbcf8503f507349da0772435/lib/surface/compiler.ex#L40.

Initial discussion started at https://github.com/msaraiva/surface/issues/182.

msaraiva:
@lnr0626 does this still apply? How https://github.com/surface-ui/surface/pull/283 impact this?

Zurga:
Maybe there is a use case for looping over renderless components: to changes to the children. 

In SurfaceBulma, I am trying to implement card the Card with the following capability:
```
<Card>
  <Footer>
     <a>Footer 1</a>
     <a>Footer 2</a>
     <a>Footer 3</a>
   </Footer>
</Card>
```
I want to either, inject the `card-footer-item` class into the `a` tags specified in the `Footer`, or I would like to wrap them in a `<div class="card-footer-item">`. 
I have not found a way yet to do this, if there is one way to do it at the moment, please let me know.

msaraiva:
Since we're deprecating `:for` in favour of `{#for}`, I think we can let this one go.

lnr0626:
@msaraiva do we want to keep it around for `:if`?

msaraiva:
I totally missed `:if`. Yeah, I guess we should keep it.

phgrey:
probably have the same issue when using `:if`, `{#if}`, `:for` & `{#for}`

surface version:
`"surface": {:hex, :surface, "0.10.0", "54eeb1bc7d3e411e3cbd1074237f4369afb24c4e3177ba26f6aaf62336533197", ...`

Have parent component Tabs.List with slot :tabs
Children component Tabs.Tab is defined to hold the :tabs slot's content - `use Surface.Component, slot: "tabs"`

When using Tabs.Tab with :if or :for directive - got error:
```
** (KeyError) key :opts not found in: nil. If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
    (surface 0.10.0) lib/surface/compiler/eex_engine.ex:777: Surface.Compiler.EExEngine.generator_binding/2
[...]

PR #283, opened 2021-03-07
allow renderable slotable components

lnr0626:


test/slot_test.exs
```diff
@@ -378,6 +426,118 @@ defmodule Surface.SlotTest do
...
+
+  test "slotable component with slot props" do
+    assigns = %{items: [%{id: 1, name: "First"}, %{id: 2, name: "Second"}]}
+
+    html =
+      render_surface do
+        ~H"""
+        <Grid items={{ user <- @items }}>
+          <ColumnWithRenderAndSlotProps title="column title" :let={{ info: info }}>
+            <b>{{ info }}: {{ user.id }}</b>
```
lnr0626, line 475: as is, this exposes the `Grid` prop named info; however, `ColumnWithRenderAndSlotProps` also has a prop named info.

@msaraiva @miguel-s @Malian any thoughts on what the expected behaviour would be in this situation?

lib/surface/compiler/eex_engine.ex
```diff
@@ -406,25 +406,39 @@ defmodule Surface.Compiler.EExEngine do
...
-         buffer,
-         state
-       ) do
+  defp handle_templates(component, [slotable | tail], buffer, state) do
+    %AST.SlotableComponent{
+      slot: name,
+      module: module,
+      let: let,
+      props: props,
+      templates: %{default: default}
```
Malian, line 415: Why filter other slots? This prevents the user to use slots inside a slottable component.

The use case I have is the following:

```elixir
defmodule MyAppWeb.Layout do
  use MyAppWeb, :surface_component

  alias MyAppWeb.Layout.PageTitle
  alias MyAppWeb.Layout.Sidebar
  alias MyAppWeb.Layout.MobileHeader

  slot default, required: true
  slot page_title, as: :page_title_slot
  prop page_title, :string, required: false

  def render(assigns) do
    ~H"""
      <div
        class="h-screen flex overflow-hidden bg-white"
        x-data="{ sidebarOpen: false }"
        @keydown.window.escape="sidebarOpen = false"
      >
        <Sidebar />
        <div class="flex flex-col w-0 flex-1 overflow-hidden">
          <main class="flex flex-col flex-1 relative z-0 overflow-y-auto focus:outline-none">
            <MobileHeader />

             <slot name="page_title">
               <PageTitle page_title={{ @page_title }} :if={{ @page_title }} />
             </slot>

             <slot name="default" />

           </main>
        </div>
      </div>
      """
  end
end
```

and the `PageTitle` component:
```elixir
defmodule MyAppWeb.Admin.Layout.PageTitle do
  use MyAppWeb, [:surface_component, [slot: "page_title"]]

  @doc "The main page title"
  prop page_title, :string, required: true

  @doc "A value that is pushed on the right of the title"
  slot actions

  def render(assigns) do
    ~H"""
      <div class="border-b border-gray-200 px-4 py-4 flex items-center justify-between sm:px-6 lg:px-8">
        <div class="flex-1 min-w-0 py-2">
          <h1 class="text-lg font-medium leading-6 text-gray-900 sm:truncate">
            {{ @page_title }}
          </h1>
        </div>
        <div class="flex" :if={{ slot_assigned?(:actions) }}>
          <slot name="actions"/>
        </div>
      </div>
    """
  end
end

```

I use it like this:

Simple page with simple title:
```elixir
<Layout page_title="dashboard">
  ...
</Layout>
```

Complexe page with additional actions at the right of the page title:

```elixir
<Layout>
  <PageTitle page_title="List Vehicles">
    <template slot="actions">
      <LiveRedirect to={{ ... }}>
          New Vehicle
      </LiveRedirect>
    </template>
  </PageTitle>
</Layout>
```
The later does not work with the current implementation. I modified the code to allow "extra" slots and it seems to work but maybe I am missing something.

```suggestion
      templates: %{default: default} = templates
```

and lines 428 to 430
```diff
-              templates: %{
-                default: default
-              },
+             templates: templates,
```

With that small change, I am even able to create a sub slotable component:
```elixir
defmodule MyAppWeb.Admin.Layout.PageTitle.Actions do
  use MyAppWeb, [:surface_component, [slot: "actions"]]

  slot default

  def render(assigns) do
    ~H"""
      <span class="order-0 sm:order-1 sm:ml-3 shadow-sm rounded-md">
        <slot name="default" />
      </span>
    """
  end
end
```
And use it in my code like this:
```elixir
<Layout>
  <PageTitle page_title="List Vehicles">
    <PageTitle.Actions>
      <LiveRedirect to={{ ... }}}>
        New Vehicles
      </LiveRedirect>
    </PageTitle.Actions>
  </PageTitle>
</Layout>
```

This opens up the field of possibilities to create a powerful layout system, doesn't it?

@msaraiva This is the kind of "layout" system I had in mind in #212

lnr0626: Good catch, I'll make those changes. Thanks for taking a look!

msaraiva:
This is also related to https://github.com/msaraiva/surface/issues/265.

lib/surface/compiler/eex_engine.ex
msaraiva: @Malian I like the `Layout` idea 👍

test/slot_test.exs
msaraiva: @lnr0626 I believe the Grid's `info` prop is not passed automatically to the slotables so there should be no conflict. The info var retrieved by `:let` should always be the one explicitly passed by `:props` inside `ColumnWithRenderAndSlotProps `. Right? If the user needs the Grid's `info` if has to retrieve it explicitly creating another variable with a different name, like:

```
<Grid items={{ user <- @items }} :let={{ info: grid_info }}>
```

As far as I can remember, only `user` will be passed automatically so if it happens to have the same name, we should have some kind of warning related to variable shadowing or something similar. Isn't the Elixir compiler already doing that for us?

lnr0626: sort of - the issue here is the info prop on the grid component is on the cols slot, not the default slot; and the info slot on the ColumnWithRenderAndSlotProps is for its default slot. Because we're enabling renderable slotables, we now have two different slots that can each have their own props (in this case, `Grid.cols` and `ColumnWithRenderAndSlotProps.default`) where it's unclear which slot's props the `:let` would refer to. With a non-renderable component, we could do something like:

```
<Grid items={{ user <- @items }}>
  <template name="cols" :let={{ info: grid_info }}>
    <ColumnWithRenderAndSlotProps title="column title" :let={{ info: col_info }}>
      ...
    </ColumnWithRenderAndSlotProps>
  </template>
</Grid>
```

however, because this makes the slotable components combine the `template` and `<Component .../>` bits, we currently don't have a way to `:let` one of the two. With how it's currently setup, `:let` on a slotable component will expose props from the parent component, not the slotable component's default slot.

the relevant bits of those components:
```
  defmodule Grid do
    use Surface.Component

    prop items, :list, required: true

    slot cols, props: [:info, item: ^items]
    ...
  end

  defmodule ColumnWithRenderAndSlotProps do
    use Surface.Component, slot: "cols"

    prop title, :string, required: true

    slot default, props: [:info]
    ...
  end
```

this is really hard to describe clearly :-/

lnr0626: playing around with svelte, it looks like their behavior is close to what's happening currently (i.e. let exposes the parent's props for the parents named slot, and the inner components default slot props aren't accessible):

Grid.svelte:
```
<div style="border: 1px solid black; padding: 4px">
	<slot name="content" info="info from grid" />
</div>
```

Column.svelte:
```
<div style="border: 1px solid blue; padding: 4px; margin: 4px">
	<slot info="info from column" />
</div>
```

App.svelte:
```
<script>
	import Grid from './Grid.svelte';
	import Column from './Column.svelte';
</script>

<Grid>
	<Column slot="content" let:info>
		{ info }
	</Column>
</Grid>

<Column let:info>
	{ info }
</Column>
```

results in:
image

msaraiva: It makes sense, although it can be a source of confusion to many users and we can't blame them. I'm confused myself :)

msaraiva: Maybe we can work around this by providing some warnings if we detect any ambiguity.

lnr0626: haha, me as well

logging a warning on conflicts would be a reasonable solution imo - i'll work on that and see what it looks like

Malian:
This is also related to #189 as this PR makes Slotable component renderable.

lnr0626:
@Malian we'll still have the same restrictions around directives for slotable components, and at least as is this PR doesn't add any additional support or error logging for unknown directives

lib/surface/api.ex
```diff
@@ -381,7 +381,38 @@ defmodule Surface.API do
     end
   end
 
-  defp maybe_warn_mutually_exclusive_opts(_, _, _, _), do: nil
+  defp maybe_warn_mutually_exclusive_opts(:slot, :default, _, opts, caller) do
```
lnr0626, line 384: this seemed the closest place we have currently for logging this type of warning; but I'm up for moving it if we think of a better place for this warning

lib/surface/compiler.ex
```diff
@@ -591,6 +591,7 @@ defmodule Surface.Compiler do
 
   defp collect_templates(mod, attributes, nodes, meta) do
     # Don't extract the template directives if this module is slotable
+    # TODO: and renderless ?
```
lnr0626, line 594: this TODO is because I wasn't sure if we needed to filter out renderable slotable components here. I don't /think/ we need to change this, but I could be missing something

lnr0626 (commented):
@msaraiva i've updated it to log a warning when users specify props for the default slot of a slotable component

msaraiva:
@lnr0626 I'll review this one over the weekend. I'll release `v0.3.2` leaving this one out. It will probably be better to include it in `v0.4.0` as it introduces important features with possible breaking changes.

msaraiva (commented):
@lnr0626 looks good to me! 

I believe the warning will help a lot to avoid confusion. I'm not sure yet if this is the best API to solve this problem but it seems it's a great first step in that direction. Maybe having a different type of components like `Surface.RenderlessComponent` or `Surface.SlotableComponent` would help somehow to organize the code and avoid some conditionals but I really don't know at this point. 

It would be great to have more real examples of this feature in action. I need to get used to it :) It's still confusing to me some times when I look at some of the examples we have. But I believe as soon as I start using it, the confusion should go away. I see it opens some nice possibilities when it comes to composability so I think it's a great addition.

It would be also nice if some of the people that requested this feature could give it an early try and see if it's what they expected or not. They may provide some great feedback about the API and if we're indeed on the right track. I believe we are.

Great work! ❤️

lib/surface/api.ex
```diff
@@ -291,7 +291,7 @@ defmodule Surface.API do
     with :ok <- validate_type(func, name, type),
          :ok <- validate_opts_keys(func, name, type, opts),
          :ok <- validate_opts(func, type, opts) do
-      maybe_warn_mutually_exclusive_opts(func, type, opts, caller)
+      maybe_warn_mutually_exclusive_opts(func, name, type, opts, caller)
```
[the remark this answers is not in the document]
lnr0626: I wasn't sure how to accumulate all of the warnings with the way this method is written, so just moved the warnings inside the relevant `validate_opt` variant. 

I also added the `opts` variable to that method so we could handle any additional mutually exclusive opts inside `validate_opt` in the future

msaraiva (approved):
@lnr0626 looks good to me!

I believe this solves https://github.com/surface-ui/surface/issues/212 and https://github.com/surface-ui/surface/issues/265.

@Malian  and @ewildgoose on the issues above you provided real scenarios for this feature. How did you end up solving the problem? Would you now replace the implementation using this feature? Does it solve the issue in a more intuitive way? Are any of you able to try it out to confirm?

@olivermt since you wrote the whole suite of components for `surface_bulma`, do you feel this feature could have been useful for any of the components you wrote?

**Note:** we need to update the docs accordingly after this gets merged. Even the vocabulary we use around slotable or renderless components may need to change to avoid confusion.

Malian (approved):
@msaraiva I'll solve my issue the way I explained it in this comment https://github.com/surface-ui/surface/pull/283#discussion_r589080332

I will move from
```elixir
<Layout>
  <PageTitle page_title="List Vehicles">
    <template slot="actions">
      <LiveRedirect to={{ ... }}>
          New Vehicle
      </LiveRedirect>
    </template>
  </PageTitle>
</Layout>
```
to
```elixir
<Layout>
  <PageTitle page_title="List Vehicles">
    <PageTitle.Actions>
      <LiveRedirect to={{ ... }}}>
        New Vehicles
      </LiveRedirect>
    </PageTitle.Actions>
  </PageTitle>
</Layout>
```

Regarding the issue with the `:props` directive, I haven't face it yet. I think the warning is enough.

LGTM

msaraiva:
Alright! @lnr0626 feel free to merge this and let us know what you have in mind about the docs and if you think we should improve/change the vocabulary around renderless/slotable in any way.

Thanks for working on this issue. Great stuff! ❤️

lnr0626:
I'll try to find some time to look at an update the docs this week

PR #284, opened 2021-03-07
Improve Label component compatibility with Phoenix label/2

miguel-s:
Follow-up to #264.

We decided to go with our own implementation for the `<Label />` component instead of relying on Phoenix's `label/2` due to #262. We should, however, make sure that our component is compatible with Phoenix's.

This PR copies over and adapts Phoenix's tests for their `label/2` implementation. I commented out all tests that are not relevant (e.g. tests for implementation details like `do:` blocks), but decided to keep them in the code as reference.

I also moved the label test file into the form folder, where it belongs.

And lastly, I also snuck in some tests for `<Link />` :D


msaraiva:
> I commented out all tests that are not relevant (e.g. tests for implementation details like `do:` blocks), but decided to keep them in the code as reference.

@miguel-s I'm probably missing something but I wonder why those tests are not relevant. Shouldn't we keep them?

miguel-s:
@msaraiva on second thought I believe you're right, they're actually pointing out some missing behaviours on our end.

Regarding tests for inline and block content, I'll translate them to Surface's props (inline) and slots (block), do you agree?

test/components/link_test.exs
```diff
@@ -180,24 +180,79 @@ defmodule Surface.Components.LinkTest do
...
+        render_surface(do: ~H[<Link label="foo" opts={{ bar: "baz" }} />])
+      end
+
+      # msg = "link/2 requires a keyword list as second argument"
+
+      # assert_raise ArgumentError, msg, fn ->
+      #   link("foo", "/login")
+      # end
+
+      msg = "<Link /> requires a label prop or contents in the default slot"
```
miguel-s, line 239: @msaraiva this Phoenix test doesn't play well with our "creates a link without label" test on line 35. Should we change our implementation to enforce a label or content like Phoenix?

test/components/link_test.exs
```diff
@@ -180,24 +180,79 @@ defmodule Surface.Components.LinkTest do
...
+    end
+
+    test "link with invalid args" do
+      msg = "expected non-nil value for :to in <Link />"
+
+      assert_raise ArgumentError, msg, fn ->
+        render_surface(do: ~H[<Link label="foo" opts={{ bar: "baz" }} />])
+      end
+
+      # msg = "link/2 requires a keyword list as second argument"
```
miguel-s, line 233: @msaraiva this Phoenix test is very specific to their implementation, not sure it makes sense to force it here, what do you think?

test/components/link_test.exs
```diff
@@ -180,24 +180,79 @@ defmodule Surface.Components.LinkTest do
...
-      html = render_surface(do: ~H[<Link to={{ {:javascript, {:safe, 'alert(<1>)'}} }} />])
-      assert html =~ ~s[<a href="javascript:alert(<1>)"></a>]
+      html =
+        render_surface(do: ~H[<Link label="foo" to={{ {:javascript, {:safe, 'alert(<1>)'}} }} />])
+
+      assert html =~ ~s[<a href="javascript:alert(<1>)">foo</a>]
+    end
+
+    test "link with invalid args" do
+      msg = "expected non-nil value for :to in <Link />"
```
miguel-s, line 227: @msaraiva not too happy how this one turned out. The `:to` prop is marked as required and Surface already gives us a warning. Don't know how to make the Surface warning and the ArgumentError (raised in the Link update function) play nice together, any ideas?

test/components/link_test.exs
msaraiva: I guess it would be nice to have that validation. After all, we can't have a link without one of them.

test/components/link_test.exs
msaraiva: Just remove it.

test/components/link_test.exs
msaraiva: No need to add this one either. As you mentioned, `to` is already required so we got it covered.

lib/surface/components/form/label.ex
```diff
@@ -23,6 +26,11 @@ defmodule Surface.Components.Form.Label do
   @doc "The CSS class for the underlying tag"
   prop class, :css_class
 
+  @doc """
+  The text for the generated `<label>` element, if no content (default slot) is provided.
+  """
+  prop text, :any
```
lnr0626, line 32: should this be a `:string` with `default: nil`?

msaraiva: TIL :)

msaraiva (approved):
Great work @miguel-s! ❤️

Squash and merge it!

PR #286, opened 2021-03-10
Add some documentation types to the API

### LINKED ISSUE #280: Add :pid as a valid type
I recently made a project where at some point I needed to store a PID as `data`. However, I was surprised to see that the `:pid` type didn't exist, so I had to use `:any`.

Would it make sense to add `:pid` to the list of valid types?

Malian:
Fixes #280 

I decided to only add the following new types to the API:

- `:struct`
- `:tuple`
- `:pid`
- `:port`
- `:reference`
- `:bitstring`

I chose the basic types that had an associated macro `is_{type}`  in the [Kernel](https://hexdocs.pm/elixir/1.11.3/Kernel.html) model to test them.

Base on the kernel page we should also consider to add the following types that do not have an associated `is_{type}` macro:

- `:range`
- `:mapset`
- `:naive_datetime`
- `:regex`
- `:time`
- `:uri`
- `:path`
- `:range`

We could also wait until demands come up.


lnr0626:
i think adding those makes some sense - we already have `:datetime` and `:date`, so i don't see a reason to not have the others you listed

msaraiva (approved):

Malian:
@lnr0626 I share  your point of view so I added them.

PR #290, opened 2021-03-12
Warn on LiveComponent with another LiveComponent as root

### LINKED ISSUE #288: Warn on LiveComponent with another LiveComponent as root
Currently, there's an LV limitation that demands a unique root node for each LiveComponent. If you use a LiveComponent as the root of another one, like:

```elixir
  use Surface.LiveComponent
  ...

  def render(assigns) do
    ~H"""
    <Modal id="company_form_modal" vertically_centered header={{ @title }}>
    ...
    </Modal>
    """
  end
```

Since `Modal` is also a live component, you'll end up with a single node for both, leading to the following runtime error:

image

As you can see, the message does not help much the user to solve the problem, which is as simple as wrapping the root component with a div, like:

```elixir
  def render(assigns) do
    ~H"""
    <div>
      <Modal id="company_form_modal" vertically_centered header={{ @title }}>
      ...
      </Modal>
    </div>
    """
  end
```

We can improve DX by providing a compile-time check. Something like:

```
warning: cannot have a LiveComponent as root node of another LiveComponent.

Hint: You can wrap the root `Modal` node in another element. Example:

  def render(assigns) do
    ~H"""
    <div>
      <Modal ... >
      ...
    </div>
    """
  end
```

Malian:
Fixes #288

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -144,10 +144,22 @@ defmodule Surface.Compiler do
     num_tags =
       ast
       |> Enum.filter(fn
-        %AST.Tag{} -> true
-        %AST.VoidTag{} -> true
-        %AST.Component{} -> true
-        _ -> false
+        %AST.Tag{} ->
+          true
+
+        %AST.VoidTag{} ->
+          true
+
+        %AST.Component{type: Surface.LiveComponent, meta: meta} ->
+          warn_live_component_as_root_node_of_another_live_component(meta, caller, offset)
+
+          true
+
+        %AST.Component{} ->
+          true
+
+        _ ->
+          false
       end)
       |> Enum.count()
 
@@ -172,6 +184,28 @@ defmodule Surface.Compiler do
   end
 
   defp validate_stateful_component(_ast, %CompileMeta{}), do: nil
+
+  defp warn_live_component_as_root_node_of_another_live_component(meta, caller, offset) do
+    IOHelper.warn(
+      """
+      cannot have a LiveComponent as root node of another LiveComponent.
+
+      Hint: You can wrap the root `#{meta.node_alias}` node in another element. Example:
+
+        def render(assigns) do
+          ~H"\""
+          <div>
+            <#{meta.node_alias} ... >
+              ...
+            </#{meta.node_alias}>
+          </div>
+          "\""
+        end
+      """,
+      caller,
+      fn _ -> offset end
+    )
+  end
 
   defp to_ast(nodes, compile_meta) do
     for node <- nodes do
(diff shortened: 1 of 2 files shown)


lib/surface/compiler.ex
```diff
@@ -144,10 +144,22 @@ defmodule Surface.Compiler do
...
+        %AST.Tag{} ->
+          true
+
+        %AST.VoidTag{} ->
+          true
+
+        %AST.Component{type: Surface.LiveComponent, meta: meta} ->
+          warn_live_component_as_root_node_of_another_live_component(meta, caller, offset)
+
+          false
```
Malian, line 156: Returning false here will result in 2 warnings. The new warning +  the old one.

msaraiva, line 156: Shouldn't be `true`? 

For the most common scenario for this warning, we'll have a single LiveComponent as root. By returning `false`, the `num_tags` will be `0`, resulting in two warnings. The one we expected and another `stateful live components must have a HTML root element`, which can be confusing. WDYT?

**Note:** it might be a good idea to `assert` against the whole `output` using `==` instead of `=~`. This way we make sure we don't have more warnings than we expect.

msaraiva (commented):
Man, you're fast! 🚀

I left just a comment for your consideration.

lib/surface/compiler.ex
Malian: I posted the same comment few minutes ago ^^ https://github.com/surface-ui/surface/pull/290#discussion_r593397013. I agree, it should be true.

I'll assert the warning with `==` but to be consistent we should update the code everywhere we use `=~` to test warnings. If you think it's important, I can take some time this weekend to update the tests.

msaraiva: Haha. Sorry, I didn't see your comment :)

> we should update the code everywhere we use =~ to test warnings. If you think it's important, I can take some time this weekend to update the tests.

On second thought, forget about using `==`. It's not worth it. Just use `=~` as before and let's move on. I'll think of a better way to assert we don't have extra warnings.

Feel free to merge it after this change. Great work! ❤️

Malian: The tests do not pass on the CI  but they pass on my computer because my terminal handles colours. The output of the warning on my computer starts with `\e[33mwarning: \e[0m` but `warning:` in the CI. It's maybe for that reason why we use `= ~` instead of `==`. I'll look a this tommorow.

Malian: Too many cross messages here :sweat_smile:

msaraiva (approved):

PR #292, opened 2021-03-13
Some small refactoring

Malian:
I started to review the codebase to better understand in depth Surface. In the course of my review, I will propose below several small changes to clarify the intentions, to strengthen the API, to reduce duplicated code, to standardize some part of the code to be more consistent, variable naming, etc.

I believe the better way to review this PR is commit by commit. I'll try to post a comment for every non-obvious decision that would have been taken. It will simplify which refactoring we keep and which we remove.

### DIFF
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -41,8 +41,6 @@ defmodule Surface.API do
     :context_get
   ]
 
-  @private_opts [:action, :to]
-
   defmacro __using__(include: include) do
     arities = %{
       prop: [2, 3],
@@ -173,15 +171,24 @@ defmodule Surface.API do
   end
 
   @doc false
+  def get_props(module) do
+    Module.get_attribute(module, :prop) || []
+  end
+
+  @doc false
+  def get_data(module) do
+    Module.get_attribute(module, :data) || []
+  end
+
+  @doc false
   def get_defaults(module) do
-    for %{name: name, opts: opts} <- Module.get_attribute(module, :data),
-        Keyword.has_key?(opts, :default) do
+    for %{name: name, opts: opts} <- get_data(module), Keyword.has_key?(opts, :default) do
       {name, opts[:default]}
     end
   end
 
   defp quoted_data_funcs(env) do
-    data = Module.get_attribute(env.module, :data) || []
+    data = get_data(env.module)
 
     quote do
       @doc false
@@ -193,11 +200,13 @@ defmodule Surface.API do
 
   defp quoted_prop_funcs(env) do
     props =
-      (Module.get_attribute(env.module, :prop) || [])
-      |> Enum.sort_by(&{&1.name != :id, !&1.opts[:required], &1.line})
+      env.module
+      |> get_props()
+      |> sort_props()
 
-    props_names = Enum.map(props, fn prop -> prop.name end)
+    props_names = for p <- props, do: p.name
     props_by_name = for p <- props, into: %{}, do: {p.name, p}
+    required_props_names = for %{name: name, opts: opts} <- props, opts[:required], do: name
 
     quote do
       @doc false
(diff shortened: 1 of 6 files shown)


lib/surface/api.ex
```diff
@@ -192,9 +201,7 @@ defmodule Surface.API do
   end
 
   defp quoted_prop_funcs(env) do
-    props =
-      (Module.get_attribute(env.module, :prop) || [])
-      |> Enum.sort_by(&{&1.name != :id, !&1.opts[:required], &1.line})
```
Malian, line 197: I haven't found a reason why the props are sorted like this. Tests are still going to pass even if we remove the sorting, except for the `Hint` inside a warning, so I moved the logic of sorting the props to the function that generate the warning message.

If there are any reasons why we have to sort props, I'll be glad to revert this change.

lib/surface/api.ex
```diff
@@ -41,8 +41,6 @@ defmodule Surface.API do
     :context_get
   ]
 
-  @private_opts [:action, :to]
```
Malian, line 44: I haven't found any reference of private_opts like `:action` or `: to` in the codebase.

lib/surface/api.ex
```diff
@@ -365,16 +369,6 @@ defmodule Surface.API do
 
   defp maybe_warn_mutually_exclusive_opts(_, _, _, _), do: nil
 
-  defp validate_required_opts(func, type, opts) do
```
Malian, line 368: This function will always return `:ok` as the `get_required_opts` always return `[]`. Moreover, I haven't found any references to required opts in the code.

lib/surface/api.ex
```diff
@@ -214,6 +219,11 @@ defmodule Surface.API do
       def __get_prop__(name) do
         Map.get(unquote(Macro.escape(props_by_name)), name)
       end
+
+      @doc false
+      def __required_props_names__() do
```
Malian, line 224: To be consistent with the existing `__required_slots_names__()`

lib/surface/base_component.ex
```diff
@@ -19,8 +19,8 @@ defmodule Surface.BaseComponent do
 
   @optional_callbacks transform: 1
 
-  defmacro __using__(opts \\ []) do
-    type = Keyword.get(opts, :type)
+  defmacro __using__(opts) do
+    type = Keyword.fetch!(opts, :type)
```
Malian, line 23: As the `:type` option is required, moving from `get` to `fetch!` will give a better message in case the developer does not provide the option.

lib/surface/api.ex
msaraiva: Those are leftovers from the old `context` API. Good catch!

lib/surface/api.ex
msaraiva: Hmm. We should have a simple test for it.

This is the order props are listed in the docs. Example:

image

So first, we have all the required props and then the rest, all keeping the order they were defined. This way components authors have a simple way to organize the props, grouping them semantically. Not sure if it's the best way but it's what we have for now. I think we should keep it unless anyone has a better idea.

lib/surface/api.ex
msaraiva: Also leftover from the old `context` API :)

lib/surface/compiler/eex_engine.ex
```diff
@@ -241,39 +241,21 @@ defmodule Surface.Compiler.EExEngine do
...
-            unquote(module),
-            unquote(meta.node_alias)
-          ),
-          unquote(do_block)
-        )
-      end
+          unquote(meta.node_alias)
+        ),
+        unquote(do_block)
+      )
```
msaraiva, line 258: There used to be an issue in `live_component/3` that would emit a warning like `variable assings is not being used...` or something similar if we passed `[]` as the last argument. With the above change, you should see that warning back if you have a component without a body like `<Component />`, unless they have fixed it :)

Malian: It seems that the bug has been fixed. I am not able to reproduce it.

lib/surface/api.ex
Malian: Ok, it makes sense to order the props. From what I have seen from other tools (Vuetify, Sotrybook, Material-Ui) the API is always sorted by name and by category (props, slots, events, ...). If the prop/slot is required, the API shows a `*` or __required__. 

I suggest keeping the sorting as is. If we have feedback later, it will be easy to define a new sorting method. I'll revert these changes.

Malian: I reverted the changes and I extracted the sorting into its function.

msaraiva (approved):
❤️

PR #294, opened 2021-03-13
Fix copying JS hooks multiple times

### LINKED ISSUE #291: Co-located JS hook files being copied on each page load in development
I am using a few co-located hooks in Surface v0.3.1, and it appears that loading the page via live_redirect or full reload always copies the hook files to the assets directory (because it is re-compiling?). 

This is triggering webpack to rebuild assets despite the hook file(s) not having changed, but I am guessing it was copied again. 

The project is leveraging webpack5, which might be a factor - but don't think it should be 🤷

msaraiva:
Fixes #291.


test/mix/tasks/compile/surface_test.exs
```diff
@@ -86,6 +86,61 @@ defmodule Mix.Tasks.Compile.SurfaceTest do
...
+    assert index_file_time < src_hooks_file_time
+    assert dest_hooks_file_time < src_hooks_file_time
+
+    run([])
+
+    %File.Stat{mtime: index_file_time} = File.stat!(index_file)
+    %File.Stat{mtime: dest_hooks_file_time} = File.stat!(dest_hooks_file)
+
+    assert index_file_time > src_hooks_file_time
+    assert dest_hooks_file_time > src_hooks_file_time
```
msaraiva, line 117: I don't like to have to touch the files to be able to check if were updated but since `File.stat!` returns the modification time in seconds, I couldn't find a better way.

msaraiva:
@lnr0626 no need to review this now. I'll change those tests to use `file_system`.

PR #296, opened 2021-03-14
Warn if slot is required and has a default value

### LINKED ISSUE #295: Warn if slot is required and has default value
The following example show a header slot that is required and that has a default value.
```elixir
defmodule ComponentWithRequiredSlots do
    use Surface.Component

    slot default
    slot header, required: true

    def render(assigns) do
      ~H"""
      <div>
        <slot name="header">Header</slot>
        <slot/>
      </div>
      """
    end
  end
```
As we did for #281 we could warn if the slot is required and as a default value.
```bash
setting a default value on a required slot has no effect. Either set the default value `<slot name="header">Default</slot>` or set the slot as required, slot header, required: true`, but not both.",
```

Malian:
Fixes #295 

The compiler will generate a proper message if a slot is required and if it has a default value.

The warning messages reflect the usage of the short syntax `<slot>...</slot>` or the full syntax `<slot name="header">...</slot>`

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -862,7 +862,7 @@ defmodule Surface.Compiler do
       slot_name_tag = if short_syntax?, do: "", else: " name=\"#{slot.name}\""
 
       message = """
-      setting a fallback content on a required slot has no effect.
+      setting the fallback content on a required slot has no effect.
 
       Hint: Either keep the fallback content and remove the `required: true`:
 
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -862,14 +862,18 @@ defmodule Surface.Compiler do
       slot_name_tag = if short_syntax?, do: "", else: " name=\"#{slot.name}\""
 
       message = """
-      setting a default value on a required slot has no effect.
+      setting a fallback content on a required slot has no effect.
 
-      Hint: Either set the default value
+      Hint: Either keep the fallback content and remove the `required: true`:
+
         slot #{slot.name}
-        <slot#{slot_name_tag}>...</slot>
+        ...
+        <slot#{slot_name_tag}>Fallback content</slot>
 
-      or set the slot as required,
+      or keep the slot as required and remove the fallback content:
+
         slot #{slot.name}, required: true`
+        ...
         <slot#{slot_name_tag} />
 
       but not both.
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -284,20 +284,23 @@ defmodule Surface.Compiler do
   defp convert_node_to_ast(:slot, {_, attributes, children, node_meta}, compile_meta) do
     meta = Helpers.to_meta(node_meta, compile_meta)
 
-    defined_slot_names =
+    defined_slots =
       meta.caller.module
       |> Surface.API.get_slots()
-      |> Enum.map(& &1.name)
 
     # TODO: Validate attributes with custom messages
     name = attribute_value(attributes, "name", :default)
+    short_slot_syntax? = not has_attribute?(attributes, "name")
 
     index =
       attribute_value_as_ast(attributes, "index", %Surface.AST.Literal{value: 0}, compile_meta)
 
     with {:ok, directives, _attrs} <-
            collect_directives(@slot_directive_handlers, attributes, meta),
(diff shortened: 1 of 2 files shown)


lib/surface/compiler.ex
```diff
@@ -284,20 +284,23 @@ defmodule Surface.Compiler do
   defp convert_node_to_ast(:slot, {_, attributes, children, node_meta}, compile_meta) do
     meta = Helpers.to_meta(node_meta, compile_meta)
 
-    defined_slot_names =
+    defined_slots =
       meta.caller.module
       |> Surface.API.get_slots()
-      |> Enum.map(& &1.name)
 
     # TODO: Validate attributes with custom messages
```
Malian, line 291: @msaraiva What did you have in mind with this TODO?

test/compiler_test.exs
```diff
@@ -905,6 +905,70 @@ defmodule Surface.CompilerSyncTest do
...
+             slot default
+             <slot>...</slot>
+
+           or set the slot as required,
+             slot default, required: true`
+             <slot />
+
+           but not both.
+
+             code.exs:11: Surface.CompilerSyncTest.TestComponent_#{id}.render/1
```
msaraiva, line 950: I think we should use "fallback content" instead of  "default value" to make the terminology more consistent with the docs in https://surface-ui.org/slots. Like:

```
           setting a fallback content on a required slot has no effect.

           Hint: Either keep the fallback content and remove the `required: true`:

             slot default
             ...
             <slot>Fallback content</slot>

           or keep the slot as required and remove the fallback content:

             slot default, required: true
             ...
             <slot />
```

WDYT?

lib/surface/compiler.ex
Malian: Indeed it results in an internal error as the compiler tries to convert `{:attribute_expr, " :slot_name ", %{line: 2}}` to an atom.

```console
== Compilation error in file test/xxx.exs ==
** (ArgumentError) argument error
    :erlang.binary_to_atom({:attribute_expr, " :body ", %{line: 2}}, :utf8)
    (surface 0.3.1) lib/surface/compiler.ex:485: anonymous fn/2 in Surface.Compiler.attribute_value/3
    (elixir 1.11.0) lib/enum.ex:3411: Enum.find_value_list/3
    (surface 0.3.1) lib/surface/compiler.ex:292: Surface.Compiler.convert_node_to_ast/3
    (surface 0.3.1) lib/surface/compiler.ex:212: anonymous fn/3 in Surface.Compiler.to_ast/2
    (elixir 1.11.0) lib/enum.ex:2181: Enum."-reduce/3-lists^foldl/2-0-"/3
    (surface 0.3.1) lib/surface/compiler.ex:211: Surface.Compiler.to_ast/2
    (surface 0.3.1) lib/surface/compiler.ex:331: Surface.Compiler.convert_node_to_ast/3
    (surface 0.3.1) lib/surface/compiler.ex:212: anonymous fn/3 in Surface.Compiler.to_ast/2
    (elixir 1.11.0) lib/enum.ex:2181: Enum."-reduce/3-lists^foldl/2-0-"/3
    (surface 0.3.1) lib/surface/compiler.ex:211: Surface.Compiler.to_ast/2
    (surface 0.3.1) lib/surface/compiler.ex:115: Surface.Compiler.compile/5
    (surface 0.3.1) expanding macro: Surface.sigil_H/2
    test/slot_test.exs:72: Surface.SlotTest.OuterWithNamedSlotAndProps.render/1
    (elixir 1.11.0) lib/kernel/parallel_compiler.ex:416: Kernel.ParallelCompiler.require_file/2
```

lib/surface/compiler.ex
```diff
@@ -853,4 +854,32 @@ defmodule Surface.Compiler do
...
   end
+
+  defp maybe_warn_required_slot_with_default_value(_, [], _, _), do: nil
+
+  defp maybe_warn_required_slot_with_default_value(slot, _, short_syntax?, meta) do
+    if Keyword.get(slot.opts, :required, false) do
+      slot_name_tag = if short_syntax?, do: "", else: " name=\"#{slot.name}\""
+
+      message = """
+      setting a fallback content on a required slot has no effect.
```
lnr0626, line 865: Small nitpick, this should be `the fallback content`

msaraiva (approved):
Ship it!

PR #297, opened 2021-03-17
Update props type according to new types

Malian:
Update props type according to new type inserted in #286

We should also update surface_catalogue & surface_website according to these new types.


miguel-s:
Love seeing that `:struct` type instead of `:any` or `:map`!

msaraiva (approved):

msaraiva:
> We should also update surface_catalogue & surface_website according to these new types.

@Malian I created the issues for both so we don't forget after this gets merged.

PR #299, opened 2021-03-20
Generate doc for slots and events

### LINKED ISSUE #293: Generated documentation should also exposes doc for slot
Currently, Surface only generates the documentation for props. We should add proper documentation for `slot` as it is part of the component API.

As `data` is private, I believe we should not generate doc for `data`.

To add documentation for `slot` we should update this part of the code:

https://github.com/surface-ui/surface/blob/2ec6c816e388f842a0254415831250908dc4499f/lib/surface/api.ex#L458-L488

Malian:
Fixes #293 

Here is an output example with the `Form` component:
image

### DIFF
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -73,7 +73,7 @@ defmodule Surface do
     * `Surface.LiveView` - A wrapper component around `Phoenix.LiveView`.
     * `Surface.MacroComponent` - A low-level component which is responsible for translating its own content at compile time.
 
-  ### Example
+  ## Example
 
       # A functional stateless component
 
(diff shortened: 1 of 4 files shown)


Malian:
@msaraiva I'll update the surface_site and surface_catalogue according to this change if the PR is accepted.

msaraiva:
@Malian the code looks good to me! One thing I just realized is that we could change the heading type from `###` to `##` so it would appear in the sidebar under the "Sections" item. Example from Ecto:

image

image

WDYT?

Malian:
> WDYT?

@msaraiva  I believe it's a good idea!

Here is an example with the `Form` component:
image

I'll update the code.

msaraiva (approved):

PR #300, opened 2021-03-21
Add Button component

miguel-s:
Adds the `<Button />` component.

Based on Phoenix's `button/2` https://hexdocs.pm/phoenix_html/Phoenix.HTML.Link.html#button/2.

### DIFF
--- a/lib/surface/components/button.ex
+++ b/lib/surface/components/button.ex
@@ -1,9 +1,10 @@
 defmodule Surface.Components.Button do
   @moduledoc """
-  Generates a link to the given URL.
+  Generates a button that uses a regular HTML form to submit to the given URL.
 
-  Provides similar capabilities to Phoenix's built-in `button/2`
-  function.
+  Useful to ensure that links that change data are not triggered by search engines and other spidering software.
+
+  Provides similar capabilities to Phoenix's built-in `button/2` function.
 
   Options `label` and `class` can be set directly and will override anything in `opts`.
   All other options are forwarded to the underlying <button> tag.
--- a/lib/surface/components/button.ex
+++ b/lib/surface/components/button.ex
@@ -75,7 +75,7 @@ defmodule Surface.Components.Button do
   slot default
 
   def update(assigns, socket) do
-    validate_label!(assigns)
+    valid_label!(assigns)
     {:ok, assign(socket, assigns)}
   end
 
@@ -89,7 +89,7 @@ defmodule Surface.Components.Button do
     """
   end
 
-  defp validate_label!(assigns) do
+  defp valid_label!(assigns) do
     unless assigns[:default] || assigns[:label] || Keyword.get(assigns.opts, :label) do
       raise ArgumentError, "<Button /> requires a label prop or contents in the default slot"
     end
--- a/lib/surface/components/button.ex
+++ b/lib/surface/components/button.ex
@@ -82,7 +82,7 @@ defmodule Surface.Components.Button do
 
   def update(assigns, socket) do
     unless assigns[:default] || assigns[:label] || Keyword.get(assigns.opts, :label) do
-      raise ArgumentError, "<Link /> requires a label prop or contents in the default slot"
+      raise ArgumentError, "<Button /> requires a label prop or contents in the default slot"
     end
 
     {:ok, assign(socket, assigns)}
--- a/lib/surface/components/button.ex
+++ b/lib/surface/components/button.ex
@@ -29,6 +29,8 @@ defmodule Surface.Components.Button do
 
   use Surface.Component
 
+  import Surface.Components.Utils
+
   @doc "The page to link to"
   prop to, :any, required: true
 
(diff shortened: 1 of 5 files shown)


lib/surface/components/button.ex
```diff
@@ -0,0 +1,183 @@
...
+  @doc "Class or classes to apply to the button"
+  prop class, :css_class
+
+  @doc """
+  The label for the generated `<button>` element, if no content (default slot) is provided.
+  """
+  prop label, :string
+
+  @doc "Triggered on click"
+  prop click, :event
```
Malian, line 50: Shouldn't we add the others usual events like the ones we find in the `Surface.Components.Form.Submit` component?
```elixir
      prop blur, :event

      prop focus, :event

      @doc "Triggered when the component receives click"
      prop capture_click, :event

      @doc "Triggered when a button on the keyboard is pressed"
      prop keydown, :event

      @doc "Triggered when a button on the keyboard is released"
      prop keyup, :event
```
Moreover, WDYT of renaming `click` to `capture_click` to be consistent in the whole codebase? In that case, we will also modify the `click` event in the `Link` component.

lib/surface/components/link.ex
```diff
@@ -2,8 +2,7 @@ defmodule Surface.Components.Link do
   @moduledoc """
   Generates a link to the given URL.
 
-  Provides similar capabilities to Phoenix's built-in `link/2`
-  function.
+  Provides similar capabilities to Phoenix's built-in `link/2` function.
```
Malian, line 5:
```suggestion
  Provides similar capabilities to Phoenix's built-in `button/2` function.
```

test/components/button_test.exs
```diff
@@ -0,0 +1,214 @@
...
+
+    test "button with invalid args" do
+      msg = """
+      unsupported scheme given to <Button />. In case you want to link to an
+      unknown or unsafe scheme, such as javascript, use a tuple: {:javascript, rest}
+      """
+
+      assert_raise ArgumentError, msg, fn ->
+        render_surface(do: ~H[<Button label="foo" to="javascript:alert(1)" method={{ :get }} />])
+        |> IO.inspect()
```
Malian, line 210:
```suggestion
```

Malian (commented):
Thanks @miguel-s 

Except the few comments below, the code looks good to me. I am just wondering if we could extract some of the code to a private API and share it between the `Link` and the `Button` components as 95% of the code is the same.

test/components/button_test.exs
miguel-s: Good catch!

lib/surface/components/button.ex
```diff
@@ -0,0 +1,183 @@
...
+    "rtsp:",
+    "svn:",
+    "tel:",
+    "fax:",
+    "xmpp:"
+  ]
+
+  def update(assigns, socket) do
+    unless assigns[:default] || assigns[:label] || Keyword.get(assigns.opts, :label) do
+      raise ArgumentError, "<Link /> requires a label prop or contents in the default slot"
```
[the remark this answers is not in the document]
miguel-s: Good catch!

lib/surface/components/link.ex
miguel-s: Wrong file or am I not seeing something?

lib/surface/components/button.ex
miguel-s: Yes makes sense, somehow missed these changes.

It seems that in `Surface.Components.Form.Submit` these events are defined, but never actualy used. In `Surface.Components.Form.Reset` they're are defined and used. Is there a reason for this or just an oversight?

miguel-s:
> I am just wondering if we could extract some of the code to a private API and share it between the `Link` and the `Button` components as 95% of the code is the same.

Yes planning on doing just that :)

Waiting on some changes that @msaraiva is planning for Phoenix that will allow us to also clean up all the `opts_to_attrs` logic.

lib/surface/components/link.ex
Malian: Forget this suggestion. I thought I deleted it.

lib/surface/components/button.ex
msaraiva: > It seems that in `Surface.Components.Form.Submit` these events are defined, but never actualy used. In `Surface.Components.Form.Reset` they're are defined and used. Is there a reason for this or just an oversight?

Probably an oversight.

msaraiva:
> Waiting on some changes that @msaraiva is planning for Phoenix that will allow us to also clean up all the `opts_to_attrs` logic.

@miguel-s those changes were applied to `phoenix_html` and are already on `master`, however, we have to wait for a new phoenix release that includes the upcoming version of `phoenix_html`. The new function is `Phoenix.HTML.Tag.attributes_escape/1`. More details can be found on the PR: https://github.com/phoenixframework/phoenix_html/pull/331.

Malian (approved):

msaraiva (approved):

PR #305, opened 2021-03-22
[WIP] experiment with allowing context passed to functions

lnr0626:
it works as is, but it's also not great and requiring users to mess around with the internal assigns manually isn't a great idea imo


msaraiva:
Yeah, having users messing with private stuff should be avoided at all costs. Besides, we'll be able to provide a nice solution very soon when we introduce the new `component` macro in LV which will be called under the hood when using something like:

```
<#my_func />
```

By calling a function with that syntax will not only enable diff tracking for the returned content but will also allow us to decorate the underlying call to the `component` macro, just like we currently do for the `live_comonent` call.

Issue #298, opened 2021-03-19
Include a wrapper for `live_file_input`, and possibly API for allowing uploads via prop/data macro


harmon25:
Surface currently includes a `FileInput` wrapper component for the phoenix `file_input` html helper. It would be nice if Surface also included a wrapper for the LiveView `live_file_input` html helper.

Here is a small prototype that should work.

```elixir
defmodule LiveFileInput do
  use Surface.Component
  @doc "Upload specified via `allow_upload`"
  prop upload, :any, required: true
  @doc "Classes to be used on the generated `input` element"
  prop class, :css_class, default: []
  @doc "Other DOM attributes to be passed to the `input` element" 
  prop opts, :keyword, default: []
   
  def render(assigns) do
    ~H"{{ live_file_input(@upload, [class: @class] ++ @opts) }}"
  end
end
```

Usage: 
(assumes `allow_upload` has been called in parent lv or live_component, with the `:component` key)
```elixir
~H"""
<LiveFileInput upload={{ @uploads.component }}  />
"""
```

Also curious if there might be room for a special `data` or `prop` API that would be able to take care of allowing uploads:
```elixir
socket = allow_upload(socket, :component, accept: ~w(.json), max_entries: 1)
```

Concept:
```elixir
prop, :file, :upload, accept: ~w(.json), max: 1

#AND/OR

data, :file, :upload, accept: ~w(.json), max: 1
```
Thoughts?

miguel-s:
The prototype looks good to me! Do you want to go ahead and open a PR?

We can explore the prop/data concept separately.

msaraiva:
Hi @harmon25!

> Also curious if there might be room for a special `data` or `prop` API that would be able to take care of allowing uploads:

Yes. There's certainly room for that! If we create a new type `:upload` and allow:

```elixir
data avatar, :upload, accept: ~w(.json), max_entries: 1
```

we could call `allow_upload` automatically so you wouldn't have to implement `mount`. Then you could just:

```jsx
<Form>
  ...
  <LiveFileInput upload={{ @avatar }}/>
</Form>

```

Malian:
@harmon25 & @msaraiva please find a first implementation of an `:upload` type for `data` here #311

msaraiva:
@harmon25 do we want to change the API to accept:

```jsx
<LiveFileInput upload={{ @avatar }}/>
```

instead of:

```jsx
<LiveFileInput upload={{ @uploads.avatar }}/>
```
?

[...]

PR #307, opened 2021-03-22
add live_file_input component & test

harmon25:
Initial PR for #298 

Test does not do much but verify it renders the expected phoenix generated content. Should we do anything else?

Happy to iterate a bit on this, feedback is very appreciated!

### DIFF
--- a/lib/surface/components/live_file_input.ex
+++ b/lib/surface/components/live_file_input.ex
@@ -28,8 +28,8 @@ defmodule Surface.Components.LiveFileInput do
   prop class, :css_class
 
   @doc """
-  Extra options for `live_file_input/3`.
-  See `Phoenix.LiveView.Helpers.live_file_input/3` for the available options.
+  Extra options for `live_file_input/2`.
+  See `Phoenix.LiveView.Helpers.live_file_input/2` for the available options.
   """
   prop opts, :keyword, default: []
 
--- a/lib/surface/components/live_file_input.ex
+++ b/lib/surface/components/live_file_input.ex
@@ -7,22 +7,35 @@ defmodule Surface.Components.LiveFileInput do
 
   ```elixir
     def mount(socket) do
-      socket = allow_upload(socket, :test, accept: ~w(.json), max_entries: 1)
+      socket = allow_upload(socket, :avatar, accept: ~w(.json), max_entries: 1)
       {:ok, socket}
     end
   ```
 
   See Phoenix.LiveView [Uploads documentation](https://hexdocs.pm/phoenix_live_view/uploads.html#content)
   """
+
   use Surface.Component
+  import Surface.Components.Form.Utils, only: [props_to_attr_opts: 2]
+
   @doc "Upload specified via `allow_upload`"
-  prop upload, :any, required: true
-  @doc "Classes to be used on the generated `input` element"
-  prop class, :css_class, default: []
-  @doc "Other DOM attributes to be passed to the `input` element"
+  prop upload, :struct, required: true
+
+  @doc "The id of the corresponding input field"
+  prop id, :string
+
+  @doc "The CSS class for the generated `<input>` element"
+  prop class, :css_class
+
+  @doc """
+  Extra options for `live_file_input/3`.
+  See `Phoenix.LiveView.Helpers.live_file_input/3` for the available options.
+  """
   prop opts, :keyword, default: []
 
   def render(assigns) do
-    ~H"{{ live_file_input(@upload, [class: @class] ++ @opts) }}"
+    attr_opts = props_to_attr_opts(assigns, [:id, class: get_config(:default_class)])
+
+    ~H"{{ live_file_input(@upload, attr_opts ++ @opts) }}"
   end
 end
--- a/lib/surface/components/live_file_input.ex
+++ b/lib/surface/components/live_file_input.ex
@@ -16,7 +16,7 @@ defmodule Surface.Components.LiveFileInput do
   """
 
   use Surface.Component
-  import Surface.Components.Form.Utils, only: [props_to_attr_opts: 2]
+  import Surface.Components.Form.Utils, only: [props_to_attr_opts: 2, events_to_opts: 1]
 
   @doc "Upload specified via `allow_upload`"
   prop upload, :struct, required: true
@@ -27,9 +27,25 @@ defmodule Surface.Components.LiveFileInput do
   @doc "Keyword list with options to be passed down to `live_file_input/2`"
   prop opts, :keyword, default: []
 
+  @doc "Triggered when the component loses focus"
+  prop blur, :event
+
+  @doc "Triggered when the component receives focus"
+  prop focus, :event
+
+  @doc "Triggered when the component receives click"
+  prop capture_click, :event
+
+  @doc "Triggered when a button on the keyboard is pressed"
+  prop keydown, :event
+
+  @doc "Triggered when a button on the keyboard is released"
+  prop keyup, :event
+
   def render(assigns) do
     attr_opts = props_to_attr_opts(assigns, class: get_config(:default_class))
+    event_opts = events_to_opts(assigns)
 
-    ~H"{{ live_file_input(@upload, attr_opts ++ @opts) }}"
+    ~H"{{ live_file_input(@upload, attr_opts ++ @opts ++ event_opts) }}"
   end
 end
--- a/lib/surface/components/live_file_input.ex
+++ b/lib/surface/components/live_file_input.ex
@@ -1,0 +1,28 @@
+defmodule Surface.Components.LiveFileInput do
+  @moduledoc """
+  Wrapper around Phoenix LiveView's built-in
+  `live_file_input/2` function.
+
+  The parent LiveView, or LiveComponent must `allow_uploads` during mount:
+
+  ```elixir
+    def mount(socket) do
+      socket = allow_upload(socket, :test, accept: ~w(.json), max_entries: 1)
+      {:ok, socket}
+    end
+  ```
+
+  See Phoenix.LiveView [Uploads documentation](https://hexdocs.pm/phoenix_live_view/uploads.html#content)
+  """
+  use Surface.Component
+  @doc "Upload specified via `allow_upload`"
+  prop upload, :any, required: true
+  @doc "Classes to be used on the generated `input` element"
+  prop class, :css_class, default: []
+  @doc "Other DOM attributes to be passed to the `input` element"
+  prop opts, :keyword, default: []
+
+  def render(assigns) do
+    ~H"{{ live_file_input(@upload, [class: @class] ++ @opts) }}"
+  end
+end
--- a/lib/surface/components/live_file_input.ex
+++ b/lib/surface/components/live_file_input.ex
@@ -21,9 +21,6 @@ defmodule Surface.Components.LiveFileInput do
   @doc "Upload specified via `allow_upload`"
   prop upload, :struct, required: true
 
-  @doc "The id of the corresponding input field"
-  prop id, :string
-
   @doc "The CSS class for the generated `<input>` element"
   prop class, :css_class
 
--- a/lib/surface/components/live_file_input.ex
+++ b/lib/surface/components/live_file_input.ex
@@ -28,7 +28,7 @@ defmodule Surface.Components.LiveFileInput do
   prop opts, :keyword, default: []
 
   def render(assigns) do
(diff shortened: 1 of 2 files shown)


lib/surface/components/live_file_input.ex
```diff
@@ -0,0 +1,28 @@
...
+    end
+  ```
+
+  See Phoenix.LiveView [Uploads documentation](https://hexdocs.pm/phoenix_live_view/uploads.html#content)
+  """
+  use Surface.Component
+  @doc "Upload specified via `allow_upload`"
+  prop upload, :any, required: true
+  @doc "Classes to be used on the generated `input` element"
+  prop class, :css_class, default: []
```
Malian, line 21: Usually we do not give a default value to the `class` prop. It is nil by default.

lib/surface/components/live_file_input.ex
```diff
@@ -0,0 +1,28 @@
...
+  @doc "Other DOM attributes to be passed to the `input` element"
+  prop opts, :keyword, default: []
+
+  def render(assigns) do
+    ~H"{{ live_file_input(@upload, [class: @class] ++ @opts) }}"
```
Malian, line 26: As we should have the most reusable component across project we usually provide a way to define a `default_class` to Surface component through the config.

Take a look at the other components to know how to do that. For example, the [ColorInput](https://github.com/surface-ui/surface/blob/master/lib/surface/components/form/color_input.ex) component

test/components/live_file_input_test.exs
```diff
@@ -0,0 +1,38 @@
...
+    data uploads, :map
+
+    def mount(socket) do
+      socket = allow_upload(socket, :test, accept: ~w(.json), max_entries: 1)
+      {:ok, socket}
+    end
+
+    def render(assigns) do
+      ~H"""
+        <LiveFileInput upload={{@uploads.test}}/>
```
Malian, line 18:
```suggestion
        <LiveFileInput upload={{ @uploads.test }} />
```

lib/surface/components/live_file_input.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Surface.Components.LiveFileInput do
+  @moduledoc """
+  Wrapper around Phoenix LiveView's built-in
+  `live_file_input/2` function.
+
+  The parent LiveView, or LiveComponent must `allow_uploads` during mount:
+
+  ```elixir
+    def mount(socket) do
+      socket = allow_upload(socket, :test, accept: ~w(.json), max_entries: 1)
```
Malian, line 10:
```suggestion
      socket = allow_upload(socket, :avatar, accept: ~w(.json), max_entries: 1)
```

Just to be consistent with the doc ;)

Malian (changes requested):
@harmon25 Thanks for the pull request.

I let a few comments regarding the code and below you will find more general comments:
- I believe we should add a `name` prop to be consistent with all input we defined in Surface. See the `component/form/input.ex` file.
- Regarding the test, it's ok like this. We could just add a test to test that if a class is passed through the prop, it's correctly rendered, and a test that if no class is passed through the prop and if a default class is set in the config, it's also correctly rendered.

WDYT?

harmon25:
Thanks @Malian!
I made most of the changes, will add more tests!

> I believe we should add a name prop to be consistent with all input we defined in Surface. See the component/form/input.ex file.

`live_file_input` being a liveview helper - it does not behave like a phoenix form helper. The name is being assigned based on the atom passed into `allow_upload` and is ignored if passed in as an opt. It does appear to accept an ID opt- which otherwise is being set by LiveView to the same value as `data-phx-upload-ref`.  (not sure if this will break the live_upload, need to test more)

Here is a sample of the rendered output: 

```html
<input accept=\".json\" class=\"test_class disabled_test\" data-phx-active-refs=\"\" data-phx-done-refs=\"\" data-phx-preflighted-refs=\"\" data-phx-update=\"ignore\" data-phx-upload-ref=\"phx-Fm8ZSnh8_rSc_RWH\" data-test=\"test-data\" id=\"phx-Fm8ZSnh8_rSc_RWH\" name=\"test\" phx-hook=\"Phoenix.LiveFileUpload\" type=\"file\"></input>\n
```

I am going to test it out a bit more in an app, and try writing a test using [render_upload](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#render_upload/3) among the other tests you suggested.

harmon25:
on a closer look at the `live_file_input/2` docs, passing ID should be OK.

> By default, the file input ID is the upload ref...

lib/surface/components/live_file_input.ex
```diff
@@ -0,0 +1,41 @@
...
+  """
+
+  use Surface.Component
+  import Surface.Components.Form.Utils, only: [props_to_attr_opts: 2]
+
+  @doc "Upload specified via `allow_upload`"
+  prop upload, :struct, required: true
+
+  @doc "The id of the corresponding input field"
+  prop id, :string
```
Malian, line 25: This will not work anymore with PhoenixLiveView `0.15.2`, see the following commit: https://github.com/phoenixframework/phoenix_live_view/commit/c263db32f446e57275f62f507f4995fef59c1975 we could simply remove it.

lib/surface/components/live_file_input.ex
```diff
@@ -0,0 +1,41 @@
...
+
+  @doc "The CSS class for the generated `<input>` element"
+  prop class, :css_class
+
+  @doc """
+  Extra options for `live_file_input/2`.
+  See `Phoenix.LiveView.Helpers.live_file_input/2` for the available options.
+  """
```
Malian, line 33: After a close look at the `live_file_input/2` function, it seems that this method does not accept any options and these options are simply forwarded to the underlying `content_tag` function that generates the `input`. I suggest to refrasing to something more general:

```suggestion 
  @doc "Keyword list with options to be passed down to `live_file_input/2`"
```

lib/surface/components/live_file_input.ex
```diff
@@ -0,0 +1,41 @@
...
+  prop opts, :keyword, default: []
+
+  def render(assigns) do
+    attr_opts = props_to_attr_opts(assigns, [:id, class: get_config(:default_class)])
```
Malian, line 37:
```suggestion
    attr_opts = props_to_attr_opts(assigns, [class: get_config(:default_class)])
```

test/components/live_file_input_test.exs
```diff
@@ -0,0 +1,102 @@
...
+
+    def mount(socket) do
+      # the second param passed to allow_upload becomes the value of the inputs html `name` attribute
+      socket = allow_upload(socket, :avatar, accept: ~w(.json), max_entries: 1)
+      {:ok, socket}
+    end
+
+    def render(assigns) do
+      ~H"""
+        <LiveFileInput id="test_id" upload={{ @uploads.avatar }} class={{ "test_class", disabled_test: true  }} opts={{ "data-test": "test-data", name: "a name?" }} />
```
Malian, line 19:
```suggestion
        <LiveFileInput upload={{ @uploads.avatar }} class={{ "test_class", disabled_test: true  }} opts={{ "data-test": "test-data", name: "a name?" }} />
```

test/components/live_file_input_test.exs
```diff
@@ -0,0 +1,102 @@
...
+
+    def mount(socket) do
+      # the second param passed to allow_upload becomes the value of the inputs html `name` attribute
+      socket = allow_upload(socket, :avatar, accept: ~w(.json), max_entries: 1)
+      {:ok, socket}
+    end
+
+    def render(assigns) do
+      ~H"""
+        <LiveFileInput id="test_id" upload={{ @uploads.avatar }} />
```
Malian, line 37:
```suggestion
        <LiveFileInput upload={{ @uploads.avatar }} />
```

test/components/live_file_input_test.exs
```diff
@@ -0,0 +1,102 @@
...
+    html =
+      render_surface do
+        ~H"""
+        <LiveFileInputWithoutId id="test" />
+        """
+      end
+
+    assert html =~ "phx-hook=\"Phoenix.LiveFileUpload\""
+    assert html =~ "id=\"phx-"
+  end
```
Malian, line 101:
```suggestion
```

No need due to remove of `id`

test/components/live_file_input_test.exs
```diff
@@ -0,0 +1,102 @@
+defmodule Surface.Components.LiveFileInputTest do
+  use Surface.ConnCase, async: true
```
Malian, line 2: I added a test to use the `:default_class` of the `LiveFileInput` component. That test requires that async must be set to `false`
```suggestion
  use Surface.ConnCase
```

test/components/live_file_input_test.exs
```diff
@@ -0,0 +1,102 @@
...
+      render_surface do
+        ~H"""
+        <LiveFileInputWithoutProps id="test" />
+        """
+      end
+
+    assert !(html =~ "class=\"\"")
+    assert html =~ "phx-hook=\"Phoenix.LiveFileUpload\""
+    assert html =~ "name=\"avatar\""
+  end
```
Malian, line 89:
```suggestion
  test "correctly renders live_file_input/2 with `:default_class` config" do
    using_config LiveFileInput, default_class: "default_class" do
      html =
        render_surface do
          ~H"""
          <LiveFileInputWithoutProps />
          """
        end

      assert html =~ ~r/class="default_class"/
    end
  end
```

test/components/live_file_input_test.exs
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+    # expected lv attrs
+    assert html =~ "phx-hook=\"Phoenix.LiveFileUpload\""
+    assert html =~ "accept=\".json\""
+    assert html =~ "name=\"avatar\""
+    # expected passed through attrs
+    assert html =~ "class=\"test_class disabled_test\""
+    assert html =~ "data-test=\"test-data\""
+    assert html =~ "id=\"test_id\""
```
Malian, line 75:
```suggestion
```

Malian (changes requested):
@harmon25 Thank for the changes, but there are still some small changes to be made and we will merge!

After a close look at the last versions of Live View, the `id` prop will stop work with Phoenix Live View `0.15.2`. After several tests, it's better to remove it right now. Most of my comments are related to this change. It was your comment about the name that tipped me off :+1: 

I also update a test to check if the `default_class` defined in the configuration is correctly used if provided.

I'll update the documentation in #311 to use the `data :avatar, :upload, accept: ...` new type. Thanks again for this suggestion :+1: 

We are close merging this request! Thanks again for your work!

lib/surface/components/live_file_input.ex
harmon25: good catch! 
I'll remove it.

miguel-s:
Thanks for the PR @harmon25 and awesome reviews @Malian!

How does `live_file_input` behave with regular LiveView events like `phx-click` or `phx-blur`? If they support passing in events we should, too.

Example of how we handle events in the regular `file_input` https://github.com/surface-ui/surface/blob/d2f68bd7a6b4ebc5b0ef00b7cdf06f3cce377352/lib/surface/components/form/file_input.ex#L30

harmon25:
Thank you both for helping support Surface! 
I am not sure if that would interfere with the hook that is being used:
https://github.com/phoenixframework/phoenix_live_view/blob/v0.15.4/lib/phoenix_live_view/helpers.ex#L446

Malian:
> How does `live_file_input` behave with regular LiveView events like `phx-click` or `phx-blur`? If they support passing in events we should, too.

@miguel-s   Good catch. Passing events through the second argument of `live_file_input` works perfectly fine, so indeed we should add them as props.

@harmon25 it should be simple as adding the following codes to the component:

https://github.com/surface-ui/surface/blob/d2f68bd7a6b4ebc5b0ef00b7cdf06f3cce377352/lib/surface/components/form/file_input.ex#L30

https://github.com/surface-ui/surface/blob/d2f68bd7a6b4ebc5b0ef00b7cdf06f3cce377352/lib/surface/components/form/input.ex#L32-L45

harmon25:
Added the event props!

harmon25 (commented):
this should be ready!

Malian (approved):
@harmon25 thanks! It should be ready. I am waiting for a decision about #298 before merging it to be sure we do not miss something but normally it's ok :+1:

miguel-s (approved):

miguel-s:
Looking good to me! Thanks @harmon25 ❤️

msaraiva:
@harmon25 thanks! ❤️

PR #309, opened 2021-03-23
inputs: make the index available as slot prop

ouven:
Make the index available as slot prop, when building nested forms with inputs_for.

Use case: add a "delete from list" button to a subform.


msaraiva:
Hey @ouven!

Thanks for the PR. Can you please add a small test using the passed index? Something similar to  https://github.com/surface-ui/surface/blob/master/test/components/form/inputs_test.exs#L8.

ouven:
Sure, sorry for being hasty :)
Great work by the way!

msaraiva:
@ouven no worries and thanks! ❤️

PR #311, opened 2021-03-23
Introduce the upload macro

Malian:
See #298 for the initial discussion.

This is a first implementation. Feedback are welcomed!

Todo:

- [x] Check the usage of the required option `:accept` or raise a `CompileError`
- [x] Validate upload options types
- [x] Add test to check multiple uploads
- [x] Add ability to pass "local" function to `:external` and `:progress` options
- [x] Validate the arity of function passed through `:external` and `:progress` options
- [ ] Wait for #307 to be merged and update it according to these changes
- [ ] ~~Consider to be able to pass a prop `event` to `progress` and `external` option~~
- [ ] Update the doc
- [ ] Update surface_catalogue to take into account the new `upload` macro
- [ ] Update VSCode Surface extension to handle the new `upload` macro

### DIFF
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -101,7 +101,7 @@ defmodule Surface.API do
     build_assign_ast(:data, name_ast, type_ast, opts_ast, __CALLER__)
   end
 
-  @doc "Defines an upload assign for the component"
+  @doc "Defines an upload available in the built in uploads assign for the component"
   defmacro upload(name_ast, opts_ast \\ []) do
     build_assign_ast(:upload, name_ast, :any, opts_ast, __CALLER__)
   end
@@ -120,6 +120,27 @@ defmodule Surface.API do
       line: line
     }
 
+    do_put_assign!(assign, caller)
+  end
+
+  defp do_put_assign!(%{func: :upload} = upload, caller) do
+    # Upload is special assigns in the sense that it is contained in the
+    # `uploads` built in assign. Compare to prop, data and slot,
+    # Surface does not create an assign and so the validation is different.
+    uploads = get_uploads(caller.module)
+    existing_upload = Enum.find(uploads, false, fn %{name: name} -> name == upload.name end)
+
+    if existing_upload do
+      details = existing_assign_details_message(false, existing_upload)
+      message = ~s(cannot use name "#{upload.name}". #{details}.)
+
+      IOHelper.compile_error(message, caller.file, upload.line)
+    end
+
+    Module.put_attribute(caller.module, :upload, upload)
+  end
+
+  defp do_put_assign!(assign, caller) do
     assigns = Module.get_attribute(caller.module, :assigns) || %{}
     name = Keyword.get(assign.opts, :as, assign.name)
     existing_assign = assigns[name]
@@ -132,11 +153,10 @@ defmodule Surface.API do
       message = ~s(cannot use name "#{assign.name}". #{details}.)
 
       IOHelper.compile_error(message, caller.file, assign.line)
-    else
-      assigns = Map.put(assigns, name, assign)
-      Module.put_attribute(caller.module, :assigns, assigns)
     end
 
+    assigns = Map.put(assigns, name, assign)
+    Module.put_attribute(caller.module, :assigns, assigns)
     Module.put_attribute(caller.module, assign.func, assign)
   end
 
@@ -169,9 +189,8 @@ defmodule Surface.API do
       data = if function_exported?(module, :__data__, 0), do: module.__data__(), else: []
       props = if function_exported?(module, :__props__, 0), do: module.__props__(), else: []
       slots = if function_exported?(module, :__slots__, 0), do: module.__slots__(), else: []
-      uploads = if function_exported?(module, :__uploads__, 0), do: module.__uploads__(), else: []
 
-      Enum.map(data ++ props ++ slots ++ uploads, fn %{name: name, line: line} -> {name, line} end)
+      Enum.map(data ++ props ++ slots, fn %{name: name, line: line} -> {name, line} end)
     end
   end
 
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -45,7 +45,8 @@ defmodule Surface.API do
     arities = %{
       prop: [2, 3],
       slot: [1, 2],
-      data: [2, 3]
+      data: [2, 3],
+      upload: [1, 2]
     }
 
     functions = for func <- include, arity <- arities[func], into: [], do: {func, arity}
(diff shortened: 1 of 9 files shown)


sodapopcan:
This is exciting.  Thank you for doing this work.

lib/surface/api.ex
```diff
@@ -395,6 +425,38 @@ defmodule Surface.API do
     [:required, :props, :as]
   end
 
+  defp get_required_opts(:data, :upload, _opts) do
+    [:accept]
+  end
+
+  defp get_required_opts(_, _, _), do: []
+
+  defp validate_opt_ast!(:data, option, args_ast, _caller)
```
Malian, line 434: It works, but I am pretty sure it's not the right way to do :sweat_smile: Moreover, I do not know how to test it.

lib/surface/api.ex
```diff
@@ -433,6 +495,36 @@ defmodule Surface.API do
...
+
+  defp validate_opt(:data, :upload, :external, value) when not is_function(value, 2) do
+    {:error, "invalid value for option :external. Expected a function/2, got: #{inspect(value)}"}
+  end
+
+  defp validate_opt(:data, :upload, :progress, value) when not is_function(value, 3) do
+    {:error, "invalid value for option :progress. Expected a function/3, got: #{inspect(value)}"}
+  end
+
+  for option <- [:max_entries, :max_file_size, :chunk_size, :chunk_timeout] do
```
Malian, line 518: Maybe we are missing some clarity in the code. We could (1) create 1 function for each option or (2) move the list into a guard clause. WDYT?

Malian (commented):
Ready for a first review! Please don't be too hard on me regarding the Macro part :grin:

lib/surface/api.ex
miguel-s: In this case I would go with a guard clause, that should be simple and clear.

I've seen this pattern used in the Phoenix Html codebase like so:

```Elixir
for scheme <- @valid_uri_schemes do
  defp valid_string_destination!(unquote(scheme) <> _ = string, _context), do: string
end

defp valid_string_destination!(to, context) do
  if not match?("/" <> _, to) and String.contains?(to, ":") do
    raise ArgumentError, """
    unsupported scheme given to #{context}. In case you want to link to an
    unknown or unsafe scheme, such as javascript, use a tuple: {:javascript, rest}
    """
  else
    to
  end
end
```

But in this case multiple functions are needed to pattern match on the start of the string `unquote(scheme) <> _ = string`, which can't be done with a guard.

lib/surface/api.ex
msaraiva: This is not gonna work for private functions. We have to escape this value so it can be evaluated only inside the `mount` scope. I'll run some experiments.

lib/surface/component.ex
```diff
@@ -41,7 +41,7 @@ defmodule Surface.Component do
 
       use Surface.BaseComponent, type: unquote(__MODULE__)
 
-      use Surface.API, include: [:prop, :slot, :data]
+      use Surface.API, include: [:prop, :slot, :data, :upload]
```
Malian, line 44: Not sure if the `upload` is relevant for a stateless component. I do not see a real use case where it is useful. If no one is against it, I suggest removing it. We can always add it if someone has a concrete example.

test/api_test.exs
```diff
@@ -86,6 +86,14 @@ defmodule Surface.APITest do
     message = ~r/cannot use name "label". There's already a prop/
     assert_raise(CompileError, message, fn -> eval(code) end)
 
+    code = """
+    prop avatar, :string
+    upload avatar, accept: :any
+    """
+
+    message = ~r/cannot use name "avatar". There's already a prop/
```
Malian, line 94: Does not make sense... This is a remnant of the implementation as a `data`. I'll remove it!

Malian:
@msaraiva @miguel-s @lnr0626 This PR implements a new `upload` macro (see the discussion in #298). The first implementation added an `upload` type that we could use with the data macro: `data avatar, :upload, accept: ...`. That implementation added the avatar assign by copying the one present in the `@uploads.avatar` built in assign. Unfortunately, it has several drawbacks, the main one being that the `@avatar` assign was not updated according to the one stored in `@uploads`, so we lost the progress feature for example.

We decided to give a try to a new `upload` macro that:
- does not create an assign
- execute the `allow_upload` from PhoenixLiveView in the `mount` function if a `upload` is used

My second implementation was simply to create a shortcut with the upload macro so that `upload avatar, ...` was equivalent to `data avatar, :upload, ...`. That implementation came with some drawbacks:
- use `data avatar, ...` was not possible if a prop, slot or data already exist with that name. It does not make much sense because there is no conflict.
- Error message referred to `data` when the user wrote `upload`
- an assign was created

I fixed these issues but it created a lot of complexity, and cognitive overhead to understand that in some case `data` was treated differently. So, I decided to add logic for a new `func`, `:upload`. It result in the current implementation.

To be honest, I am not completely happy with this implementation, especially the part in `put_assign!` and `do_put_assign!` functions: https://github.com/surface-ui/surface/blob/536aeb780b1446a089c33cabdde8fd0b2f7bf882/lib/surface/api.ex#L110-L161

Anyway, I like the idea of a new macro because I think it improves the DX, but I'm not convinced by the implementation, nor by the cost of it.

lib/surface/component.ex
miguel-s: Do you know if LiveView allows `allow_upload` to be used in components (stateful or stateless) or just in the live views? Would suggest keeping it as close as possible to that.

olivermt: allow_upload works in stateful components, using it in a stateless should(would?) just bubble up to the stateful parent imo.

Malian: @olivermt is right. The component does not care about the @uploads assigns. What is important is the event that are send back to the component or the live view.

msaraiva:
From @Malian on slack:

> I am wondering if this new macro is a good idea. We are replacing a simple function `allow_upload(socket, :avatar, opts)` in the mount function, with a macro upload avatar, opts that adds complexity to the Surface base code without real benefit

I confess that I'm starting to agree with that. I see benefits in the declarative approach. No doubt. That's one of the reasons I created Surface in the first place :)

However, I'm wondering if, in this case, it's worth the trouble of adding that extra complexity just to avoid calling `allow_upload/3` on mount.

Maybe the only extra benefit I can think of right now would be the possibility to statically validate properties passed as `@updates.whatever` but it's not a bug deal, IMO.

@Malian, @harmon25, @miguel-s, @olivermt and @lnr0626. I'm bringing this discussion here so we keep whatever decision we make documented.

harmon25:
Sticking with `allow_upload/3` is fine!
The `LiveFileInput` component improves the DX of vanilla `live_uploads` within Surface enough for now.

If more people request something like this in the future maybe we could revisit the idea

msaraiva:
@Malian I agree with @harmon25. I think we should close this for now and maybe revisit later in case we're convinced it's a good idea. The PR itself was already worth it as it gave us the right feeling about the tradeoffs. Thanks! ❤️

Issue #312, opened 2021-03-24
Validate the default value for prop and data against to the given type


Malian:
The current implementation allows the following code:

```elixir
prop title, :string, default: 0
```

We could check that the default value respects the type `string` at compile time and raise a `CompilationError` if the default value does not match the given type.

We should be able to perform this validation for the following types:
   - `:list`
   - `:boolean`
   - `:string`
   - `:time`
   - `:date`
   - `:datetime`
   - `:naive_datetime`
   - `:number`
   - `:integer`
   - `:decimal`
   - `:map`
   - `:fun`
   - `:atom`
   - `:module`
   - `:keyword`
   - `:struct`
   - `:tuple`
   - `:pid`
   - `:port`
   - `:reference`
   - `:bitstring`
   - `:range`
   - `:mapset`
   - `:regex`
   - `:uri`
   - `:path`

yusufyildirim:
It would be great if we could check if the given `default` value exist in the `values` list also.

For example:
`prop weight, :string, values: ["light", "normal", "medium", "semibold", "bold"], default: "something_else"`

This complies and does not throw any error even though the given default value does not exist in the values list.

Malian:
@yusufyildirim good catch! 

I believe we could add this to the `maybe_warn_mutually_exclusive_opts` and rename the method to `maybe_warn` as suggested by @msaraiva in https://github.com/surface-ui/surface/pull/283#discussion_r598364639

A simple check like this should do the trick.
```elixir
values = Keyword.get(opts, :values, nil)
default = Keyword.get(opts, :default, nil)

if default && values && default not in values do
  IOHelper.warn("A proper message", caller, fn _ -> caller.line end)
end
```
Maybe we should handle the case of an `accumulate` prop where the following code is valid and where my quick solution will not work:
```elixir
data count, :integer, default: [], values: [0, 1, 2]
```

Would you like to open a PR?

yusufyildirim:
Hi, @Malian. First of all, thank you to all the maintainers for this project.

I'm pretty new to the language and the ecosystem in general but It looks easy enough to fix and I think I can do that 😄  I'll look into it on the weekend.

Thank you for the tip!

msaraiva:
This will be addressed by LV itself after the new declarative API is released.

PR #314, opened 2021-03-28
Warn if given default value doesn't exist in values list

yusufyildirim:
Hi,

This PR's scope discussed a bit in #312. I've found a couple of cases:

* We should be checking if the given default value exists in the values list (for data and prop and if not accumulated)
Example (should give warning): `prop type, :string, values: ["small", "medium", "large"], default: "x-large"`

* We should be checking if the given default value(s) exist in the value list (only for prop and if accumulated)
Example(should give warning): `prop count, :integer, default: [3], values: [0, 1, 2], accumulate: true`

* Should be valid
 `prop count, :integer, default: [], values: [0, 1, 2], accumulate: true`
 
 There are some more cases but that is the core idea.

I'm an Elixir newbie and this is why I struggled to bring idiomatic Elixir code to the table 🙈I definitely need your feedback and this code definitely needs splitting. Please, guide me through if you have time.

Thank you in advance!

### DIFF
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -408,10 +408,45 @@ defmodule Surface.API do
     {:error, "invalid value for option :required. Expected a boolean, got: #{inspect(value)}"}
   end
 
-  defp validate_opt(:prop, _name, _type, opts, :default, _value, caller) do
+  defp validate_opt(func, _name, _type, opts, :default, _value, caller)
+       when func == :prop or func == :data do
+    values = Keyword.get(opts, :values, nil)
+    default = Keyword.get(opts, :default, nil)
+    accumulate = Keyword.get(opts, :accumulate, false)
+
+    if default && accumulate && !is_list(default) do
+      IOHelper.warn(
+        "invalid value for option :default. Expected a list of values, got: #{inspect(default)}",
+        caller,
+        fn _ -> caller.line end
+      )
+    end
+
+    if default && values && accumulate && is_list(default) do
+      invalid_default_values = Enum.filter(default, fn element -> element not in values end)
+
+      if length(invalid_default_values) > 0 do
+        IOHelper.warn(
+          "given default value(s) doesn't exist in the :values list",
+          caller,
+          fn _ -> caller.line end
+        )
+      end
+    end
+
+    if default && values && !accumulate && default not in values do
+      IOHelper.warn(
+        "given default value doesn't exist in the :values list",
+        caller,
+        fn _ -> caller.line end
+      )
+    end
+
     if Keyword.get(opts, :required, false) do
       IOHelper.warn(
-        "setting a default value on a required prop has no effect. Either set the default value or set the prop as required, but not both.",
+        "setting a default value on a required #{func} has no effect. Either set the default value or set the #{
+          func
+        } as required, but not both.",
         caller,
         fn _ -> caller.line end
       )
--- a/test/properties_test.exs
+++ b/test/properties_test.exs
@@ -664,4 +664,50 @@ defmodule Surface.PropertiesSyncTest do
              code.exs:8:\
            """
   end
+
+  test "warn if given default value doesn't exist in values list" do
+    id = :erlang.unique_integer([:positive]) |> to_string()
+    module = "TestComponentWithDefaultValueThatDoesntExistInValues_#{id}"
+
+    code = """
+    defmodule #{module} do
+      use Surface.Component
+
+      prop type, :string, values: ["small", "medium", "large"], default: "x-large"
+      data data_type, :string, values: ["small", "medium", "large"], default: "x-large"
+      prop count, :integer, default: [], values: [0, 1, 2]
+      prop count_acc, :integer, default: [3], values: [0, 1, 2], accumulate: true
+
+      def render(assigns) do
+        ~H""
+      end
+    end
+    """
+
+    output =
+      capture_io(:standard_error, fn ->
+        {{:module, _, _, _}, _} =
+          Code.eval_string(code, [], %{__ENV__ | file: "code.exs", line: 1})
+      end)
+
+    assert output =~ ~r"""
+           given default value doesn't exist in the :values list
+             code.exs:4:\
+           """
+
+    assert output =~ ~r"""
+           given default value doesn't exist in the :values list
+             code.exs:5:\
+           """
+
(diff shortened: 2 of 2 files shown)


msaraiva:
Hi @yusufyildirim!

Thanks for the PR. Before moving on with this change, we need to decide how strict the `values` option will be. Currently, those values are mostly for docs and tooling, e.g. the catalogue, code complete, etc. There's no validation which means the user can set any value of the same type. This might be the desired behaviour. For instance, if you define a prop, let's say `color`, you might want to provide a set of predefined colours like `red`, `blue`, `black`, whatever, but leave the possibility for the user to set any other valid colour, like: `#00DD00` or `#FFFF00`. So we can't simply always emit a warning. We need alternatives. Here are some for consideration:

1. Create an additional, mutual exclusive option in regard to `values`, let's say `suggestions`. The difference is that unlike `values `, `suggestions` are not validated.
2. Extend the syntax of `values` to accept something like: `values: ["red", "green", "blue', ...]`. The 3 dots, indicates that there are other valid values so we don't warn.

Any other alternative we should consider?

miguel-s:
I always assumed `values` were strict and would be validated. All my use-cases have fallen into this category so far, but you make a very good point with the colours example.

I like option 1, the naming is clear enough to me. I'm surprised option 2 is even possible and I believe it would cause confusion, as it's not very standard Elixir.

Another option I can think of is to have a boolean `strict` option that toggles validation in `values` (and maybe others, too?).

msaraiva:
> I'm surprised option 2 is even possible and I believe it would cause confusion, as it's not very standard Elixir.

@miguel-s that notation is valid Elixir code. That syntax is mostly used to define type specs. For example, the following type is valid:

```
@type my_list :: [integer, ...]
```

Where `[integer, ...]` is just a shorthand for `nonempty_list(integer)`

However, I agree it might be confusing, although it's very handy :)

yusufyildirim:
I didn't think that way 😄 I assumed `values` should be validated too.  Elixir is a dynamic language and the current behavior is pretty reasonable. But I think having some sort of `strict` option would be helpful too. Introducing a new option like `strict: true` may not be that handy though. I'm okay with `...` notation or something similar that doesn't introduce a new option to add everywhere I put `:values` 😄

msaraiva:
Another option would be to keep `values` and add `values!`. The latter, obviously, validate it :)

msaraiva:
Alright. Time to vote!

`Strict` and `non-strict` versions:

1. `values` and `suggestions`
2. `values` and `values!`
3. `values: ["v1", "v2"]` and `values: ["v1", "v2", ...]`

/cc @surface-ui/core

yusufyildirim:
I'd say 2. is the most obvious and least verbose 😄

Malian:
> Alright. Time to vote!
> 
> `Strict` and `non-strict` versions:
> 
>     1. `values` and `suggestions`
> 
>     2. `values` and `values!`
> 
>     3. `values: ["v1", "v2"]` and `values: ["v1", "v2", ...]`
> 
> 
> /cc @surface-ui/core

3

lnr0626:
> Alright. Time to vote!
> 
> `Strict` and `non-strict` versions:
> 
> 1. `values` and `suggestions`
> 2. `values` and `values!`
> 3. `values: ["v1", "v2"]` and `values: ["v1", "v2", ...]`
> 
> /cc @surface-ui/core 

2, but only a slight preference over 3

msaraiva:
Although I slightly prefer `#3` over `#2`, I believe the former will be more intuitive to the user. Besides, it has the benefit of not introducing a breaking change as it keeps the current behaviour of `values`. So I guess `#2` is the winner!

@yusufyildirim feel free to move on and let's know if you need any assistance. 

Thanks for working on this!

msaraiva:
Hi @yusufyildirim!

Just to make it clear, we're only checking if the default value is in the list if the user sets `values!`, not `values`. 

We'll also need to find a way to indicate this difference `values` vs `values!` in the docs. For instance, the `Card` component from SurfaceBootstrap has a `width` prop. Although it suggests one of `25, 50, 75, 100`, it can actually be any percentage value so it should keep it as `values`, not `values!`.

image

The first thing that comes to my mind is to add `, ...` to the end of the list, like:

```
25, 50, 75, 100, ...
``` 

Any other idea?

Another thing we could validate in a separate PR is if all values in the list are valid according to the specified type. This validation can be applied to both, `values` and `values!`.

yusufyildirim:
Hi @msaraiva 

Yes, everything is clear. I didn't have time to work on this for now BTW.

> The first thing that comes to my mind is to add, ... to the end of the list, like:

Totally makes sense. I'm a bit worried about number values though. Adding `...` to the end of a sequence of numbers makes me feel like there is a `pattern` going on. For example, seeing `...` at the end of a sequence like  `5, 10, 15, 20, ...` might make me think if like the next possible number would be 25, because there is a pattern 😄 This is a bit related to how `...` works in math I believe 😄 

I don't know if I explained it well but this might be worth thinking a bit. Except that, it's ok.

msaraiva:
> Adding `...` to the end of a sequence of numbers makes me feel like there is a pattern going on.

Yeah. It could give that impression. Another option could be:

`25`, `50`, `75`, `100` or a valid `:integer`

But that's quite verbose. Not sure.

yusufyildirim:
I don't know, maybe we should just add a sentence or a paragraph that explains what `...` means, and assuming people read the explanation, we use `...` everywhere in the docs. People not going to read it anyway... 😄

davydog187:
Hello @yusufyildirim, do you need help with this PR? I've similarly become interested in this problem, as well as adding some possibilities with ranges in values so you can do things like

```elixir
prop age, :integer, values!: 1..100
```

davydog187:
Regardless of my suggestion, which probably is a separate topic, I'm happy to take over this PR to push it over the line.

yusufyildirim:
Hello @davydog187, I'm pretty busy these days -even weekends- and this is why I couldn't find time to implement this. If you want to continue, please feel free to take over

davydog187:
Great, and I'll add support for my current PR #355 

Thank you @yusufyildirim

davydog187:
I just opened https://github.com/surface-ui/surface/pull/374

yusufyildirim:
Closing this one, thank you @davydog187!

PR #317, opened 2021-03-31
Add a warning if props are specified many times, but accumulate is false

### LINKED ISSUE #302: Warn on duplicated props without accumulate: true
Hey there,
I think it might be helpful to raise a compile error in case of passing duplicated props or slots to a component.
Additionally the current behavior varies between props and slots:
```
defmodule MyApp.Components.Modal do
  use Surface.LiveComponent

  slot header
  prop paragraph, :string

  def render(assigns) do
    ~H"""
    <div>
      <slot name="header" />
      <p>@paragraph</p>
    </div>
    """
  end
end
```

```
<Modal paragraph="one" paragraph="two">
  <template slot="header">
    <p>one</p>
  </template>
  <template slot="header">
    <p>two</two>
  </template>
</Modal>
```

This example would render the `Modal` with `paragraph=two` but with the `header-slot = one`.

I will try to submit a PR on this topic if people agree with this proposal (might take me some time though).

Malian:
FIxes #302

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -533,13 +533,15 @@ defmodule Surface.Compiler do
     if not accumulate? and Keyword.has_key?(acc, name) do
       IOHelper.warn(
         """
-        The prop `#{name}` has been set many times, but accumulate is false.
+        The prop `#{name}` has been passed multiple times. Considering only the last value.
 
-        Hint: You can remove the multiple use of the #{name} attribute, or set accumulate to `true`:
+        Hint: Either remove all redundant definitions or set option `accumulate` to `true`:
 
         ```
-        prop #{name}, :#{type}, accumulate: true
+          prop #{name}, :#{type}, accumulate: true
         ```
+
+        This way the values will be accumulated in a list.
         """,
         meta.caller,
         fn _ -> attr_meta.line end
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -515,10 +515,36 @@ defmodule Surface.Compiler do
 
   defp process_attributes(_module, [], _meta), do: []
 
-  defp process_attributes(mod, [{name, value, attr_meta} | attrs], meta) do
+  defp process_attributes(mod, attrs, meta), do: process_attributes(mod, attrs, meta, [])
+
+  defp process_attributes(_module, [], _meta, acc) do
+    acc
+    |> Keyword.values()
+    |> Enum.reverse()
+  end
+
+  defp process_attributes(mod, [{name, value, attr_meta} | attrs], meta, acc) do
     name = String.to_atom(name)
     attr_meta = Helpers.to_meta(attr_meta, meta)
     {type, type_opts} = Surface.TypeHandler.attribute_type_and_opts(mod, name, attr_meta)
+
+    accumulate? = Keyword.get(type_opts, :accumulate, false)
+
+    if not accumulate? and Keyword.has_key?(acc, name) do
+      IOHelper.warn(
+        """
+        The prop `#{name}` has been set many times, but accumulate is false.
+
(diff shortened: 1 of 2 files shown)


lib/surface/compiler.ex
```diff
@@ -515,11 +515,37 @@ defmodule Surface.Compiler do
...
+  end
+
+  defp process_attributes(mod, [{name, value, attr_meta} | attrs], meta, acc) do
     name = String.to_atom(name)
     attr_meta = Helpers.to_meta(attr_meta, meta)
     {type, type_opts} = Surface.TypeHandler.attribute_type_and_opts(mod, name, attr_meta)
 
+    accumulate? = Keyword.get(type_opts, :accumulate, false)
+
+    if not accumulate? and Keyword.has_key?(acc, name) do
```
msaraiva, line 533: I'm a bit concern about `Keyword.has_key?(acc, name)`. Since most of the attributes are not accumulative, this will run for almost every attribute, traversing the remaining attributes multiple times (unlike maps, keywords must be fully traversed to check for a key). We have this kind of logic in many other places but we need to start avoiding this so we don't make compilation slower (it's already a bit slower than it should, IMO). 

My suggestion is to instead of doing this validation on every `process_attributes` call, we move this logic to a single place (maybe `validate_properties`?). There we can do a `reduce` or use a `group_by` to process the validation. This way we guarantee we traverse the list only twice. One for grouping/counting, another one to emit the warnings. 

In the future, we should consider having a single optimized data structure built by a `reduce`, gathering all kinds of information in sets and maps, so we can pass it and use it efficiently in multiple processings/validations.

If you prefer, feel free to keep the current implementation. We can search for all similar cases and create another PR with the optimizations.

test/properties_test.exs
```diff
@@ -624,4 +624,39 @@ defmodule Surface.PropertiesSyncTest do
...
+    id = :erlang.unique_integer([:positive]) |> to_string()
+    module = "TestComponentWithPropSpecifiedMultipleTimes_#{id}"
+
+    code = """
+    defmodule #{module} do
+      use Surface.Component
+
+      def render(assigns) do
+        ~H"\""
+        <StringProp label="first label" label="second label" />
```
msaraiva, line 638: At which attribute should the warning be emitted? At the first, the second or both occurrences? I have no preference but we could write this code like:

```
<StringProp 
  label="first label" 
  label="second label" 
/>
```

just to make sure it warns on the expected line.

msaraiva (commented):
@Malian great stuff!

A left a couple of minor comments.

test/properties_test.exs
Malian: Good catch! The PR is doing it wrong, it warns on the first line no matter on which line is the prop. I'll update it. It will warn on the second occurrence.

lib/surface/compiler.ex
Malian: > unlike maps, keywords must be fully traversed to check for a key

I didn't know that, I really though it was like a map. Neither the Elixir nor the Erlang documentation mention that.

I have done benchmarks and there is no increase in time no matter how big the list is if you search for an item in the same place.

> In the future, we should consider having a single optimized data structure built by a single reduce, gathering all kinds of information in sets and maps, so we can pass it and use it efficiently in multiple processings/validations.

I agree.

> If you prefer, feel free to keep the current implementation. We can search for all similar cases and create another PR with the optimizations.

My first thought is to do the right thing right away but I wonder if it wouldn't be better to put this in a pr with the other optimsiations.

msaraiva (approved):

lib/surface/compiler.ex
Malian: > Your comment about the complexity is totally right! A Map as a complexity of O(1) and a list of O(n) to check if a key exists. I was challenging the part that says we have to traverse all the list to check for a key. In the worst case this is right.

PR #325, opened 2021-04-08
update CHANGELOG

lnr0626:
not sure the best wording for this, open to suggestions :-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.4.0 (WIP)
+
+  * Call render when defined in slotable components (#283)
+
 ## v0.3.2 (2021-03-19)
 
   * Warn if prop is required and has default value (#282)


msaraiva (approved):
@lnr0626 I think this is fine for now. We can change it later before the release since people may not understand what this is about as we didn't have explicitly in the docs the concept of "slotable components", have we?

Do you know if we should change the naming we use for those kinds of components? Should we create a `Surface.SlotableComponent` and then say in the docs that a renderless component is just a slotable component without implementing `render/1`?. Or maybe we should keep it as it is and just explain the concept?

We'll probably need some basic examples in the docs too to make it clear when to use each and what are the tradeoffs. @Malian since you're already using this feature, you may have some good insights. WDYT?

lnr0626:
I think splitting them up would be a good idea to help distinguish the concept, and might make it a bit easier to explain in the docs

miguel-s:
I also like the idea of splitting them! It would help us to introduce the concept of slots more gradually to new users.

I use slots quite extensively for a while now (love them!), but even so I have had trouble keeping track of slottables and renderless slottables etc. 😅

miguel-s (approved):

Issue #320, opened 2021-03-31
Add class prop on Form component


gushonorato:
Surface let me pass classes into form inputs using a class property like this:

```
<TextInput class="lero" .../>
```

but when working with `Form` I must pass class using opts:

```
<Form opts={{ class: "lero" }} .../>
```

I think Surface should permit pass classes in forms using props, like we do in inputs. It would become the design more orthogonal.

Malian:
Hi @gushonorato !

You are right, we should be consistent with other Form components.

Would you like to give a try to fix this? You can take a look at the LiveFileInput component for inspiration https://github.com/surface-ui/surface/blob/master/lib/surface/components/live_file_input.e

msaraiva:
Fixed in #327.

PR #327, opened 2021-04-09
Add class prop to Form component

miguel-s:
Related to #320.

### DIFF
--- a/test/components/form_test.exs
+++ b/test/components/form_test.exs
@@ -180,4 +180,16 @@ defmodule Surface.Components.FormTest do
 
     assert html =~ ~r/class="form form-user"/
   end
+
+  test "setting multiple classes as css_class" do
+    html =
+      render_surface do
+        ~H"""
+        <Form for={{:user}} action="#" class={{ "form", "form-user": true }}>
+        </Form>
+        """
+      end
+
+    assert html =~ ~r/class="form form-user"/
+  end
 end
--- a/test/components/form_test.exs
+++ b/test/components/form_test.exs
@@ -156,4 +156,28 @@ defmodule Surface.Components.FormTest do
 
     assert html =~ ~s(><input name="_method" type="hidden" value="put"/>)
   end
+
+  test "setting the class" do
+    html =
+      render_surface do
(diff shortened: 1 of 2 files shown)


test/components/form_test.exs
```diff
@@ -156,4 +156,28 @@ defmodule Surface.Components.FormTest do
...
+      end
+
+    assert html =~ ~r/class="form"/
+  end
+
+  test "setting multiple classes" do
+    html =
+      render_surface do
+        ~H"""
+        <Form for={{:user}} action="#" class="form form-user">
```
msaraiva, line 176:
```suggestion
        <Form for={{:user}} action="#" class={{ "form", "form-user": true }}>
```

Just to make sure it accepts `:css_class` ;)

msaraiva (approved):
❤️

test/components/form_test.exs
miguel-s: I added it as a separate test.

In other form components we always test for class as single string `class="input"` and multiple strings `class="input primary"` (this form used to have a bug), but we don't test for `class={{ "input", "primary": true }}`, should I add it to all components?

msaraiva: > should I add it to all components?

No need to do that now, however, we should consider a way to avoid repeating all those tests for each input. Currently,  42% of all tests are related to form/inputs components. Most of them are repeated. We can try to address this after `v0.4` is out.

Malian (changes requested):
@miguel-s I suggest to add one last test about the `default_class` for the Form component.

```elixir
  test ":default_class config" do
    using_config Form, default_class: "default_class" do
      html =
        render_surface do
          ~H"""
          <Form for={{ :user }} action="#" />
          """
        end

    end
  end
```

miguel-s:
@Malian Yes good catch, added the test!

Malian (approved):

test/components/form_test.exs
miguel-s: You are absolutely right! I've been thinking about this, but don't have a plan yet...

PR #335, opened 2021-04-15
Add Context get/3, fetch/2, and fetch!/2

### LINKED ISSUE #260: Create function Context.get(socket, ...) to access the context
Is there a public API for accessing the context inside the `update` callback of a component?

dvic:
Closes #260

### DIFF
--- a/lib/surface/components/context.ex
+++ b/lib/surface/components/context.ex
@@ -109,6 +109,18 @@ defmodule Surface.Components.Context do
     """
   end
 
+  def get(assigns, key, default \\ nil) do
+    Map.get(assigns.__context__, key, default)
+  end
+
+  def fetch(assigns, key) do
+    Map.fetch(assigns.__context__, key)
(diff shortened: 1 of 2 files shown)


dvic:
@msaraiva Not sure if this is something that you had in mind? I'm not familiar with the internals of Surface so I don't know if there is a better way to write a test for these accessors. The simple test in this PR should at least ensure that we update the accessors whenever the implementation of the context changes.

dvic:
Let me know if this is something and I'll add docs for the three new public functions.

dvic:
Wait, I just found out that `__context__` is not always passed? So how is this function supposed to work without having `Context get={{ ... }}` in the template? I guess we can't get that to work right?

dvic:
In light of the discussion in #260 I'll go ahead and close this PR for now because I have a feeling there's more to this feature than I first thought :)

PR #336, opened 2021-04-15
First try of a plugins system

Malian:
- Implemented only for LiveView.
- Only mount
- Dummy implemenation

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -6,7 +6,9 @@ locals_without_parens = [
   data: 3,
   data: 2,
   slot: 1,
-  slot: 2
+  slot: 2,
+  plugin: 1,
+  plugin: 2
 ]
 
 [
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -45,7 +45,8 @@ defmodule Surface.API do
     arities = %{
       prop: [2, 3],
       slot: [1, 2],
-      data: [2, 3]
+      data: [2, 3],
+      plugin: [1, 2]
     }
 
     functions = for func <- include, arity <- arities[func], into: [], do: {func, arity}
@@ -56,6 +57,7 @@ defmodule Surface.API do
       @after_compile unquote(__MODULE__)
 
       Module.register_attribute(__MODULE__, :assigns, accumulate: true)
+      Module.register_attribute(__MODULE__, :plugins, accumulate: true)
       # Any caller component can hold other components with slots
       Module.register_attribute(__MODULE__, :assigned_slots_by_parent, accumulate: false)
 
@@ -74,6 +76,7 @@ defmodule Surface.API do
       quoted_prop_funcs(env),
       quoted_slot_funcs(env),
       quoted_data_funcs(env),
+      quoted_plugins_funcs(env),
       quoted_context_funcs(env)
     ]
   end
@@ -100,6 +103,11 @@ defmodule Surface.API do
     build_assign_ast(:data, name_ast, type_ast, opts_ast, __CALLER__)
   end
 
+  @doc "Defines plugins for the component"
+  defmacro plugin(name_ast, opts_ast \\ []) do
+    build_plugin_ast(name_ast, opts_ast, __CALLER__)
+  end
+
   @doc false
   def get_assigns(module) do
     Module.get_attribute(module, :assigns, [])
@@ -121,10 +129,10 @@ defmodule Surface.API do
   end
 
   @doc false
-  def get_defaults(module) do
-    for %{name: name, opts: opts} <- get_data(module), Keyword.has_key?(opts, :default) do
-      {name, opts[:default]}
-    end
+  def get_plugins(module) do
+    module
+    |> Module.get_attribute(:plugins, [])
+    |> Enum.reverse()
   end
 
   @doc false
@@ -141,6 +149,12 @@ defmodule Surface.API do
 
     Module.put_attribute(caller.module, :assigns, assign)
     Module.put_attribute(caller.module, assign.func, assign)
+  end
+
+  @doc false
+  def put_plugin(caller, module, opts, _opts_ast, _line) do
+    plugin = {module, opts}
+    Module.put_attribute(caller.module, :plugins, plugin)
   end
 
   @doc false
@@ -216,6 +230,17 @@ defmodule Surface.API do
       @doc false
       def __data__() do
         unquote(Macro.escape(data))
+      end
+    end
+  end
+
+  defp quoted_plugins_funcs(env) do
+    plugins = get_plugins(env.module)
+
+    quote do
+      @doc false
+      def __plugins__() do
+        unquote(Macro.escape(plugins))
       end
     end
   end
@@ -408,7 +433,7 @@ defmodule Surface.API do
   end
 
   defp get_valid_opts(:data, _type, _opts) do
-    [:default, :values, :values!]
+    [:default, :values, :values!, :temporary]
   end
 
   defp get_valid_opts(:slot, _type, _opts) do
@@ -722,6 +747,45 @@ defmodule Surface.API do
             line: caller.line
           ] do
       Surface.API.put_assign(__ENV__, func, name, type, opts, opts_ast, line)
+    end
+  end
+
+  defp validate_plugin_name_ast!({:__aliases__, _, _} = name, caller) do
+    Macro.expand(name, caller)
+  end
+
+  defp validate_plugin_name_ast!(name_ast, caller) do
+    message = """
+    invalid plugin name. Expected a module, got: #{Macro.to_string(name_ast)}\
+    """
+
+    IOHelper.compile_error(message, caller.file, caller.line)
+  end
+
+  defp validate_plugin_opts_ast!(_name, opts, _caller) when is_list(opts) do
+    opts
+  end
+
+  defp validate_plugin_opts_ast!(name, opts, caller) do
+    message = """
+    invalid options for the plugin #{name}. \
+    Expected a keyword list of options, got: #{Macro.to_string(opts)}
+    """
+
+    IOHelper.compile_error(message, caller.file, caller.line)
+  end
+
+  defp build_plugin_ast(name_ast, opts_ast, caller) do
+    name = validate_plugin_name_ast!(name_ast, caller)
+    opts = validate_plugin_opts_ast!(name, opts_ast, caller)
+
+    quote bind_quoted: [
+            name: name,
+            opts: opts,
+            opts_ast: Macro.escape(opts_ast),
+            line: caller.line
+          ] do
+      Surface.API.put_plugin(__ENV__, name, opts, opts_ast, line)
     end
   end
 end
--- a/lib/surface/live_component.ex
+++ b/lib/surface/live_component.ex
@@ -50,8 +50,6 @@ defmodule Surface.LiveComponent do
       end
   """
 
-  alias Surface.BaseComponent
-
   defmacro __using__(_) do
     quote do
       @before_compile Surface.Renderer
@@ -61,11 +59,16 @@ defmodule Surface.LiveComponent do
 
       @before_compile unquote(__MODULE__)
 
-      use Surface.API, include: [:prop, :slot, :data]
+      use Surface.API, include: [:prop, :slot, :data, :plugin]
       import Phoenix.HTML
 
       alias Surface.Constructs.Deprecated.{For, If}
       alias Surface.Components.{Context, Raw}
+
+      plugin Surface.Plugins.InitializeSurfacePlugin
+      plugin Surface.Plugins.DefaultAssignsPlugin
+      plugin Surface.Plugins.TemporaryAssignsPlugin
+      plugin Surface.Plugins.RestorePrivateAssignsPlugin
 
       @doc """
       The id of the live component (required by LiveView for stateful components).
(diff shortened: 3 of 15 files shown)


msaraiva (commented):
@Malian fantastic job! ❤️

A couple of things for us to consider:

### 1. No global plugins

Aside from the built-in plugins, I think we should avoid loading plugins globally from configs. I'd rather force the user to explicitly set them either using a new `plugin` macro like:

```elixir
use Surface.Liveview
...

plugin MyPlugin
plugin OtherPlugin, whatever: true 
```

or passing the list as an option:

```elixir
use Surface.Liveview, plugins: [
  {MyPlugin, []},
  {OtherPlugin, whatever: true }
]
...
```

This way, the user can apply different plugins to different groups of views/components using `lib/my_app_web.ex`. For instance:

```elixir
# lib/my_app_web.ex

...

def surface_view do
  quote do
    ...
    plugin MyPlugin
    plugin OtherPlugin, whatever: true 
  end
end
```

### 2. Options conflict

I'm having second thoughts about allowing plugins to define their own options to `prop`, `data`, etc.

I believe you have already exposed the same concern. The main issue here is how to avoid conflicts between plugins if each one of them can define their own options. 

We could allow the user to define options for a specific plugin. Like:

```elixir
data items, :list, options_for: [{MyPlugin, some_option: ...}]

plugin MyPlugin
```

But that can get really verbose.

Another option is to invert the logic, not allowing extra options for `data` and use the plugins' own set of options to pass any required information about the data. Example:

```elixir
data items, :list

plugin MyPlugin, options_for: [{:items, some_option: ...}]
```

There are other things that comes to my mind but I think we need to find a solution for those two issues before moving on. WDYT?

Malian:
> No global plugins

Good point. You convinced me to be more explicit with the plugin macro or passing the plugins as an option to  `use Surface.LiveView, pluings: []`

> I'm having second thoughts about allowing plugins to define their own options to prop, data, etc.

I agree. I allowed plugins to modify the list of options because for, let's say, core plugins, I would not force the user to be verbose for `default` and `temporary` options. By hardcoding these two options, we are coupling it to the plugins system. That being said, it could be a tradeoff. 

Regarding the two propositions of how passing options to plugins, I am wondering if we could avoid the `plugin` macro and infer the list of plugin based on the options that are passed to `data`

Instead of
```elixir
data items, :list, options_for: [{MyPlugin, some_option: ...}]

plugin MyPlugin
```
We could have
```elixir
data items, :list, plugins: [{MyPlugin, some_option: [...]}]
```

Why did you define options by data? I suppose we could have different options for different data but I do not find a concrete example in mind... Why should a plugin act differently depending on the data?

```elixir
data items, :list, options_for: [{MyPlugin, some_option: ...}]
data contacts, :list, options_for: [{MyPlugin, some_option: ...}]
```

If it is not needed, I believe we could have a less verbose syntax:
```elixir
plugin MyPlugin, opts_one
plugin MyPlugin, opts_two

data items, :list, plugins: [MyPluginOne, MyPluginTwo]
```
If I have to choose, I prefer the syntax where I do not have to scroll up and down to have the full info about a `data`.

```elixir
data items, :list, options_for: [{MyPlugin, some_option: ...}]
```

msaraiva:
> I agree. I allowed plugins to modify the list of options because for, let's say, core plugins, I would not force the user to be verbose for `default` and `temporary` options. By hardcoding these two options, we are coupling it to the plugins system. That being said, it could be a tradeoff.

I'm fine to have `temporary` as built-n since it also a built-in feature in LV itself.

This whole plugin thing overlaps with the idea of data providers too. As far as I can see, all examples we have could be implemented as data providers too. This topic keeps coming back as @olivermt mentioned in https://github.com/surface-ui/surface/issues/260#issuecomment-821995260 so maybe we should discuss the whole thing if we want to have a consistent API. A data provider behaviour would have a very similar set of callbacks.

msaraiva:
@Malian another solution would be to use a module attribute to add metadata to `data` instead of manipulating the options. Something like:

```elixir
@plugin MyPlugin
data items, :list
```

with options:

```elixir
@plugin {MyPlugin, opt1: "whatever", opt2: "other stuff"}
data items, :list
```

Or if we go with the data provider approach:

```elixir
@provider {PubSub, key: "items_from_pubsub"}
data items, :list
```

That information would not be available in the same set of the built-in options but in a separate data structure, e.g. `metadata` or `plugin_metadata` or `provider_metadata`.

olivermt:
I have read the test that was added, and I am still not quite clear on what the usecase for a Plugin is.

Do you have a better non-abstract example of what MyPlugin could do?

msaraiva:
> Do you have a better non-abstract example of what MyPlugin could do?

@olivermt plugins are basically just a way to inject common code into components/LV's callbacks. Currently, we hardcode that common code at compile-time. With plugins, we just have to hardcode the basic plugin logic that iterates over the plugins list. This way any other logic that adds a specific feature can be isolated in a separate module. Here are the 3 built-in plugins we could identify: 

https://github.com/surface-ui/surface/blob/589f968ec3ad91434dc28d63d6b7581e9f834681/lib/surface.ex#L162-L164

Other plugins could then be configured by the user, for instance, an `AuthPlugin` that automatically do all the logic related to authentication.

@Malian maybe we need both abstractions. Plugins for things that are not directly related to any specific `data` and Data Providers for those features that add behaviour to a single `data`. 

Data provider config should be added individually for each `data`. Like:

```elixir
@provider {PubSub, key: "items_from_pubsub"}
data items, :list
```

and Plugin config should be added alongside the plugin definition:

```elixir
plugin Auth, user_data: :current_user, ...
```

Plugins should probably also have a callback to inject code in the component itself. This way they could add their own `data` definitions or anything else they need.

Malian:
Here are some thoughts based on the project I am working on. No need for answers ;)

I am working on a client project wherein most of the live views I have to:

1. Check that the user is logged in
    1. If the user is logged in, assign some defaults: current_user and current_company where current_company depends if current_user has been set or not.
    2. If the user is not logged id, redirect him to the log in page
2. Assign default values that come from database and/or modules
    1. Simple as `assign(:values, Context.list_values())`
    2. I also have to perform simple transformations on these data (sorting, prepare data for views, etc.)
    3. I also have to assign default assigns based on value from other assigns. The interesting thing here is that these transformations should be applied several times depending on some parameters that could be changed by the end-user.

Here is an example of a simple live view that implements all these cases. 
```elixir
defmodule MyLiveView do
  use Surface.LiveView

  data current_user, :struct
  data current_company, :struct
  data my_list, :list
  data my_sorted_list, :list

  data var1, :string, default: "var1"
  data var2, :string, default: "var2"
  data my_dynamic_list, :list

  def mount(_params, session, socket) do
    socket = 
      socket
      |> assign_defaults() # Will check if user is logged in + assign current_user + redirect if not logged in
      |> assign_my_list()
      |> assign_my_sorted_list()
      |> assign_my_dynamic_list()

    {:ok, socket}
  end

  defp assign_my_list(socket) do
    {:noreply, assign(socket, :my_list, MyList.list_lists()}
  end

  defp assign_my_sorted_list(socket) do
    {:noreply, assign(socket, :my_sorted_list, Enum.sort(MyList.list_lists(order: :name), &sort_my_list)}
  end

  defp assign_my_dynamic_list(socket) do
    {:noreply, assign(socket, :my_list, MyList.list_lists(socket.assigns.var1, socket.assigns.var2))}
  end

  def handle_event("update_my_dynamic_list", %{"var1": var1, "var2": var2, socket) do
    socket = 
      socket
      |> assign(:var1, var1)
      |> assign(:var2, var2)
      |> assign_my_dynamic_list()    

    {:noreply, socket}
  end

end
```

All of these examples can indeed be implemented as data providers but several concerns pop in my mind:

- I do not want to write `data current_user, :struct` and `data current_company` for all my live views. That being said, I am afraid losing some clarity if we let the user define `data` without explicitly declare it. Moreover, it could be implemented as a data provider but not sure it is the right way to do as, some side effects exist in that kind of data provider like redirection.
- `current_company` is a value computed based on `current_user` so it could be a data provider that references a key from another `data` that comes from another data providers... It means that data should be computed in the same order, every time. 
- `my_sorted_list` could be implemented as a local provider simple as `&assign_my_sorted_list`
- `my_dynamic_list` is a computed data based on other data
- How to differentiate a simple provider that does not require any reference to socket, etc. than providers that require reference to assigns, socket, session, etc. ?

Let's try to rewrite the above example with the above propositions:
```elixir
defmodule MyLiveView do
  use Surface.LiveView

  @provider MyAuth
  data current_user, :struct

  @provider ??? No idea what could be a simple syntax to reference a key from another data/prop
  data current_company, :struct

  @provider &MyList.list_lists/0
  data my_list, :list

  @provider &assign_my_sorted_lists/1
  data my_list, :list

  data var1, :string, default: "var1"
  data var2, :string, default: "var2"

  @provider &assign_my_dynamic_list/1, data: [:var1, :var2]
  data my_dynamic_list, :list

  defp assign_my_sorted_list(socket) do
    {:noreply, assign(socket, :my_sorted_list, Enum.sort(MyList.list_lists(order: :name), &sort_my_list)}
  end

  defp assign_my_dynamic_list(socket) do
    {:noreply, assign(socket, :my_list, MyList.list_lists(socket.assigns.var1, socket.assigns.var2))}
  end

  def handle_event("update_my_dynamic_list", %{"var1": var1, "var2": var2, socket) do
    socket = 
      socket
      |> assign(:var1, var1)
      |> assign(:var2, var2)
      |> assign_my_dynamic_list()    

    {:noreply, socket}
  end

end
```

I am indeed wondering if we should have two abstractions. One for provider and one for plugins as suggested but I am a little bit lost of what could be other examples of plugins instead of the default ones:
- InitializeSurface
- Default assigns <-- Could be a provider too
- Temporary assigns

msaraiva:
> but I am a little bit lost of what could be other examples of plugins instead of the default ones

@Malian I think any functionality that requires interaction with the socket through `life-cycle callbacks` or `event handlers` may be a good candidate. Let's take this example from the [Uploads](https://hexdocs.pm/phoenix_live_view/uploads.html#appendix-a-uploadlive) guide.

A plugin could be responsible for most (or even all) that common logic. We could define different types of uploads, e.g. `local` or `S3` and then ask for different configuration options that could be used by the plugin to fill the gaps, in this case, the non-generic pieces for each type of upload. For instance, the local one, could the set as something like:

```elixir
plugin Upload,  
  name: :avatar,  
  uploaded_entries: :uploaded_files, 
  accept: ~w(.jpg .jpeg), 
  dest: "priv/static/uploads"
```

We don't even have to address all possible scenarios. The user can still implement it manually if necessary. However, if this could work for most common scenarios by just adding a plugin entry + injecting a couple of components, that would be awesome ;)

msaraiva:
@Malian do you think we could have a simpler version of this feature but instead of loading the plugins with `Application.get_env(:surface, :plugins, [])` we could use an explicit API? For instance, either:

```elixir
use Surface.Component

plugin ThePlugin,  opt1: ...,  opt2: ...
```

or

```elixir
use Surface.Component, plugins: [
  {ThePlugin,  opt1: ...,  opt2: ...},
  ...
]
```

It seems to me that even if we keep it as just a private API for now, allowing us to move the logic that is injected in the different components into their related plugins, it will already be an improvement from the design PoV.
 
Also, besides the builtin-in plugins you already identified, I'm planning to add another one to handle Surface Catalogue's examples/playground's common behaviour.

I'd also like to make some experiments with the previously mentioned `Upload` plugin. I've seen many people having trouble setting up the uploads properly. It might be worth it to give it a try. WDYT?

Malian:
@msaraiva I'll try to do this next weekend.

Malian:
@msaraiva Quick update. I have a version that handles both LiveComponent and LiveView with the following callbacks:

- before_mount_live_view
- after_mount_live_view
- before_mount_live_component
- after_mount_live_component
- before_update_live_component
- after_update_live_component

I didn't add the callback for `Component` as I am waiting to know what we will do with #400 

I have to add some tests as I moved some part of the code into plugin ,but the tests still pass even without the plugins activated, strange... I'll also clean it before updating the branch.

I'll do this in the next few days.

msaraiva:
@Malian that's great news! I'm excited about the possibilities this might bring.

> I have to add some tests as I moved some part of the code into plugin ,but the tests still pass even without the plugins activated, strange...

Very strange 🤔

Malian:
@msaraiva & @miguel-s I updated the code to the last working version I have on my computrer. Actually, you can pass options to the plugin but these options are not used yet, and are not part of the API. It is quite easy to fix.

If you disabled the `RestorePrivateAssignsPlugin` from `LiveComponent` all the test remain green :thinking: 

Feel free to start from scratch or from this branch if you want to go further with the plugin system.

Here are some thought:
- This does not work for stateless `Component`
- Order matters! Actually, there is no way to execute a plugin before the defaults ones.
- The API should be reviewed:
  - Should we pass the `module` as the first argument or through options? etc.
  - How should we pass `options`?
  - Should we define a `%Plugin{}` struct that handles the `env`? Something like: `%Plugin{module: MyLiveView, opts: [...]}`. I believe so.
- There are lack of duplicated code in this version and there is a lack of tests
- There is no callback for `handle_event`. It is probably something we should add with an `Upload` plugin @msaraiva

Feel free to comment and ask questions, I would be glad to help to go further with this feature.

msaraiva:
@Malian thanks for working on this. I'll certainly give it a try this week ;)

msaraiva:
I'm closing this since LV will support something similar in the next version. If it turns out it's not what we want, we reopen it :)

Issue #330, opened 2021-04-11
Add support for all LiveView events in built-in components


miguel-s:
We have implemented support for a few LiveView events in most of the built-in components (mostly form inputs, buttons and links), but we're still missing a few.

### Params
- [ ] phx-value-*

### Click Events
- [x] phx-capture-click
- [ ] phx-click

### Focus/Blur Events
- [ ] phx-window-focus
- [ ] phx-window-blur
- [x] phx-focus
- [x] phx-blur

### Key Events
- [ ] phx-key
- [ ] phx-window-keyup
- [ ] phx-window-keydown
- [x] phx-keyup
- [x] phx-keydown

### Form Events
- [ ] phx-auto-recover
- [ ] phx-trigger-action
- [ ] phx-disable-with
- [ ] phx-feedback-for
- [x] phx-submit
- [x] phx-change

msaraiva:
@miguel-s what are your plans for `phx-value-*`? Since they are not events and don't require any compile-time stuff (like infer the phx-target), they don't need to be implemented as a directive. What we could do is to create a prop `values` like:

```elixir
prop values, :keyword, default: []
```

and use it like:

```jsx
<button values={{ id: "213", name: "My name" }} ...>
```

then we could traverse it at runtime, generating `phx-value-id` and `phx-value-name` dynamically and pass it along with the other attributes. Is that somehow what you have in mind?

miguel-s:
@msaraiva Yes that's essentially all we need.

I'm considering two points:
1. Giving the prop a more explicit name (maybe `event_values` or `click_values`), so it doesn't get confused with the input `value` prop
2. Adding a warning if `values` is defined without one of the click events (compile-time or on mount?)

msaraiva:
> * Giving the prop a more explicit name (maybe `event_values` or `click_values`), so it doesn't get confused with the input `value` prop

I'd just go with `values`. Since one expects a string and the other a keyword list, if the user mess it up, a proper error should be raised.

> * Adding a warning if `values` is defined without one of the click events (compile-time or on mount?)

That might be tricky to do at compile-time as the user can still pass any of those events and values using the `opts` prop.

Malian:
I feel that the name `values` is confusing regarding `value`. Even with a proper message:
> The property "values" requires a keyword list, a string was given.
Hint: Use the property "value" instead of "values" or pass a keyword list [key1: "val, ...]...

the name "values" does not help the user to understand what values are we talking about? data-value? phx-value?

[...]

PR #337, opened 2021-04-15
Add missing LV events

miguel-s:
Related to #330.

### DIFF
--- a/lib/surface/components/events.ex
+++ b/lib/surface/components/events.ex
@@ -1,0 +1,31 @@
+defmodule Surface.Components.Events do
+  @moduledoc false
+
+  defmacro __using__(_) do
+    quote do
+      @doc "Triggered when the page loses focus"
+      prop window_blur, :event
+
+      @doc "Triggered when the page receives focus"
+      prop window_focus, :event
+
+      @doc "Triggered when the component loses focus"
+      prop blur, :event
+
+      @doc "Triggered when the component receives focus"
+      prop focus, :event
+
+      @doc "Triggered when the component captures click"
+      prop capture_click, :event
+
+      @doc "Triggered when the component receives click"
+      prop click, :event
+
+      @doc "Triggered when a button on the keyboard is pressed"
+      prop keydown, :event
+
+      @doc "Triggered when a button on the keyboard is released"
+      prop keyup, :event
+    end
+  end
+end
--- a/lib/surface/components/button.ex
+++ b/lib/surface/components/button.ex
@@ -28,6 +28,7 @@ defmodule Surface.Components.Button do
   """
 
   use Surface.Component
+  use Surface.Components.Events
 
   import Surface.Components.Utils
 
@@ -47,21 +48,6 @@ defmodule Surface.Components.Button do
   The label for the generated `<button>` element, if no content (default slot) is provided.
   """
   prop label, :string
-
-  @doc "Triggered when the component loses focus"
-  prop blur, :event
-
-  @doc "Triggered when the component receives focus"
-  prop focus, :event
-
-  @doc "Triggered when the component receives click"
-  prop capture_click, :event
-
-  @doc "Triggered when a button on the keyboard is pressed"
-  prop keydown, :event
-
-  @doc "Triggered when a button on the keyboard is released"
-  prop keyup, :event
 
   @doc """
   Additional attributes to add onto the generated element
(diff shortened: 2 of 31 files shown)


lib/surface/components/events.ex
```diff
@@ -0,0 +1,37 @@
+defmodule Surface.Components.Events do
+  @moduledoc false
```
Malian, line 2: I believe we should make this module public, so developer/user can use it in their own components. WDYT?

We could even group the event by type, like in the [doc](https://hexdocs.pm/phoenix_live_view/bindings.html), and let the user choose which type of events, he would like to import:

```elixir
use Surface.Components.Events, only: [:click, :focus_blur, :form, :key] 
```

lib/surface/components/form/label.ex
```diff
@@ -42,10 +44,11 @@ defmodule Surface.Components.Form.Label do
   def render(assigns) do
     helper_opts = props_to_opts(assigns)
     attr_opts = props_to_attr_opts(assigns, class: get_config(:default_class))
+    event_opts = events_to_opts(assigns) |> opts_to_attrs()
```
Malian, line 47: It's just a matter of taste but I suggest writing it with two pipe operators.

```suggestion
    event_opts = 
       assigns 
       |> events_to_opts()
       |> opts_to_attrs()
```
or
 ```suggestion
     event_opts = assigns |> events_to_opts() |> opts_to_attrs()
```

**Review by Malian — APPROVED** [review]
@miguel-s Amazing job!

I left few minor comments.

It looks good to me. Not sure if you want to implement @msaraiva [suggestion](https://github.com/surface-ui/surface/issues/330#issuecomment-821816696) before emerging this. We could do it in a separate PR. As you whish.

File: lib/surface/components/events.ex (hunk 1 above)
Comment (miguel-s) [reply lib/surface/components/events.ex:2]: The thought had crossed my mind, but not sure if the benefits outweigh the cost of increasing the complexity of the Surface API.

**Comment by miguel-s** [comment]
> Not sure if you want to implement @msaraiva suggestion before emerging this

I thought I had covered this with `components/events.ex` and `test/components/events_test.exs` and merging the tests in all components. What did I miss?

**Comment by Malian** [comment]
> 
> 
> > Not sure if you want to implement @msaraiva suggestion before emerging this
> 
> I thought I had covered this with `components/events.ex` and `test/components/events_test.exs` and merging the tests in all components. What did I miss?

@miguel-s from my understanding you could safty convert that tests:
```
        <DateInput form="user" field="color" value="mybirthday"
          capture_click="my_capture_click"
          click="my_click"
          window_focus="my_window_focus"
          window_blur="my_window_blur"
          focus="my_focus"
          blur="my_blur"
          window_keyup="my_window_keyup"
          window_keydown="my_window_keydown"
          keyup="my_keyup"
          keydown="my_keydown"
        />

    assert html =~ ~s(phx-capture-click="my_capture_click")
    assert html =~ ~s(phx-click="my_click")
    assert html =~ ~s(phx-window-focus="my_window_focus")
    assert html =~ ~s(phx-window-blur="my_window_blur")
    assert html =~ ~s(phx-focus="my_focus")
    assert html =~ ~s(phx-blur="my_blur")
    assert html =~ ~s(phx-window-keyup="my_window_keyup")
    assert html =~ ~s(phx-window-keydown="my_window_keydown")
    assert html =~ ~s(phx-keyup="my_keyup")
    assert html =~ ~s(phx-keydown="my_keydown")
````
to
```
<DateInput form="user" field="color" value="mybirthday" capture_click="my_capture_click" />
assert html =~ ~s(phx-capture-click="my_capture_click")
````
as all the components use `events_to_opts` and you tested it in `events_test.ex`, so we know that the component can handle all the events defined in `events_to_opts`. If tommorow a new event is added to Phoenix Live View, we will have to add only 1 tests, instead of n (n = number of components).

Another way could be to abstract the list of events tests into a function `assert_surface_events html`. It's more explicit.

BTW, bmybe I am the guy who missed something 😅

miguel-s:
> If tomorrow a new event is added to Phoenix Live View, we will have to add only 1 tests, instead of n (n = number of components).

No, you're right. I was getting a bit concerned about removing all those tests, so I opted for a safe middle-ground, but had not considered the impact on maintenance. Good catch again and I'll change it.

msaraiva (approved):
@miguel-s great work! I love PRs that remove stuff :)

Malian (approved):

Issue #342, opened 2021-04-20
Form.Submit cannot contain custom components


PatrickSachs:
Hello there,

I am currently trying out surface. Rather enjoying the experience so far, although it takes a bit to shift my mental model from the SPA world I am used to :)

I am running into an issue that the Form.Submit button cannot contain custom components in its label:
```elixir
<Surface.Components.Form.Submit class="bg-sahnee-pink-500 rounded hover:bg-sahnee-pink-400 p-2 text-white">
  <Components.Icon icon="mdi-content-save" />
  Save
</Surface.Components.Form.Submit>
```
I'd assume that this is due to it being passed down to Phoenix which then in turn actually renders it. (Which uses the content_tag function that apparently can not contain custom components)

```
** (exit) an exception was raised:
    ** (ArgumentError) cannot convert component Sahnee.Business.Web.Components.Icon with id nil to HTML.

A component must always be returned directly as part of a LiveView template.

For example, this is not allowed:

    <%= content_tag :div do %>
      <%= live_component @socket, SomeComponent %>
    <% end %>
```

Is there any workaround or can this be fixed on the Surface side? I was considering ditching the form wrappers altogether anyways later on, but am wary of doing so since I suspect that would also make me loose out of the built in security features like CSRF.

Thank you for your time and the great library!

msaraiva:
Hi @PatrickSachs!

Assuming the underlying `submit/2` function provided by phoenix just creates a `<button type="submit">...</button>`, I believe you could use Surface's own `<Button>` component, which is `content_tag` free. Something like this should work:

```jsx
<Button opts={{ type: "submit" }} ...>
  ...
</Button>
```

@miguel-s do you think we should reimplement `<Submit>` wrapping our `<Button>` instead of `submit/2`? Or maybe phoenix's implementation does something extra? Also, would a `type` prop be welcome in `<Button>`?

miguel-s:
@msaraiva just double checked Phoenix's `submit/2` and seems to be pretty straighforward to reimplement our own version, let's go for it.

We might also want to rename the `<Button>` component to something closer to what it actually does.
The current name implies that this component builds a generic button, but actually it's based on Phoenix's `button/2`, which builds a very specialized button for submitting data to the server.

This is what the [docs](https://hexdocs.pm/phoenix_html/Phoenix.HTML.Link.html#button/2) say:
> Generates a button that uses a regular HTML form to submit to the given URL.
[...]

PR #344, opened 2021-04-21
Reimplement form Submit component

miguel-s:
Related to #342.

Reimplement the form `<Submit>` instead of relying on Phoenix's `button/2`.
The `button/2` function uses `content_tag/2` under the hood and prevents us from passing custom components to `<Submit>`.

### DIFF
--- a/lib/surface/components/form/submit.ex
+++ b/lib/surface/components/form/submit.ex
@@ -8,8 +8,7 @@ defmodule Surface.Components.Form.Submit do
   use Surface.Component
   use Surface.Components.Events
 
-  import Phoenix.HTML.Form, only: [submit: 2]
-  import Surface.Components.Utils, only: [events_to_opts: 1]
+  import Surface.Components.Utils, only: [events_to_opts: 1, opts_to_attrs: 1]
 
   @doc "The label to be used in the button"
   prop label, :string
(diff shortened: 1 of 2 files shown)


lnr0626 (commented):
lgtm :shipit:

msaraiva (approved):
Looks good to me!

Regarding:

> We might also want to rename the `<Button>` component to something closer to what it actually does.
> The current name implies that this component builds a generic button, but actually it's based on Phoenix's `button/2`, which builds a very specialized button for submitting data to the server.

Do you mean renaming `Surface.Components.Button` to `Surface.Components.Form.Button`? If so, yes. I agree.

miguel-s:
@msaraiva not quite and I think that proves my point that the naming is extremely confusing, no idea how Phoenix got away with it 😅

`<Button>` and `button/2` are not really buttons and are not related to forms. They're weird pseudo-links that submit data (usually post) to the server. Like this:

> button("hello", to: "/world")
> #=> `<button class="button" data-csrf="csrf_token" data-method="post" data-to="/world">hello</button>`

Maybe a better naming would be `Surface.Components.DataButton`? 

@lnr0626 @Malian do you have any experience using Phoenix's `button/2`?

lnr0626:
I do not, I hadn't realized it was used to submit data to the server

msaraiva:
> do you have any experience using Phoenix's `button/2`?

Not really. Maybe nobody is using it :)

mcrumm:
To be fair it is `Phoenix.HTML.Link.button/2` :)
TIL there is no _`Phoenix.HTML.Form.button/2`_, but perhaps there should be?

miguel-s:
@mcrumm you're absolutely right, thanks for pointing that out!

I will open a new issue for the renaming discussion, thanks everybody for your comments.

Malian:
@miguel-s  as the main purpose of reimplementing the form Submit component is to allow it to contains sub-component, I am wondering why there is no test to assert this behaviour? :sweat_smile: 

Except that, ok for me!

Malian (approved):

miguel-s:
@Malian we don't usually test for that in components, as it's basic Surface behaviour. Do you think we should add these tests to all components that accept children?

PR #345, opened 2021-04-21
Add documentation groups/nesting

davydog187:
Attempts to make the hex docs sidebar a bit more readable


msaraiva:
@davydog187 thanks!

PR #346, opened 2021-04-21
Parser adjustments

### LINKED ISSUE #258: Allowing the parser to return HTML comments
When SurfaceFormatter re-writes Surface code to format it, it's not aware of HTML comments because the Surface parser doesn't return them. Are there any options that would allow SurfaceFormatter to see HTML comments returned by the parser?

```
<!-- Comments like these -->
```

Thanks!

msaraiva:
Closes #258.

  * Allow new slot shorthand notation, e.g. `<:header>`
  * Don't ignore comments at the parser level
  * Make the compiler ignore comment nodes

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -208,8 +208,10 @@ defmodule Surface.Compiler do
   end
 
   defp to_ast(nodes, compile_meta) do
-    for node <- nodes do
-      case convert_node_to_ast(node_type(node), node, compile_meta) do
+    for node <- nodes,
+        result = convert_node_to_ast(node_type(node), node, compile_meta),
+        result != :ignore do
+      case result do
         {:ok, ast} ->
           process_directives(ast)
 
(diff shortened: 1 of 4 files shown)


msaraiva:
@paulstatezny with this change, I believe you'll be able to properly handle comments in the formatter. Please let me know if I missed anything.

lnr0626 (approved):
lgtm

lnr0626:
> Allow new slot shorthand notation

i see the changes to allow the slot shorthand in the parser (https://github.com/surface-ui/surface/pull/346/files#diff-adc759eb3677b0e309adbec1ecbf3d79646422e01743810147dd8561af668cd0R81), and don't see any changes to use that in the compiler

did you intend for this to be just the parser changes?

msaraiva:
> did you intend for this to be just the parser changes?

@lnr0626 yes. This is just to prepare the parser. The next step is to update the compiler. Maybe in the context of #268.

PR #347, opened 2021-04-22
Improve the error message when missing `id` on a LiveComponent

davydog187:
This should point new users in the direction that they may be using `Surface.LiveComponent` instead of `Surface.Component`

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -615,9 +615,7 @@ defmodule Surface.Compiler do
 
   defp validate_tag_children([%AST.Template{name: name} | _]) do
     {:error,
-     "templates are only allowed as children elements of components, but found template for #{
-       name
-     }"}
+     "templates are only allowed as children elements of components, but found template for #{name}"}
   end
 
   defp validate_tag_children([_ | nodes]), do: validate_tag_children(nodes)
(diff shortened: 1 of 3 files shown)


lib/surface/compiler.ex
```diff
@@ -615,9 +615,7 @@ defmodule Surface.Compiler do
 
   defp validate_tag_children([%AST.Template{name: name} | _]) do
     {:error,
-     "templates are only allowed as children elements of components, but found template for #{
-       name
-     }"}
+     "templates are only allowed as children elements of components, but found template for #{name}"}
```
davydog187, line 618: formatter seems to have changed a few lines. I can remove if we want

davydog187: ah i see why, i'm on elixir 1.12 and its formatting differently there. i'll send another PR to set `.tool-versions`

lib/surface/compiler.ex
```diff
@@ -719,6 +719,14 @@ defmodule Surface.Compiler do
 
       for prop_name <- missing_props_names do
         message = "Missing required property \"#{prop_name}\" for component <#{meta.node_alias}>"
+
+        message =
+          if prop_name == :id do
```
lnr0626, line 724: we do some stuff that allows using an `id` prop within a `Surface.Component` without it turning into a stateful component, and so we need to check if the module is stateful or stateless; which you can do with the `is_stateful_component/1` function:

```
if prop_name == :id and is_stateful_component(meta.module) do
  ...
else
  ...
end
```

Could you also update the error message to something like:

```elixir 
"""
Missing required property "#{prop_name}" for component <#{meta.node_alias}>

Hint: Components using Surface.LiveComponent automatically define a required `id` component to make them stateful. If you meant to create a stateless component, you can switch to `use Surface.Component`.
"""
```

lnr0626 (commented):
thanks for the pr @davydog187!

lib/surface/compiler.ex
davydog187: you got it!

davydog187: @leandrocp updated!

Malian (approved):

PR #348, opened 2021-04-22
Set .tool-versions for asdf

davydog187:
These match what your CI is using. We can bump to 1.11.4 as well

Noticed in #347

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,3 @@
+elixir 1.11.1-otp-23
+erlang 23.1.2
+nodejs 14.5.0


Malian:
@davydog187 I do not see this file in projects like Phoenix, Ecto, Jason, ...An the other hand, I have seen it in Bamboo. Could you explain to us what are the advantages and disadvantages of adding such a file from your point of view?

davydog187:
Yea of course!

## Pros
`.tool-versions` is a file read by the [asdf](https://github.com/asdf-vm/asdf) package manager that is used to switch between different [Elixir versions](https://github.com/asdf-vm/asdf-elixir). 

I actually had thought it was used a lot more in the popular Elixir packages, but I see its not, although I do know, anecdotely, it is used by a lot of people. If you're using `asdf`, it will switch you to the correct elixir versioin as prescribed by the `.tool-versions` files.

## Cons
The downside is having to keep it in sync with your CI (your CI and the tool versions file should probably be the same).

## Why did I decide to open this PR?
Currently, I'm running Elixir 1.12.0-rc.0 locally. When I run `mix format` in this project, it changes a number of lines because there is some difference between the formatter in 1.11.1. Having this file can be helpful to developers who may not know the exact Elixir version this project is developed against

Malian:
@davydog187 Thanks for the explanations

I am not against your proposition. Find below some through about it.

#### Pros:
- I like the idea to enforce the version for people working on the same project. I believe it could simplify things.

#### Cons/Questions:
- It will work only for people who use `asdf` as a package manager, so the problem you described above with the formatted is not fully solved.
- The minimum elixir version is defined in the `mix.exs` file, is this not redundant?
- People who would like to use different OTP/Erlang/Elixir version, will maybe not understand why it does not work.
- If I would like to test the latest elixir version, 1.12.0-rc.0, it means that I have to modify the file. Does it mean that we should bump this file, and the CI one, to the latest version as soon as the release is available?

As I said, I am not against :sweat_smile: but I do not have a strong opinion. I could leave with both choices. @msaraiva @miguel-s @lnr0626 WDYT?

davydog187:
> The minimum elixir version is defined in the mix.exs file, is this not redundant?
Its not redundant because this wouldn't solve the problem. In my case, even if the minimum was set to `1.11.4`, I would have experienced the issue still since I'm on `1.12.0`

> People who would like to use different OTP/Erlang/Elixir version, will maybe not understand why it does not work.
I'm not sure what you mean

> If I would like to test the latest elixir version, 1.12.0-rc.0, it means that I have to modify the file. Does it mean that we should bump this file, and the CI one, to the latest version as soon as the release is available?
It's really a question of what you're looking to support as a library. For something like this project, I would probably expect you to be more on the cutting edge, but maybe if you were more conservative, you would want to make sure the project works well on `1.9.0` for example. This is also a good reason to use the `matrix` features of Github actions to test on multiple versions, and then this part of the problem partially becomes moot.

davydog187:
I updated the matrix https://github.com/surface-ui/surface/pull/351

davydog187:
Now that I've seen the CI matrix, the problem that I particularly see is that the formatter can differ between elixir versions. In this projects case, you check if the project is formatted against Elixir 1.11.1 (currently), as defined by CI.

davydog187:
Hello @Malian did you have any more thoughts here?

I'm not going to die on the asdf hill, but I do think its valuable. If there's a preference not to add to the project, thats fine by me, but I'd rather close than leave this open. Let me know, thank you :)

msaraiva:
Hey @davydog187!

Thanks for the PR. I think we should avoid adding config files for external tooling. Otherwise, in no time, we'll have dozens of those files around. I remember people asking to include Vs Code related setting and other stuff I can't remember. 

Regarding the formatter, I think the rule should be to always be compatible with the latest stable version. So as soon as `v1.12.0` is out, we should update it and run the formatter. WDYT?

davydog187:
Sounds good to me. i'll close

PR #350, opened 2021-04-23
Rename Button to better align with Phoenix HTML

miguel-s:
Follow-up to discussions started here #342 and here #344.

### Summary

The naming of Surface's `<Button>` component leads to confusion as it looks like a wrapper for a generic html `<button>`, but is actually based on Phoenix's `Phoenix.HTML.Link.button/2` (thanks @mcrumm for pointing this out!).

`Phoenix.HTML.Link.button/2` does not wrap a generic html `<button>`, it is a special helper for creating buttons that submit data to the server.

Here examples from Phoenix's documentation:

> button("hello", to: "/world")
#=> `<button class="button" data-csrf="csrf_token" data-method="post" data-to="/world">hello</button>`

> button("hello", to: "/world", method: :get, class: "btn")
#=> `<button class="btn" data-method="get" data-to="/world">hello</button>`

### Reasons for this PR

We should rename Surface's `<Button>` to avoid confusion and help users better understand the purpose of this component.

Naming Suggestions:
1. Surface.Components.LinkButton
2. Surface.Components.Link.Button

I went with 1. as it's close enought to the Phoenix name, but also fits better into our module structure, but I'm open to discussion/suggestions.


Malian (approved):

msaraiva:
@miguel-s changes look good to me, however, wouldn't be more intuitive to follow phoenix's naming as:

 `Phoenix.HTML.Link.button` -> `Surface.Components.Link.Button`

It's not a strong opinion actually. It's more like a hunch :)

msaraiva:
And just to make it clear, I would prefer `LinkButton` if we weren't trying to map Phoenix's components.

miguel-s:
@msaraiva I did think about it. But then should we also change the `<Link>` component, which is based on `Phoenix.HTML.Link.link`?

Do we keep it as `Surface.Components.Link` or should we be strict and change it to `Surface.Components.Link.Link` (looks a bit silly 😅) to match the Phoenix helper, too?

msaraiva:
Oh crap. Let's leave it as `LinkButton` then :)

paulstatezny:
Personally I think `LinkButton` will make it less intuitive to newbies that it mirrors Phoenix's `button` helper.

I'm training multiple people on Phoenix/LiveView/Surface at work so this opinion doesn't come from a vacuum.

miguel-s:
@paulstatezny great to have an opinion from developers in training 👍

So looks we have 3 options for naming these components:

#### Option 1: Strict mapping of Phoenix to Surface
`Phoenix.HTML.Link.link/2` -> `Surface.Components.Link.Link`
`Phoenix.HTML.Link.button/2` -> `Surface.Components.Link.Button`

#### Option 2
`Phoenix.HTML.Link.link/2` -> `Surface.Components.Link`
`Phoenix.HTML.Link.button/2` -> `Surface.Components.LinkButton`

#### Option 3
`Phoenix.HTML.Link.link/2` -> `Surface.Components.Link`
`Phoenix.HTML.Link.button/2` -> `Surface.Components.Link.Button`

Either 2. or 3. work for me. @msaraiva @Malian @paulstatezny @lnr0626 what's your vote?

lnr0626:
I have a slight preference for 3

Malian:
I also have a slight preference for 3.

msaraiva:
Let's go with 3 then :)

PR #353, opened 2021-04-26
Fix missing quotations in docs

imsekun:


Malian:
@imsekun Thanks! :heart:

PR #354, opened 2021-04-26
Add support for phx-value-*

miguel-s:
Related to #330.

Discussion about the name of the prop is still open. Options so far are:
- `values`
- `event_values`
- `events_values`

Please let me know if you have any other suggestions and which option you prefer.

### DIFF
--- a/lib/surface/components/events.ex
+++ b/lib/surface/components/events.ex
@@ -32,6 +32,9 @@ defmodule Surface.Components.Events do
 
       @doc "Triggered when a key on the keyboard is pressed"
       prop keydown, :event
+
+      @doc "List values that will be sent as part of the payload triggered by an event"
+      prop values, :keyword, default: []
     end
   end
 end
--- a/lib/surface/components/utils.ex
+++ b/lib/surface/components/utils.ex
@@ -126,6 +126,10 @@ defmodule Surface.Components.Utils do
     |> List.flatten()
   end
 
+  defp values_to_opts([]) do
+    []
+  end
+
   defp values_to_opts(values) when is_list(values) do
     {:values, values}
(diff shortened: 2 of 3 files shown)


harmon25:
I like just `values` over `event_values` and `events_values`.

`phx_values` would be my more explicit suggestion, if we wanted to go in that direction.

msaraiva:
@miguel-s we'll eventually have to create a related `:values` directive to be able to do the same thing for HTML tags since we don't use attributes to do that in regular HTML nodes, only directives. 

Do you think it would be confusing to users to have both `values` and `:values`? If so, is there any alternative? I can't see any so I believe we'll have to provide both and make sure we document clear enough that `values` is just a custom property defined some of the built-in components (not a general use prop that can be used anywhere) and `:values`, a directive that applies only to HTML elements.

Thoughts?

miguel-s:
@msaraiva for this specific case I think we should go ahead. It mirrors how we handle events currently (e.g. `:on-click` & `click`).

However, the difference between props and directives seems to be of the more common questions for new Surface users. 
For example a common question I see is: Why does `<Button :on-click="foo" />` not work?

miguel-s:
@harmon25 thanks for adding your voice.

In Surface we try to avoid the `phx` prefix, e.g. `phx-click` (LiveView) and `:on-click` (Surface).

msaraiva:
> However, the difference between props and directives seems to be a common question for new Surface users.
> For example a common question I see is: Why does `<Button :on-click="foo" />` not work?

Agreed. We should probably do a better job with the docs.

msaraiva (approved):

Malian (approved):

PR #355, opened 2021-04-28
Support Range in :values

davydog187:
This PR adds support to prop `:values` such that a range can be used to communicate the possible values.

### Example
```elixir
defmodule MyComponent do
  use Surface.Component
  
  prop age, :integer, values: 1..100, default: 1
  
  ...
end
```

### Considerations
This adds `is_struct/2` as a guard, which was added in Elixir 1.11. I'm not sure which version we're looking to support. If we need to support earlier versions, then I believe I'll need to refactor `validate_opt/7` as the way it currently works only supports rejecting values

### Follow ups
If this PR is accepted, I will also PR to update the docs and the catalogue to support this new feature

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,15 +14,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - elixir: 1.8.2
-            otp: 21.3.8.17
-          - elixir: 1.9.4
-            otp: 21.3.8.17
-            warnings_as_errors: true # not 1.10 as its --warnigs-as-errors has bugs https://github.com/elixir-lang/elixir/issues/10073
-          - elixir: 1.10.4
-            otp: 21.3.8.17
-          - elixir: 1.10.4
-            otp: 23.1.1
           - elixir: 1.11.1
             otp: 21.3.8.17
           - elixir: 1.11.4
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -421,9 +421,11 @@ defmodule Surface.API do
   end
 
   defp validate_opt(_func, _name, _type, _opts, :values, value, _caller)
-       when not is_list(value) do
+       when not is_list(value) and not is_struct(value, Range) do
     {:error,
-     "invalid value for option :values. Expected a list of values, got: #{inspect(value)}"}
+     "invalid value for option :values. Expected a list of values or a Range, got: #{
+       inspect(value)
+     }"}
   end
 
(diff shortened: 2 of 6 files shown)


miguel-s (commented):
Looks good to me!

You'll need to update the error message in the exisiting test at `api_test.exs:46`.

davydog187:
@miguel-s done, and changed the minimum supported elixir version to 1.11, and updated the CI matrix to reflect.

Per slack, @msaraiva seems ok with this.

msaraiva:
Hi @davydog187!

Thanks for the PR. Do you know if this requires any change in [surface_catalogue](https://github.com/surface-ui/surface_catalogue) and [surface_site](https://github.com/surface-ui/surface_site)? Since both present the list of values in the docs, we need to make sure it doesn't break anything.

image

davydog187:
@msaraiva yes it will! I mentioned it in the description, you may have missed it. I'll add those PRs now

msaraiva:
@davydog187 yes. I missed it. Sorry. 😁

miguel-s (approved):
Thanks @davydog187 ❤️

msaraiva:
@davydog187 since this one sets the minimum version of Elixir to `v1.11`, I'll leave it out of `v0.4`. We can merge it before `v0.5`, which should not take long as we already have the new changes in the syntax approved and the work has already begun.

davydog187:
I have a similar one coming. No worries

davydog187:
Updated the documentation https://github.com/surface-ui/surface_site/pull/33

msaraiva:
Hey @davydog187!

Since we already have an "official" branch for `v0.5`, which is https://github.com/surface-ui/surface/tree/ms-use-custom-tokenizer, it would be nice to merge this there. Can you submit this PR against that branch instead of master? You can do the same for any follow-up PR you might have that should be applied to `v0.5.0`.

Cheers.

davydog187:
You got it!

msaraiva:
@davydog187 quick correction. Please use https://github.com/surface-ui/surface/tree/surface-next as the target for any PR that should be applied to `v0.5`.  Sorry for the last-minute change. Thanks.

davydog187:
I probably won't get to it until Monday. No worries, thanks for the update

davydog187:
@msaraiva rebased with `surface-next` and set that branch as the target of this PR.

davydog187:
Once this is merged i'll do the same with my other PR as it will have merge conflicts

msaraiva:
@davydog187 thanks! ❤️

PR #357, opened 2021-04-29
Add :values directive

miguel-s:
Related to #354.

Adds the `:values` directive for generating `phx-value-*` attributes.

#### Example
```elixir
<div :values={{ hello: :world, foo: "bar", one: 2, yes: true }}>
  Some Text
</div>
```
```html
<div phx-value-foo="bar" phx-value-hello="world" phx-value-one="2" phx-value-yes="true">
  Some Text
</div>
```

### DIFF
--- a/lib/surface/directive/values.ex
+++ b/lib/surface/directive/values.ex
@@ -30,17 +30,7 @@ defmodule Surface.Directive.Values do
         for {name, value} <- unquote(value) || [],
             attr_name = :"phx-value-#{name}",
             not Enum.member?(unquote(Macro.escape(attr_names)), attr_name) do
-          unless String.Chars.impl_for(value) do
-            message = """
-            invalid value for key "#{inspect(name)}" in attribute ":values".
-
-            Expected a type that implements the String.Chars protocol (e.g. string, boolean, integer, atom, ...), \
-            got: #{inspect(unquote(value))}\
-            """
-
-            IOHelper.runtime_error(message)
-          end
-
+          unquote(__MODULE__).validate_value!(name, value)
           {attr_name, {Surface.TypeHandler.attribute_type_and_opts(attr_name), to_string(value)}}
         end
       end
@@ -64,5 +54,18 @@ defmodule Surface.Directive.Values do
       value: Surface.TypeHandler.expr_to_quoted!(value, ":values", :map, meta),
       meta: meta
     }
+  end
+
+  def validate_value!(name, value) do
+    unless String.Chars.impl_for(value) do
+      message = """
+      invalid value for key "#{inspect(name)}" in attribute ":values".
+
+      Expected a type that implements the String.Chars protocol (e.g. string, boolean, integer, atom, ...), \
+      got: #{inspect(value)}\
+      """
+
+      IOHelper.runtime_error(message)
+    end
   end
 end
--- a/lib/surface/directive/values.ex
+++ b/lib/surface/directive/values.ex
@@ -1,0 +1,56 @@
+defmodule Surface.Directive.Values do
+  use Surface.Directive
+
+  def extract({":values", {:attribute_expr, value, expr_meta}, attr_meta}, meta) do
+    expr_meta = Helpers.to_meta(expr_meta, meta)
+    attr_meta = Helpers.to_meta(attr_meta, meta)
+
+    %AST.Directive{
+      module: __MODULE__,
+      name: :values,
+      value: directive_value(value, expr_meta),
+      meta: attr_meta
+    }
+  end
+
+  def extract(_, _), do: []
+
+  def process(
+        %AST.Directive{value: %AST.AttributeExpr{value: value} = expr, meta: meta},
+        %type{attributes: attributes} = node
+      )
+      when type in [AST.Tag, AST.VoidTag] do
+    attr_names =
+      attributes
+      |> Enum.filter(fn
+        %AST.Attribute{} -> true
+        _ -> false
+      end)
+      |> Enum.map(fn %AST.Attribute{name: name} -> name end)
+
+    new_expr =
+      quote generated: true do
+        for {name, value} <- unquote(value) || [],
+            attr_name = :"phx-value-#{name}",
+            not Enum.member?(unquote(Macro.escape(attr_names)), attr_name) do
+          {attr_name, {Surface.TypeHandler.attribute_type_and_opts(attr_name), to_string(value)}}
+        end
+      end
+
+    %{
+      node
+      | attributes: [
+          %AST.DynamicAttribute{name: :attrs, meta: meta, expr: %{expr | value: new_expr}}
+          | attributes
+        ]
+    }
+  end
+
+  defp directive_value(value, meta) do
+    %AST.AttributeExpr{
+      original: value,
+      value: Surface.TypeHandler.expr_to_quoted!(value, ":values", :map, meta),
+      meta: meta
+    }
+  end
+end
--- a/lib/surface/directive/values.ex
+++ b/lib/surface/directive/values.ex
@@ -16,23 +16,31 @@ defmodule Surface.Directive.Values do
   def extract(_, _), do: []
 
   def process(
-        %AST.Directive{value: %AST.AttributeExpr{value: value} = expr, meta: meta},
+        %AST.Directive{
+          value: %AST.AttributeExpr{value: value} = expr,
+          meta: meta
+        },
         %type{attributes: attributes} = node
       )
       when type in [AST.Tag, AST.VoidTag] do
-    attr_names =
-      attributes
-      |> Enum.filter(fn
-        %AST.Attribute{} -> true
-        _ -> false
-      end)
-      |> Enum.map(fn %AST.Attribute{name: name} -> name end)
+    attr_names = for %AST.Attribute{name: name} <- attributes, do: name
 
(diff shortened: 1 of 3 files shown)


lib/surface/directive/values.ex
```diff
@@ -0,0 +1,56 @@
...
+        _ -> false
+      end)
+      |> Enum.map(fn %AST.Attribute{name: name} -> name end)
+
+    new_expr =
+      quote generated: true do
+        for {name, value} <- unquote(value) || [],
+            attr_name = :"phx-value-#{name}",
+            not Enum.member?(unquote(Macro.escape(attr_names)), attr_name) do
+          {attr_name, {Surface.TypeHandler.attribute_type_and_opts(attr_name), to_string(value)}}
```
miguel-s, line 36: @lnr0626 I had to convert the value with `to_string/1` so booleans don't get swallowed. Without it `:values={{ yes: true }}>` would turn into `phx-value-yes` instead of `phx-value-yes="true"`.

Is this ok or is there a better solution?

msaraiva: @miguel-s LV converts any value that is sent from the client to the server into strings. In a perfect world, we should validate the values to force the user to pass a keyword list with all values as strings or at least emit a warning when that's not the case. Otherwise, the user may face unexpected behaviour if we automatically convert it. For instance:

```
<div :values={{ doit: false }}>
```

In this case, the user may expect to receive a boolean on `handle_event` and do something like:

```elixir
def handle_event("myevent", %{"doit" => doit}, ...) do
  if doit do
    ...
  else
    ...
  end
end
```

but the value will always be truthy. The same thing will happen when passing variables or assigns that can be `nil`. The user might test if the value is falsy, but it can never be.

We could perform a very nice validation at compile time if the user explicitly passes the keys and the values as literals or even go further and validate the values if they are assigns, e.g. `<div :values={{ id: @id }}` but that would be way too much for now. A runtime warning could be enough. We may also need to validate nested keyword/map as they are not supported. I believe if the user tries that, it would raise that Char protocol message at runtime, right?

Having said that, I confess I'm not sure what's the best way to handle that. Forcing the user to do something like `<div :values={{ id: to_string(@id) }}` or `<div :values={{ id: "#{@id}" }}` also sucks.

Any other suggestions?

A plugin, like we've been discussing in https://github.com/surface-ui/surface/pull/336 could convert the values to the right type reading some kind of specs:

```elixir
@values %{"doit" => boolean, "how_many_times" => integer}
def handle_event("myevent", %{"doit" => doit, how_many_times: times}, ...) do
  ...
end
```

But we don't have that yet and even if we had, we'd still need a solution that could work nicely without it so I'm not even sure why I'm bringing this idea in the first place. Sorry about that. I guess I got carried away :)

miguel-s: @msaraiva wow great feedback, thanks!

#### Type conversion
Good point. It could be an issue particularly for newer devs. Also it goes against the new direction of no magical type conversions in Surface.

My only counterpoints to this argument are:
1. The LV docs are quite clear that `phx-value-*` and `handle_event/3` deal with string values. Should we rely on Surface users to have read the LV docs?
2. We already allow something like this `<button :on-click="my_event" phx-value-id={{ 100 }}>Hi</button>` that does type conversion to string.

For me personally, if we can't think of another solution, then converting the values to string is a lesser evil than forcing users to write this `<div :values={{ id: to_string(@id) }}`.

#### Compiletime/Runtime validations
> We could perform a very nice validation at compile time if the user explicitly passes the keys and the values as literals or even go further and validate the values if they are assigns

Will try it and see how far I get, it's pretty new territory for me 😄 

> We may also need to validate nested keyword/map as they are not supported. I believe if the user tries that, it would raise that Char protocol message at runtime, right?

Good catch, will add validations for keyword/maps.

#### Plugins
Yes, please!

msaraiva: > 2\. We already allow something like this `<button :on-click="my_event" phx-value-id={{ 100 }}>Hi</button>` that does type conversion to string.

@miguel-s In that case, let's convert it to string to keep it consistent 👍

> Will try it and see how far I get, it's pretty new territory for me 😄

No need to do anything fancy for now. Since we're converting to string, we can keep the same behaviour as in `phx-value-id={{ 100 }}`, which I believe is to validate at runtime here: https://github.com/surface-ui/surface/blob/36b16e50c134b3e69f18052bab4059b91b596a74/lib/surface/type_handler/default.ex#L47-L56

The only difference is that we can validate each value in the keyword/map and check if they implement `String.Chars` individually, providing a better error them just `(Protocol.UndefinedError) protocol String.Chars not implemented for %{} of type Map`. BTW, I'm assuming that's what would happen but it would be better to confirm that. Maybe LV is already doing that check now? Have you tried to pass something like `<div :values={{ a: "A", b: %{} }}`? Do we get an error saying `:b` is invalid or just the generic `Protocol.UndefinedError` thing?

miguel-s: Looks like I was on the right track 😃 

> Do we get an error saying :b is invalid or just the generic Protocol.UndefinedError thing?

Passing a **map** raises the generic `Protocol.UndefinedError`

Passing a **list** is partially supported and raises a nice `ArgumentError`

> ** (ArgumentError) cannot convert the given list to a string.
>
> To be converted to a string, a list must either be empty or only
> contain the following elements:
> 
>   * strings
>   * integers representing Unicode code points
>   * a list containing one of these three elements
> 
> Please check the given list or call inspect/1 to get the list representation, got:
> 
> [hello: :world, foo: "bar", one: 2, yes: true]

miguel-s: @msaraiva added validation and test. Feedback is welcome 🙂

lib/surface/directive/values.ex
```diff
@@ -0,0 +1,68 @@
...
+            not Enum.member?(unquote(Macro.escape(attr_names)), attr_name) do
+          unless String.Chars.impl_for(value) do
+            message = """
+            invalid value for key "#{inspect(name)}" in attribute ":values".
+
+            Expected a type that implements the String.Chars protocol (e.g. string, boolean, integer, atom, ...), \
+            got: #{inspect(unquote(value))}\
+            """
+
+            IOHelper.runtime_error(message)
```
msaraiva, line 41: Maybe you could move this code to a separate function, similar to what we do here: https://github.com/surface-ui/surface/blob/71048b56bc391e1e115219cafea5b2934c521d76/lib/surface/directive/events.ex#L53

This way we avoid generating too much code inside `quote`, which makes it easier to debug when things go wrong.

You could call it `validate_values!`.

test/directives_test.exs
```diff
@@ -755,6 +755,90 @@ defmodule Surface.DirectivesTest do
...
+    end
+
+    test "passing unsupported types" do
+      assert_raise(RuntimeError, ~r(invalid value for key ":map" in attribute ":values".), fn ->
+        render_surface do
+          ~H"""
+          <div :values={{ map: %{}, tuple: {} }}>
+            Some Text
+          </div>
+          """
```
msaraiva, line 801: Can you try using something like:

```elixir
  render_surface do
    ~H"""
    <div :values={{ 
      map: %{}, 
      tuple: {} }}
    >
      Some Text
    </div>
    """
  end
```

This way we can make sure the error is pointing to the right line by using something like:

```elixir
assert_raise(RuntimeError, ~r(code:2: invalid value for key ":map" in attribute ":values".) 
```

msaraiva (commented):
@miguel-s looks good! I left just a couple of suggestions for your consideration.

test/directives_test.exs
miguel-s: I believe we only get the code line `code:2` on a `CompileError`, but not on a `RuntimeError`. Am I missing something?

lib/surface/directive/values.ex
miguel-s: Done!

I must admit that I did try and fail to extract this to a separate function before your hint... quote/unquote is still mindbending to me...

test/directives_test.exs
msaraiva: Yeah, you're right. Sorry.

msaraiva:
@miguel-s now we need to decide where to merge this. 

Version `v0.4.0` has been released today. We can either merge this to `master` along with any other non-breaking change and release `v0.4.1` or merge it directly to `v0.5.0`, which is being developed in https://github.com/surface-ui/surface/tree/ms-use-custom-tokenizer. 

What do you prefer? I'd probably apply any new feature to `v0.5.0`, however, you'd have to rebase your branch to retrieve the new stuff and adjust the tests to use `{ }` instead of `{{ }}`. What do you prefer?

miguel-s:
Releasing with `v0.5.0` works for me, don't think this is an urgent feature.

We could even wait for `v0.6.0` if you want to keep `v0.5.0` focused on the syntax changes?

I'll do the same for #354.

msaraiva:
> We could even wait for `v0.6.0` if you want to keep `v0.5.0` focused on the syntax changes?

No need to wait for `v0.6.0`. It's fine to add both to `v0.5.0`.

msaraiva:
@miguel-s quick update, let's use https://github.com/surface-ui/surface/tree/surface-next as the target branch for PRs related to `v0.5`.

msaraiva (approved):

PR #361, opened 2021-04-30
Allow ranges to be passed for :list props

davydog187:
Similar to #355, allows ranges to be passed for `:list` values. This is particularly useful when you have a fixed number of components that would be generated from within the prop. I found the need for this when implementing a `RadioGroup` component that composes `Surface.Components.Form.RadioButton`

There is an argument for allowing any enumerable to be passed as a list, I'm not sure if that's a good idea or not.

## Example

```elixir
<MyComponent list={ 1..4 } />
```

### DIFF
--- a/lib/surface/type_handler/list.ex
+++ b/lib/surface/type_handler/list.ex
@@ -32,7 +32,7 @@ defmodule Surface.TypeHandler.List do
     quote generated: true do
       case unquote(expr) do
         value when is_list(value) or is_struct(value, Range) ->
-          value
+          Enum.to_list(value)
 
         value ->
           raise "invalid value for property \"#{unquote(name)}\". Expected a :list, got: #{
--- a/lib/surface/type_handler/list.ex
+++ b/lib/surface/type_handler/list.ex
@@ -31,7 +31,7 @@ defmodule Surface.TypeHandler.List do
   defp handle_list_expr(name, expr) do
     quote generated: true do
       case unquote(expr) do
-        value when is_list(value) ->
+        value when is_list(value) or is_struct(value, Range) ->
           value
 
         value ->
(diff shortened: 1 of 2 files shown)


davydog187:
updated the CI matrix as well here to only support 1.11 and up

Malian:
LGTM.

@msaraiva If I well understood, we should merge this to the https://github.com/surface-ui/surface/tree/surface-next branch?

davydog187:
@msaraiva rebased this against `surface-next` and changed the target to that branch as well

lib/surface/type_handler/list.ex
```diff
@@ -31,7 +31,7 @@ defmodule Surface.TypeHandler.List do
   defp handle_list_expr(name, expr) do
     quote generated: true do
       case unquote(expr) do
-        value when is_list(value) ->
+        value when is_list(value) or is_struct(value, Range) ->
           value
```
msaraiva, line 35: @davydog187 maybe we should call `Enum.to_list/1` to make sure the value will be a list (unless this is already done somewhere else?). After all, the value is defined as `:list` so the user might have code depending on the value being a list, e.g. `List.first/1` and friends. This is similar to what we do when accepting keyword lists in props defined as `:map`, we accept it as syntactic sugar but always convert it to map. WDYT?

davydog187: Seems reasonable. Effectively you're suggesting doing an implicit cast so that when it crosses the component boundary, it's in the type expected. I buy that

msaraiva: Exactly! If the user actually needs a range, then it's better to explicitly define the prop as `:range`.

davydog187: Ok, updated the code, feels like it should be tested though! Is `compiler_test.exs` the best place to test this?

davydog187: Done. Now the tests make more sense too. Good call @msaraiva

msaraiva:
@davydog187 thanks! ❤️

PR #362, opened 2021-04-30
Deprecate string to atom conversion

msaraiva:
After this gets merge I'll release `v0.4.0`.

### DIFF
--- a/lib/surface/components/form/field_context.ex
+++ b/lib/surface/components/form/field_context.ex
@@ -12,7 +12,7 @@ defmodule Surface.Components.Form.FieldContext do
   alias Surface.Components.Form.Field
 
   @doc "The field name"
-  prop name, :atom, required: true
+  prop name, :any, required: true
 
   @doc """
   The content for the field
--- a/lib/surface/components/form/label.ex
+++ b/lib/surface/components/form/label.ex
@@ -23,7 +23,7 @@ defmodule Surface.Components.Form.Label do
   prop form, :form
 
   @doc "The field name"
-  prop field, :atom
+  prop field, :any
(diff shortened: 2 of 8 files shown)


lib/surface/components/form/field_context.ex
```diff
@@ -12,7 +12,7 @@ defmodule Surface.Components.Form.FieldContext do
   alias Surface.Components.Form.Field
 
   @doc "The field name"
-  prop name, :atom, required: true
+  prop name, :any, required: true
```
msaraiva, line 15: @miguel-s I believe this change makes it consistent with the changes in https://github.com/surface-ui/surface/pull/319/. Correct? Is there anything else to be done?

lib/surface/components/form/label.ex
```diff
@@ -23,7 +23,7 @@ defmodule Surface.Components.Form.Label do
   prop form, :form
 
   @doc "The field name"
-  prop field, :atom
+  prop field, :any
```
msaraiva, line 26: Same here. Confirm?

Malian (approved):

lib/surface/components/form/field_context.ex
miguel-s: Looks good to me 👍 

We might want to add a test to `field_contest_test.exs` to check `name` with atom, we only have one for strings now.

lib/surface/components/form/label.ex
miguel-s: Also 👍 

Same thing with the test.

miguel-s (approved):

PR #364, opened 2021-05-02
Implement shorthand notation for assigning slots

Malian:
This pull request implements the new syntax for assigning slots:

Instead of 
```
<Card>
  <template slot="header">
    ...
  </template>
  <template slot="footer">
    ...
  </template>
</Card>
```
You will be able to use the following syntax:
```
<Card>
  <:header>
    ...
  </:header>
  <:footer>
    ...
  </:footer>
</Card>
```

@msaraiva a couple of questions still need an answer:
- Should we deprecate the syntax `<#template name="">`?
- Should we use the new syntax in the built-in Surface components?
- Should we use the new syntax in the tests that are not directly related to the old syntax?
- Should I wrote a convert function to handle convert old syntax to the new one? Which one are the old one, `<template>` or `<#template>` or both?

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -629,7 +629,9 @@ defmodule Surface.Compiler do
 
   defp validate_tag_children([%AST.Template{name: name} | _]) do
     {:error,
-     "templates are only allowed as children elements of components, but found template for #{name}"}
+     "templates are only allowed as children elements of components, but found template for #{
+       name
+     }"}
   end
 
   defp validate_tag_children([_ | nodes]), do: validate_tag_children(nodes)
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -230,6 +230,7 @@ defmodule Surface.Compiler do
   defp node_type({"#template", _, _, _}), do: :template
   defp node_type({"#slot", _, _, _}), do: :slot
   defp node_type({"template", _, _, _}), do: :template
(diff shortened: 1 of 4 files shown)


msaraiva:
@Malian thanks for working on this! ❤️

I'm really excited to finally see `v0.5` taking shape.

> * Should we deprecate the syntax `<#template name="">`?

Not really. I believe, at least for now, we should consider the new one as just a special shorthand syntax. We can review that in the future after getting some feedback from users.

> * Should we use the new syntax in the built-in Surface components?

I don't think we should do that now, especially because the `<#...>` variants will be kept. This way we can focus on more immediate issues and update the components later. For new components, however, we can certainly use the shorthand version.

> * Should we use the new syntax in the tests that are not directly related to the old syntax?

For new stuff, sure. But for the general existing tests, no. For the same reasons mentioned above.

> * Should I wrote a convert function to handle convert old syntax to the new one? Which one are the old one, `<template>` or `<#template>` or both?

I did that already:

https://github.com/surface-ui/surface/blob/ffc5c7b819fefcbb7d0461ad7a66293ced7b3518/lib/surface/compiler/converter_0_5.ex#L18-L24

We just need to uncomment that code and the related tests :)

We'll not convert it to the shorthand syntax. This will be done by `surface_formatter` which can take user preferences into account. The converter is supposed to be as simple as possible, just translates old code into new valid code.

msaraiva (approved):

Issue #340, opened 2021-04-19
Syntax changes and deprecations for `v0.4` and `v0.5`


msaraiva:
I'm working on a proposal for a new unified syntax for built-in constructs, slots, dynamic props/attrs and directives.

Any syntax-related discussion will be moved here and consolidated into a single proposal for v`0.5.0`. A couple of those changes might be introduced already in `0.4.0`, deprecating a few aspects of the current syntax. ~The first draft of the proposal should be available for general feedback soon.~

The full proposal can be found [here](https://gist.github.com/msaraiva/dd40ab537e1508f1af2954a19faa2d54).

We also have plans to create a mix task based on the Surface formatter to automatically apply those changes.

### v0.4.0 :white_check_mark:

  - [x] Remove support for auto-conversion of literal strings into atoms when the `prop` is of type `:atom`

### v0.5.0

  - [x] Replace `{{ }}` with `{ }`
  - [x] Replace `~H` with `~F` to avoid conflict with `HEEx`
  - [x] Introduce `<#template>` and `<#slot>` and deprecate `<template>` and `<slot>` with proper warnings
  - [x] Introduce shorthand notation for `<#template slot="my_slot">` -> `<:my_slot>`
  - [x] Introduce `{#if}` and deprecate `<If>` with proper warning
  - [x] Introduce `{#for}` and deprecate `<For>` with proper warning
  - [x] Introduce `{#unless}`
   - [x] Deprecate passing attribute values as `selected=true` or `tabindex=1`. Any non-string value should be passed as `selected={true}` (just like React)
  - [x] Introduce the concept of root properties, e.g. `<Component {@var}>`
  - [x] Introduce the new shorthand  notation `{...@attrs}` for dynamic attributes/properties ~~and deprecate `:props` and `:attrs` with proper warnings~~ (we can deprecate them in future versions after getting some feedback)
  - [x] Introduce the new shorthand notation for attribute assignment, e.g., `<div {=@class}>`
  - [x] Support `{#if}...{#elseif}...{#else}...{/if}`
  - [x] Support `{#case}...{#match}...{/case}`
  - [x] Support `{#for}...{#else}...{/for}`
  - [x] Introduce the `s-` prefix for directives as an alternative to `:`
[...]

PR #365, opened 2021-05-02
Introduce "s-" prefix for directives

miguel-s:
Related to #340.

Extends the Surface compiler to handle any number of prefixes for directives.

Prefixes can be added/removed with the `@directive_prefixes ` module attribute.

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -47,6 +47,8 @@ defmodule Surface.Compiler do
     Surface.Directive.For
   ]
 
+  @directive_prefixes [":", "s-"]
+
   @void_elements [
     "area",
     "base",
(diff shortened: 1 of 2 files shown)


Malian (approved):

lib/surface/compiler.ex
```diff
@@ -47,6 +47,8 @@ defmodule Surface.Compiler do
     Surface.Directive.For
   ]
 
+  @directive_prefixes [":", "s-"]
```
lnr0626, line 50: is this something we'd want to pull out into the config?

lnr0626 (commented):
lgtm

lnr0626 (approved):

lib/surface/compiler.ex
msaraiva: Maybe in the future. Let's start with this and see if anyone will ever use it :)

msaraiva (approved):

PR #366, opened 2021-05-03
refactoring parse error handling

lnr0626:
At a high level, this PR does the following:

- swapping out the mixture of raise/returning error tuples in the tokenizer/parser with returning error tuples and adding a bang variant (i.e. tokenize! and parse!)
- consolidating to a single ParseError struct (we had two serving the same purpose - Surface.Compiler.ParseError and Surface.Compiler.Tokenizer.ParseError)
- passing file/column/line options through the parser to the tokenizer so the line numbers were all consistent (and therefor removing the need for line_offset)
- propagating the line/column/file meta for attributes (I wasn't quite sure why we didn't have those initially, but it allows us to handle errors consistently regardless of where they come from)
- propagating column info into the ast meta (we didn't have this initially as it was added in 1.11 and we started the compiler work before that was released)
- fixed a TODO for the message we raise when a closing tag is missing
- replace all of our uses of line offset with just line

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -6,6 +6,7 @@ defmodule Surface.Compiler do
   """
 
   alias Surface.Compiler.Parser
+  alias Surface.Compiler.ParseError
   alias Surface.IOHelper
   alias Surface.AST
   alias Surface.Compiler.Helpers
@@ -68,15 +69,6 @@ defmodule Surface.Compiler do
     "wbr"
   ]
 
-  defmodule ParseError do
-    defexception file: "", line: 0, message: "error parsing HTML/Surface"
-
-    @impl true
-    def message(exception) do
-      "#{Path.relative_to_cwd(exception.file)}:#{exception.line}: #{exception.message}"
-    end
-  end
-
   defmodule CompileMeta do
     defstruct [:line, :file, :caller, :checks]
 
@@ -111,6 +103,9 @@ defmodule Surface.Compiler do
     |> case do
       {:ok, nodes} ->
         nodes
+
+      {:error, error} ->
+        raise error
 
       {:error, message, line} ->
         raise %ParseError{line: line, file: file, message: message}
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -98,7 +98,7 @@ defmodule Surface.Compiler do
     }
 
     string
-    |> Parser.parse!(file: file, line: line)
+    |> Parser.parse!(file: file, line: line, column: opts[:column] || 1)
     |> to_ast(compile_meta)
     |> validate_component_structure(compile_meta, caller.module)
   end
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -98,7 +98,12 @@ defmodule Surface.Compiler do
     }
 
     string
-    |> Parser.parse!(file: file, line: line, column: opts[:column] || 1)
+    |> Parser.parse!(
+      file: file,
+      line: line,
+      column: Keyword.get(opts, :column, 1),
+      indentation: Keyword.get(opts, :indentation, 0)
+    )
     |> to_ast(compile_meta)
     |> validate_component_structure(compile_meta, caller.module)
   end
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -78,10 +78,10 @@ defmodule Surface.Compiler do
   end
 
   defmodule CompileMeta do
-    defstruct [:line_offset, :file, :caller, :checks]
+    defstruct [:line, :file, :caller, :checks]
 
     @type t :: %__MODULE__{
-            line_offset: non_neg_integer(),
+            line: non_neg_integer(),
             file: binary(),
             caller: Macro.Env.t(),
             checks: Keyword.t(boolean())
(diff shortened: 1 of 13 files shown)


lnr0626:
this has some changes to the tokenizer, parser, and compiler to make our error handling a bit more consistent (we had a few places where we were raising errors, and then catching them to convert them into `{:error, error}` tuples, and the like). This also consolidates the different ParseError structs we generated from the tokenizer and the compiler, and enables passing info from the compiler through the parser to the tokenizer for file/line/columns

lmk what you think

i also cleaned up a few of the leftover TODOs around error messages, those commits should be fairly easy to extract if we decide we don't want to go down this path.

lib/surface/compiler/parser.ex
```diff
@@ -117,71 +124,76 @@ defmodule Surface.Compiler.Parser do
...
+    if parent_name in @sub_blocks_valid_parents[name] do
+      :ok
+    else
+      {:error,
+       %ParseError{
+         file: meta.file,
+         line: meta.line,
+         column: meta.column,
+         message: "cannot use <#{name}> inside <#{parent_name}>. #{valid_parents_str}"
+       }}
```
msaraiva, line 196: I think this is probably the only thing I'd avoid from this PR. Propagating errors from inner function calls in a recursive algorithm is probably one of the best usages of exceptions, IMO. In our case, since any exception always halts the parser, we don't have to worry about performance so I think the whole logic is much easier to follow without all those `with` that had to be introduced to handle the returned error.

lib/surface/compiler/parser.ex
```diff
@@ -206,43 +218,54 @@ defmodule Surface.Compiler.Parser do
...
       _ ->
-        message = "unexpected value for attribute \"#{name}\""
-        raise %ParseError{line: line, column: 1, message: message}
+        {:error,
+         %ParseError{
+           line: meta.line,
+           column: meta.column,
+           file: meta.file,
+           message: "unexpected value for attribute \"#{name}\""
+         }}
```
msaraiva, line 254: Same here. This makes the callers more polluted in order to handle the return value as well as the code in each `translate_attr` variant. I'd just leave it as an exception.

lib/surface/compiler/tokenizer.ex
```diff
@@ -415,14 +432,14 @@ defmodule Surface.Compiler.Tokenizer do
...
-  defp handle_attr_value_unquoted(<<c::utf8, _::binary>>, line, column, _buffer, _acc, state)
+  defp handle_attr_value_unquoted(<<c::utf8, _::binary>>, line, column, _buffer, _acc, _state)
        when c in @unquoted_value_invalid_chars do
     message = """
     unexpected character `#{<<c::utf8>>}`. \
     Unquoted attribute values cannot contain `\"`, `'`, `=` nor `<`
     """
 
-    raise %ParseError{file: state.file, line: line, column: column, message: message}
+    {:error, message, line, column}
```
msaraiva, line 442: I think the same idea applies to the tokenizer. I'd just use exceptions for most cases.

msaraiva (commented):
@lnr0626 great work! I left some comments for your consideration.

lib/surface/compiler.ex
```diff
@@ -981,7 +970,7 @@ defmodule Surface.Compiler do
       Hint: replace `<slot>` with `<#slot>`
       """
 
-      IOHelper.warn(message, compile_meta.caller, &(&1 + line))
+      IOHelper.warn(message, compile_meta.caller, fn _ -> line end)
```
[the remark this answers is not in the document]
lnr0626: i think that makes sense - i actually did make changes to just pass in a line instead of a function, but this had already grown quite a bit. i'll do those as a separate PR after this one

lib/surface/compiler/parser.ex
lnr0626: I've switched it over to just raising exceptions and agree that it simplifies the logic considerably - i think my issue wasn't the choice between the two, and it wasn't clear to my why we raised in one place and return ok/error tuples in another. lmk what you think about the new commits

Malian (approved):
@lnr0626 What an amazing PR!

I left a small comment. Feel free to move on regardless of it.

lib/surface/compiler/parser.ex
msaraiva: > it wasn't clear to my why we raised in one place and return ok/error tuples in another

That's my fault. I should've mentioned the reason in the last message. The reason is that in some cases, depending on what we're parsing, we may want to have some kind of "checkpoint" where we can identify the context of what is being parsed. For instance, the `handle_interpolation` function is reused in 3 different places, each one in different contexts:

1. Classical interpolation in the body

https://github.com/surface-ui/surface/blob/175a5427e9e7a83a0493b3c4bfa4fb4d7f3e12ce/lib/surface/compiler/tokenizer.ex#L77-L87

2. Root property expression

https://github.com/surface-ui/surface/blob/175a5427e9e7a83a0493b3c4bfa4fb4d7f3e12ce/lib/surface/compiler/tokenizer.ex#L260-L269

3. Attribute value expression

https://github.com/surface-ui/surface/blob/175a5427e9e7a83a0493b3c4bfa4fb4d7f3e12ce/lib/surface/compiler/tokenizer.ex#L434-L444

By returning a tuple here, we explicitly indicate to the caller that this call, unlike `handle_text`, must return the expected "complete" token, if not, it's up to the caller to properly handle the error, for instance, to provide a more specific message based on the context of the call (we're not doing that yet but we will). Or to ignore the line/column where the parsing error happened and emit it pointing to the beginning of the token, if that makes more sense to the user. 

This kind of fine-grained error handling is one of the things it was really hard to achieve with `nimble_parcec`. With the custom tokenizer, we have a chance to define those checkpoints where we know the call must return a specific token. That also makes it easier to reason about the behaviour of the tokenizer during the recursion. Function calls returning tuples will not move the flow to other top-level handlers. They are complete. They either return the expected token or an error. They can't call, for instance, `handle_text` again.

So, here are some questions that can help to identify those special token handlers:

  * Must the call return a specific token? Or can it move the flow back to an upper-level call (restarting the recursion)?
  * Can it be reused by other handlers in different contexts?
  * Could an internal parsing error provided by an inner call provide sub-optimal information (message or position) to the user?

If the answer to any of those question is "yes", then there's a big change we should return a tuple.

lnr0626: thanks, that explanation helps me understand your reasoning a lot better!

msaraiva (approved):
@lnr0626 thanks for the great work! Very nice improvements to the code ❤️

PR #367, opened 2021-05-03
fix column tracking for self-closing tags

lnr0626:
this fixes a small issue i found while working on #366 with our column tracking


msaraiva (approved):
❤️

miguel-s (approved):

PR #368, opened 2021-05-04
Introduce <#if>...<#elseif>...<#else>...</#if>

Malian:
Related to #340 

Todo:

- [x] Deprecate `<If>`
- [x] Rewrite test names
- [x] Add some tests to the compiler for the ast structure

### Design choices:

I decided to treat the `if`, `elseif`, `else` structure as a nested `if`, `else` structure. In a nutshell, the code converts the following structure:
```html
<#if condition={true}>
  IF TRUE
<#elseif condition={true}>
  ELSEIF TRUE
<#elseif condition={false}>
  ELSEIF FALSE
<#else>
  ELSE
</#if>
```
to
```html
<#if condition={true}>
IF TRUE
<#else>
  <#if condition={true}>
    ELSEIF TRUE
  <#else>
     <#if condition={false}>
       ELSEIF FALSE
     <#else>
       ELSE
     </#if>
   </#if>
</#if>
```

Additionnaly, I used the `AST.Container` to wrap the `else` block, but I am not sure it is the best way to do. We could introduce a new `AST.Else` struct. Ideas are welcome.

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -219,11 +219,18 @@ defmodule Surface.Compiler do
     end
   end
 
+  # Slots
   defp node_type({"#template", _, _, _}), do: :template
   defp node_type({"#slot", _, _, _}), do: :slot
   defp node_type({"template", _, _, _}), do: :template
   defp node_type({":" <> _, _, _, _}), do: :template
   defp node_type({"slot", _, _, _}), do: :slot
+
+  # Conditionnal blocks
+  defp node_type({"#if", _, _, _}), do: :if_elseif_else
+  defp node_type({"#elseif", _, _, _}), do: :if_elseif_else
+  defp node_type({"#else", _, _, _}), do: :if_elseif_else
+
   defp node_type({"#" <> _, _, _, _}), do: :macro_component
   defp node_type({<<first, _::binary>>, _, _, _}) when first in ?A..?Z, do: :component
   defp node_type({name, _, _, _}) when name in @void_elements, do: :void_tag
@@ -258,6 +265,54 @@ defmodule Surface.Compiler do
      %AST.Interpolation{
        original: text,
        value: expr,
+       meta: meta
+     }}
+  end
+
+  defp convert_node_to_ast(
+         :if_elseif_else,
+         {"#else", _attributes, children, node_meta},
+         compile_meta
+       ) do
+    meta = Helpers.to_meta(node_meta, compile_meta)
+
+    {:ok,
+     %AST.Container{
+       children: to_ast(children, compile_meta),
+       meta: meta,
+       directives: []
+     }}
+  end
+
+  defp convert_node_to_ast(
+         :if_elseif_else,
+         {_name, attributes, children, node_meta},
+         compile_meta
+       ) do
+    meta = Helpers.to_meta(node_meta, compile_meta)
+    default = %AST.AttributeExpr{value: false, original: "", meta: node_meta}
+    condition = attribute_value_as_ast(attributes, "condition", default, compile_meta)
+
+    [if_children, else_children] =
+      case children do
+        [{:default, [], default, _}, {"#else", _, _, _} = else_block] ->
+          [default, [else_block]]
+
+        [{:default, [], default, _}, {"#elseif", a, c, m} | rest] ->
+          [default, [{"#elseif", a, [{:default, [], c, %{}} | rest], m}]]
+
+        [{:default, [], default, _}] ->
+          [default, []]
+
+        children ->
+          [children, []]
+      end
+
+    {:ok,
+     %AST.IfElse{
+       condition: condition,
+       if: to_ast(if_children, compile_meta),
+       else: to_ast(else_children, compile_meta),
        meta: meta
      }}
   end
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -309,7 +309,7 @@ defmodule Surface.Compiler do
       end
 
     {:ok,
-     %AST.IfElse{
+     %AST.If{
        condition: condition,
        if: to_ast(if_children, compile_meta),
        else: to_ast(else_children, compile_meta),
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -435,7 +435,13 @@ defmodule Surface.Compiler do
     end
   end
 
-  defp convert_node_to_ast(:component, {name, attributes, children, node_meta}, compile_meta) do
+  defp convert_node_to_ast(
+         :component,
+         {name, attributes, children, node_meta} = node,
+         compile_meta
+       ) do
+    maybe_warn_on_deprecated_if_notation(node, compile_meta)
+
     # TODO: validate live views vs live components ?
     meta = Helpers.to_meta(node_meta, compile_meta)
     mod = Helpers.actual_component_module!(name, meta.caller)
@@ -1031,6 +1037,19 @@ defmodule Surface.Compiler do
       future versions.
 
       Hint: replace `<slot>` with `<#slot>`
+      """
+
+      IOHelper.warn(message, compile_meta.caller, fn _ -> line end)
+    end
+  end
+
+  defp maybe_warn_on_deprecated_if_notation({name, _, _, %{line: line}}, compile_meta) do
+    if name == "If" do
+      message = """
+      using <If> to wrap elements in an if experssion has been depreacated and will be removed in \
+      future versions.
+
+      Hint: replace `<If>` with `<#if>`
       """
 
       IOHelper.warn(message, compile_meta.caller, fn _ -> line end)
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -435,13 +435,7 @@ defmodule Surface.Compiler do
     end
   end
 
-  defp convert_node_to_ast(
-         :component,
-         {name, attributes, children, node_meta} = node,
-         compile_meta
-       ) do
-    maybe_warn_on_deprecated_if_notation(node, compile_meta)
-
+  defp convert_node_to_ast(:component, {name, attributes, children, node_meta}, compile_meta) do
     # TODO: validate live views vs live components ?
     meta = Helpers.to_meta(node_meta, compile_meta)
     mod = Helpers.actual_component_module!(name, meta.caller)
@@ -1037,19 +1031,6 @@ defmodule Surface.Compiler do
       future versions.
 
       Hint: replace `<slot>` with `<#slot>`
-      """
-
-      IOHelper.warn(message, compile_meta.caller, fn _ -> line end)
-    end
-  end
-
-  defp maybe_warn_on_deprecated_if_notation({name, _, _, %{line: line}}, compile_meta) do
-    if name == "If" do
-      message = """
-      using <If> to wrap elements in an if experssion has been depreacated and will be removed in \
-      future versions.
-
-      Hint: replace `<If>` with `<#if>`
       """
 
       IOHelper.warn(message, compile_meta.caller, fn _ -> line end)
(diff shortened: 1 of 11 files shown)


lnr0626:
> I decided to treat the if, elseif, else structure as a nested if, else structure

ideally we'd translate the if/elseif/else structure into a `cond do ... end` expression, though nested if/else is a good first pass as the cond is going to be a bit more invasive. I do think we should keep the AST flat and make the nested if structure an implementation detail of `eex_engine.ex` - that'll make it easier to change the code we generate from the ast in the future without having to change the compiler/etc.

> Additionnaly, I used the AST.Container to wrap the else block, but I am not sure it is the best way to do. We could introduce a new AST.Else struct. Ideas are welcome.

I haven't looked through in depth, but I may have some suggestions once i do that

---

my initial thought was to extend the current `AST.If` with an `else` property - essentially what you have as your IfElse struct. this doesn't allow us to maintain that flat structure, though we could add something like `:elseif` that's a list of `{condition, children}` tuples (or similar)

Another option would be to treat if and if/else as the same (AST.If with your updates to accept an else), and if/elseif/else as a new Cond node:

```
defmodule Surface.AST.Cond do
  @moduledoc """
  An AST node representing an if/else expression
  ## Properties
      * `:conditions` - a list of Surface.AST.Condition representing the if/elseif/elseif/else structure
      * `:meta` - compilation meta data
      * `:debug` - keyword list indicating when debug information should be printed during compilation
  """
  defstruct [:conditions, :meta, debug: []]

  @type t :: %__MODULE__{
          conditions: list(Condition.t()),
          debug: list(atom()),
          meta: Surface.AST.Meta.t()
        }

  defmodule Condition do
      @moduledoc """
        An AST node representing an if/else expression
        ## Properties
            * `:children` - a list of Surface.AST nodes representing the body of this condition
            * `:expression` - a Surface.AST.Expression representing the condition to evaluate
            * `:meta` - compilation meta data
        """
    defstruct [:expression, :children]
    @type t :: %__MODULE__{
        expression: Surface.AST.Expr.t(),
        children: list(Surface.AST.t()),
        meta: Surface.AST.Meta.t(),
    }
  end
end
```


that would result in the following ast structure:

```
<#if condition={true}>
  IF TRUE
<#elseif condition={true}>
  ELSEIF TRUE
<#elseif condition={false}>
  ELSEIF FALSE
<#else>
  ELSE
</#if>
```

```
%Cond{
  conditions: [
    %Cond.Condition{expression: %Expr{value: true}, children: [%Literal{value: "IF TRUE"}]},
    %Cond.Condition{expression: %Expr{value: true}, children: [%Literal{value: "ELSEIF TRUE"}]},
    %Cond.Condition{expression: %Expr{value: false}, children: [%Literal{value: "ELSEIF FALSE"}]},
    %Cond.Condition{expression: %Expr{value: true}, children: [%Literal{value: "ELSE"}]} # the else is translated into a condition with just true as the expression
  ],
  ...
}
```

for completeness, we'd also need to append a `%Cond.Condition{expression: %Expr{value: true}, children: [%Literal{value: nil}]}` (i think nil is accurate here) if the user doesn't have the final `else`. We could either do that in the compiler, or add some flag on the ast node to indicate it needs that.

test/compiler/converter_0_5_test.exs
```diff
@@ -107,28 +107,28 @@ defmodule Surface.Compiler.Converter_0_5Test do
...
-  #     <div>
-  #       <If condition={{ @var }}>
-  #         1
-  #         </If>
-  #       <If   condition={{@var}}>2</If>
-  #     </div>
-  #     """)
+    assert expected == """
+           <div>
+             <#if condition={@var}>
```
lnr0626, line 123: should we also automatically switch this to using the root prop? (i.e. `<#if {@var}>`)

test/constructs/if_test.exs
```diff
@@ -66,4 +66,321 @@ defmodule Surface.Constructs.IfTest do
...
+          """
+        end
+
+      assert html =~ """
+             <span>The inner content</span>
+             <span>with multiple tags</span>
+             """
+    end
+
+    test "renders inner else condition is falsly" do
```
lnr0626, line 145: small typo, should be `falsey` i think, maybe `falsy` - i'm honestly not sure which one 😂

lnr0626 (commented):
good work!

we'll also want to add some tests to the compiler for the ast structure once we decide how we want to handle that

lnr0626:
> Additionnaly, I used the AST.Container to wrap the else block...

that's pretty much exactly what the container node is for - basically just providing a wrapper for when we have a list of nodes that has some meta (and potentially directives) associated with it, but no other ast node is appropriate

test/compiler/converter_0_5_test.exs
msaraiva: That would not be possible with the current implementation of the converter. Since this is also valid syntax, I think we can keep it as it is for now. The `surface_formatter` on the other hand, can do this quite well.

msaraiva (commented):
@Malian great stuff!

I'm so glad we finally have that. I'm even more excited because the model is generic enough to implement all other planned constructs. Two things I think we should do as soon as we merge this:

1. Introduce a generic `Construct` AST node. The following suggestion from @lnr0626 on slack could be a starting point:

    ```elixir
    %AST.Construct{
      name: ...,
      expression: <quoted elixir>,
      body: [<AST nodes>],
      sub_blocks: [
        %AST.Construct.SubBlock{ # maybe unnecessary?
          name: ...,
          expression: nil | <quoted elixir>,
          body: [<AST nodes>],
          # the other properties common to ast nodes
        },
        ...
      ],
      # the other properties common to ast nodes
    }
    ```

    That could replace the existing nodes that represent constructs, i.e. `If`, `IfElse`, and and `For`, as well as the new `Case` or any other construct we come up with in the future. Depending on how it goes, we could even see if it's feasible to have a even more generic structure that would hold attributes instead of a single expression. That would allow replacing `Template` and `Slot` too.

2. Remove all implementation of constructs to their own file, similar to what we do with `For` and `If` components. Not sure yet if we need a new abstraction like `Surface.Construct` but the idea is the same. I think @lnr0626 already have some plans in that direction too.

Thoughts?

Malian:
@lnr0626 Thanks for the review!

> I do think we should keep the AST flat and make the nested if structure an implementation detail of eex_engine.ex - that'll make it easier to change the code we generate from the ast in the future without having to change the compiler/etc.

I agree. When I was debugging the AST quickly became a mess with nested `if/elseif/else` constructs.

> my initial thought was to extend the current AST.If with an else property

It was exactly my first implementation. I did not want to touch the actual `If` struct and `:if` directive for presenting the draft, but indeed we could now merge both.

```elixir
%Cond{
  conditions: [
    %Cond.Condition{expression: %Expr{value: true}, children: [%Literal{value: "IF TRUE"}]},
    %Cond.Condition{expression: %Expr{value: true}, children: [%Literal{value: "ELSEIF TRUE"}]},
    %Cond.Condition{expression: %Expr{value: false}, children: [%Literal{value: "ELSEIF FALSE"}]},
    %Cond.Condition{expression: %Expr{value: true}, children: [%Literal{value: nil}]}
  ],
  ...
}
```
I really like the idea, but I do not understand why you proposed to use the `%AST.If{}` with `else` for
```
<#if condition={true}>
  IF TRUE
<#else>
  ELSE
</#if>
```
Why couldn't we use the same syntax?
```elixir
%Cond{
  conditions: [
    %Cond.Condition{expression: %Expr{value: true}, children: [%Literal{value: "IF TRUE"}]},
    %Cond.Condition{expression: %Expr{value: true}, children: [%Literal{value: "ELSE"}]}
  ],
  ...
}
```

@msaraiva 
> Introduce a generic Construct AST node. The following suggestion from @lnr0626 on slack could be a starting point

I completely missed this conversation. I would have moved faster if I had seen your suggestions before :sweat_smile: 

About the `%AST.Construct{}` I don't see all the ins and outs yet but it sounds promising.  

@msaraiva & @lnr0626 How can I move forward? (1) Keep this PR as it and merge after few adjustments (2) try a new implementation for if/else/elseif as suggested by @lnr0626 (3) move forward with the `%AST.Construct{}`

Malian:
@lnr0626 I wrote a test to check that the compiler generates a proper AST for a nested `if/elseif/else` construct. I now understand why you would like to keep the AST as flat as possible :sweat_smile:

@msaraiva & @lnr0626  I introduced a warning if old `<If>` syntax is used. It results in some warnings when we run test. My first trough has been to convert `<If>` to `<#if>` in the tests but I am afraid to do that as we may miss some regressions in the future. The `<If>` construct component and the `<#if>` construct are two different implementations and, as we would like to keep both, I believe removing tests are not the solution. WDYT?

Other changes:

- Revert IfElse in favour of If with an `else` attribute. In that way, we have only one AST node and we do not break library that could use `AST.If` directly. First I removed the AST.If in favour of AST.IfElse, but I believe we should keep backward compatibility for 0.5.

lib/surface/compiler.ex
```diff
@@ -380,7 +435,13 @@ defmodule Surface.Compiler do
     end
   end
 
-  defp convert_node_to_ast(:component, {name, attributes, children, node_meta}, compile_meta) do
+  defp convert_node_to_ast(
+         :component,
+         {name, attributes, children, node_meta} = node,
+         compile_meta
+       ) do
+    maybe_warn_on_deprecated_if_notation(node, compile_meta)
```
lnr0626, line 443: we could also put this warning in `Surface.Constructs.If.transform/1` - that will always be called when `If` is used

lnr0626 (commented):
This looks good!

> in the tests but I am afraid to do that as we may miss some regressions

agreed that we should keep some tests around using `<If>`, though we may want to keep those isolated to `test/constructs/if_test.exs` (that'll make it easy to remove them once we've removed `If`). we could add another flag in the `checks` list named something like `deprecation` to conditionally log those warnings - that would allow us to keep `If` usage around and indicate that we intentionally want to ignore the deprecation warnings in specific cases.

lnr0626:
> I now understand why you would like to keep the AST as flat as possible 😅

the ast is really painful to create and assert on by hand :-(

> First I removed the AST.If in favour of AST.IfElse, but I believe we should keep backward compatibility for 0.5

i think that makes sense - and keeping it named `If` doesn't hurt my head as if expressions (nearly?) always have the ability to specify an `else`

> I really like the idea, but I do not understand why you proposed to use the `%AST.If{}` with `else` for... Why couldn't we use the same syntax?

we can - the ast structure you suggested would work (both for `if` and for `if/else`). There were a couple thoughts behind my suggestion:
- we currently use `AST.If` in both the `debug` and `if` directives - it's a bit easier to build that structure than it is to build the `AST.Cond` structure (simply because there's less to do), and represents what we want to represent a bit more directly
- when we're translating these from our ast into elixir code, it'd be nice to generate the simplest expression that correctly represents the users code. for `if` and `if/else`, that's an if expression; but for `if/elseif/.../else` that's a cond expression. We could make this distinction in `eex_engine.ex` as different patterns to match, or by inspecting the `AST.Cond`; however, i think it'll be a bit simpler to represent that distinction directly in the nodes themselves

lib/surface/compiler.ex
Malian: It is what I would have liked to do, but we do not have access to `compile_meta` in `transform/1` so it means I would be able to use `IOHelper.warn/3`. I could create a `%CompileMeta{}` struct with the `node.meta`. What do you think is the best?

Malian:
> it'd be nice to generate the simplest expression that correctly represents the users code

That finally convinced me!

About the warnings, we could also replace the test with
```elixir
    quote do
      ~H"""
      <If condition={{ true }}>...</If>
      """
    end

    output =
      capture_io(:standard_error, fn ->
        compile_surface(code)
      end)

    assert output =~ ~r"""
      the message
    """
```

lib/surface/compiler.ex
lnr0626: node.meta should have all of the needed information on it

Malian: Thanks, @lnr0626, I did see the `line` but not the `caller` :facepalm:

msaraiva (approved):
🔥🔥🔥

PR #370, opened 2021-05-05
Fix macro component compile-time dependency

msaraiva:
This fixes the issue of not recompiling dependent components when macro components change.

By simply injecting `require TheMacroComponent` doesn't work anymore. It used to work in the past. Either way, I couldn't see we injecting that anymore.

Injecting a call to a macro belonging to the macro component seems to be the only way to force a compile-time dependency between the caller component and the macro component.

This bug was reported by @miguel-s on slack.

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -467,10 +467,10 @@ defmodule Surface.Compiler do
 
       compile_dep_expr = %AST.Expr{
         value:
-          quote generated: true, line: node_meta.line do
+          quote generated: true, line: meta.line do
             require(unquote(mod)).__compile_dep__()
           end,
-        meta: %AST.Meta{}
+        meta: meta
       }
 
       {:ok,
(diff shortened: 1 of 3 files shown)


lib/surface/compiler.ex
```diff
@@ -465,9 +465,17 @@ defmodule Surface.Compiler do
          :ok <- validate_properties(mod, attributes, directives, meta) do
       expanded = mod.expand(attributes, children, meta)
 
+      compile_dep_expr = %AST.Expr{
+        value:
+          quote generated: true, line: node_meta.line do
+            require(unquote(mod)).__compile_dep__()
+          end,
+        meta: %AST.Meta{}
```
msaraiva, line 473: @lnr0626 I'm not sure if I should use `node_meta.line` and `%AST.Meta{}` here but since this call is not supposed to fail, maybe it doesn't matter :)

lnr0626: i /think/ we might want to use `meta: node_meta` instead of `meta: %AST.Meta{}`

lnr0626 (approved):
just had one comment, but looks good to me

lib/surface/compiler.ex
msaraiva: @lnr0626 ended up using the `meta` which was already a `%Surface.AST.Meta{}`. The `node_meta` was just a map with value `%{line: line}` with `line` being the line in the template, not the file.

PR #373, opened 2021-05-06
Update link documentation

philippeqd:
Updates the `Surface.Components.Link` documentation as it contained an outdated code example. The link `method` was put into the `opts`. However, this does not work. It is now replaced with the correct `method` property.


Malian:
@philippeqd Thanks for your first contribution to Surface! :heart:

PR #374, opened 2021-05-06
Warn on invalid prop/data defaults

davydog187:
Builds off of https://github.com/surface-ui/surface/pull/314

Warns when a prop's default value is not in the `:values!` list

- [ ] Documentation updates
- [ ] Surface site PR
- [ ] Support in the docs for :values!

### DIFF
--- a/test/properties_test.exs
+++ b/test/properties_test.exs
@@ -740,15 +740,15 @@ defmodule Surface.PropertiesSyncTest do
       end)
 
     assert output =~ ~S"""
-           prop `type` default value "x-large" does not exist in :values!
+           prop `type` default value `"x-large"` does not exist in `:values!`
            """
 
     assert output =~ ~S"""
-           data `data_type` default value "x-large" does not exist in :values!
+           data `data_type` default value `"x-large"` does not exist in `:values!`
            """
 
     assert output =~ ~S"""
-           prop `invalid_type` default value [] does not exist in :values!
+           prop `invalid_type` default value `[]` does not exist in `:values!`
            """
 
     refute output =~ ~S"""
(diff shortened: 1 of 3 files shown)


test/properties_test.exs
```diff
@@ -706,4 +706,61 @@ defmodule Surface.PropertiesSyncTest do
...
+    refute output =~ ~S"""
+           prop `valid_acc`
+           """
+
+    assert output =~ ~S"""
+           prop `invalid_acc1` default value [3] does not exist in :values!
+           """
+
+    assert output =~ ~S"""
+           prop `invalid_acc2` default value 3 must be a list when accumulate: true
```
davydog187, line 763: should i wrap the value in "``" like i do for the prop/data name?

msaraiva: Yeah, I think we should start using  \`  \`  whenever we have code-related stuff, just like we do in markdown. If in the future, we want to turn those boring monochromatic messages into beautiful colourful ones (like in Rust), it won't have to update any of the messages.

Maybe this is also valid for https://github.com/surface-ui/surface/pull/374/files#diff-fb7dcaf7105c49bd4b887f35e2c32958116d06d3ef6639370ed6cb3741f287cbR491 for `accumulate: true`.

msaraiva (approved):
@davydog187 looks great to me!

davydog187:
@msaraiva updated the formatting. Are you suggesting to make a change for `accumulate: true`? That's unclear to me

msaraiva (commented):
@davydog187 looks great!

Do you think we should enhance the message a bit so we can instruct the user how to fix the issue? Something like:

```
prop `type` default value `"x-large"` does not exist in `:values!`

Hint: Either choose an existing value or replace `:values!` with `:values` to skip validation.
```

davydog187:
yea, i like that idea

davydog187:
@msaraiva I added the hint, should be good to go

msaraiva:
@davydog187 thanks! ❤️

PR #376, opened 2021-05-07
Introduce <#unless>

miguel-s:
Related to #340

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -329,24 +329,11 @@ defmodule Surface.Compiler do
 
     condition = attribute_value_as_ast(attributes, "condition", default, compile_meta)
 
-    inverted_condition =
-      case condition do
-        %AST.AttributeExpr{value: value} = expr ->
-          value =
-            quote generated: true do
-              !unquote(value)
-            end
-
-          %AST.AttributeExpr{expr | value: value}
-
-        condition ->
-          condition
-      end
-
     {:ok,
      %AST.If{
-       condition: inverted_condition,
-       children: to_ast(children, compile_meta),
+       condition: condition,
+       children: [],
+       else: to_ast(children, compile_meta),
        meta: meta
      }}
   end
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -226,7 +226,7 @@ defmodule Surface.Compiler do
   defp node_type({":" <> _, _, _, _}), do: :template
   defp node_type({"slot", _, _, _}), do: :slot
 
-  # Conditionnal blocks
+  # Conditional blocks
   defp node_type({"#if", _, _, _}), do: :if_elseif_else
   defp node_type({"#elseif", _, _, _}), do: :if_elseif_else
   defp node_type({"#else", _, _, _}), do: :if_elseif_else
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -324,7 +324,7 @@ defmodule Surface.Compiler do
          compile_meta
        ) do
     meta = Helpers.to_meta(node_meta, compile_meta)
-    default = %AST.AttributeExpr{value: false, original: "", meta: node_meta}
+    default = %AST.AttributeExpr{value: false, original: "", meta: meta}
     condition = attribute_value_as_ast(attributes, "condition", default, compile_meta)
 
     {:ok,
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -324,9 +324,7 @@ defmodule Surface.Compiler do
          compile_meta
        ) do
     meta = Helpers.to_meta(node_meta, compile_meta)
-
     default = %AST.AttributeExpr{value: false, original: "", meta: node_meta}
-
     condition = attribute_value_as_ast(attributes, "condition", default, compile_meta)
 
     {:ok,
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -230,6 +230,7 @@ defmodule Surface.Compiler do
   defp node_type({"#if", _, _, _}), do: :if_elseif_else
   defp node_type({"#elseif", _, _, _}), do: :if_elseif_else
   defp node_type({"#else", _, _, _}), do: :if_elseif_else
+  defp node_type({"#unless", _, _, _}), do: :unless
 
   defp node_type({"#" <> _, _, _, _}), do: :macro_component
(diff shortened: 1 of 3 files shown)


lib/surface/compiler.ex
```diff
@@ -317,6 +318,39 @@ defmodule Surface.Compiler do
...
+
+    condition = attribute_value_as_ast(attributes, "condition", default, compile_meta)
+
+    inverted_condition =
+      case condition do
+        %AST.AttributeExpr{value: value} = expr ->
+          value =
+            quote generated: true do
+              !unquote(value)
+            end
```
miguel-s, line 338: @msaraiva @lnr0626 so this works, but I need a sanity check if this is the right way to go :)

lib/surface/compiler.ex
```diff
@@ -317,6 +318,39 @@ defmodule Surface.Compiler do
...
+        condition ->
+          condition
+      end
+
+    {:ok,
+     %AST.If{
+       condition: inverted_condition,
+       children: to_ast(children, compile_meta),
+       meta: meta
+     }}
```
Malian, line 351: @miguel-s Another way to achieve what you want is to use the new `else` attribute of the `AST.If` struct.

```suggestion
    {:ok,
     %AST.If{
       condition: condition,
       children: [],
       else: to_ast(children, compile_meta),
       meta: meta
     }}
```

miguel-s: This makes a lot of sense and would make it trivial to support `#unless..#else` if we ever want to.

miguel-s:
@Malian great suggestions, thanks!

> I am torn between keeping the AST.If structure and adding a new AST.Unless structure

My reasons for going with `AST.If`:
1. Reusing existing code to speed up implementation, reduce our cognitive overhead and reduce maintenance costs
2. Inspired by Elixir's `unless`, which uses and inverts `if`

But I'm not married to the idea :)

> I would also add tests in the if_test.exs to check the behaviour of #unless and not only the AST.

Definitely. Wasn't very happy with my solution, so wanted to check my approach with you before investing more time into this.

lib/surface/compiler.ex
```diff
@@ -317,6 +318,24 @@ defmodule Surface.Compiler do
      }}
   end
 
+  defp convert_node_to_ast(
+         :unless,
+         {_, attributes, children, node_meta},
+         compile_meta
+       ) do
+    meta = Helpers.to_meta(node_meta, compile_meta)
+    default = %AST.AttributeExpr{value: false, original: "", meta: node_meta}
```
lnr0626, line 327:
```suggestion
    default = %AST.AttributeExpr{value: false, original: "", meta: meta}
```

The ast structures should all reference an ast.meta struct for their meta prop

lnr0626:
I think diving with ast.if makes sense for the reasons @miguel-s listed. 

I'd originally thought inverting the condition was the way to go (inspired by elixir's unless), but using the else block is a clever solution.

lib/surface/compiler.ex
msaraiva: Oh dear lord! My head just short-circuited now picturing `#unless..#else` in my mind. @miguel-s you're crazy, mate!

miguel-s: `unless..else` is a thing in Elixir 😂 https://hexdocs.pm/elixir/Kernel.html#unless/2

I tried using it once...

msaraiva:
One downside of using `AST.If` is that since we're generating an `if` instead of an `unless`, any related error presented will be mentioning `if`, not `unless`, which might be confusing. For instance:

```elixir
if true do
  1
else
  2 
else
  3
end
```

will raise:

```
(ArgumentError) invalid or duplicate keys for if, only "do" and an optional "else" are permitted
```

I'm not sure if we're already validating duplicated `else` when handling `AST.If`. If we do and the message is generic (just mentioning `else`), then I guess we're good. After all, it's an edge case.

Regarding `unless...else`, I've been working with Elixir for 6 years and I never saw that in any code. However, since Elixir's `unless` have it, we should consider supporting it too. I'm not the best person to judge that as I never used `unless` in my life and my head already hurts reading `unless` alone. Reading `unless...else` just short-circuits my brain :)

Malian:
> Regarding unless...else, I've been working with Elixir for 6 years and I never saw that in any code.

:scream: I was glad to believe this did not exist!

> However, since Elixir's unless have it, we should consider supporting it too. I'm not the best person to judge that as I never used unless in my life and my head already hurts reading unless alone. Reading unless...else just short-circuits my brain :)

I suggest not allow this at the first stage. If we receive feedback from users who want to add it, we will be able to consider supporting it too.

> I'm not sure if we're already validating duplicated else when handling AST.If.
No we don't. 

The following case is not validated too:
```elixir
<#if {true}>
IF
<#else>
ELSE
<#elseif {true}>
ELSEIF
</#if>
```

miguel-s:
> I suggest not allow this at the first stage. If we receive feedback from users who want to add it, we will be able to consider supporting it too.

Agree. It might be even positive to not support it. We can always revisit this it if we get enough feedback.

> One downside of using AST.If is that since we're generating an if instead of an unless, any related error presented will be mentioning if, not unless, which might be confusing.

I've looked and not been able to find any case like this. @Malian can you think of any place this could happen?

Malian:
> I've looked and not been able to find any case like this. @Malian can you think of any place this could happen?

With the current implementation I believe we cannot arrive in that specific invalid state. In the EEXEngine we have that piece of code that cannot have more than one `if` and one `else` option.
```elixir
{:if, [generated: true], [condition, [do: if_buffer, else: else_buffer]]}
```
Nevertheless, a code that do not respect the `if`, `if/else`, `if/elseif/elseif/...`, `if/elseif/.../elseif/else` will fail at compile time with a useless message for the end-user:
```
** (KeyError) key :line not found in: %{}
```

Malian:
@miguel-s It is easy to validate both cases (see my branch https://github.com/Malian/surface/tree/validate-wrong-if-elseif-else-constructs).

That being said, I am not sure we should merge it as it increase make the compiler file more complex and the work that @lnr0626 is doing on #378 will help us to have a more maintainable code, so I suggest to wait until #378 is merge before tackling the validations.

WDYT?

msaraiva:
> I suggest to wait until #378 is merge before tackling the validations.

Agreed. 👍

miguel-s:
> I suggest to wait until #378 is merge before tackling the validations.

Sounds good to me.

@Malian @msaraiva what does it mean for this PR? Any changes needed, should I wait or ready to go?

Malian (approved):

PR #378, opened 2021-05-07
WIP: introduce node translator into parser

lnr0626:
there's still work to be done to make this actually able to replace parts of the compiler, but please add suggestions if you have any. i'll be keeping this branch up to date as I make progress on it


msaraiva (commented):
@lnr0626 this is fantastic! 

I can see already how much easier it would be to maintain the code. Just the fact that we can move each construct to separate modules respecting a common behaviour is already a big win. I'm also really curious about the performance improvement we may get after ditching the extra pass.

Awesome work! ❤️

lib/surface/compiler/ast_translator.ex
```diff
@@ -0,0 +1,281 @@
...
+      checks: state.checks,
+      node_alias: extra[:node_alias],
+      module: extra[:module]
+    }
+  end
+
+  defp raise_default_subblock_not_allowed_error!(parent, meta) do
+    message = """
+    #{parent.name} requires all content to be wrapped in one of the following \
+    sub blocks: #{Enum.join(parent.allowed_subblocks)}
```
Malian, line 210: We also be able to give a hint depending on the parent:

Case:
```
Hint: case construct should be used like
<#case {@var}>
  <#match {[first | _ ]}>
    ...
  <#match {_}>
    ...
</#case>
```

lib/surface/compiler/ast_translator.ex
```diff
@@ -0,0 +1,281 @@
...
+
+    if directive do
+      {String.to_atom(normalized_name),
+       %Surface.Directive{
+         module: directive,
+         original_name: name,
+         name: normalized_name,
+         value: parse_value(state, directive.type(), name, value, meta),
+         meta: meta
+       }}
```
Malian, line 72: Is there a reason not to create an `handle_directive` and move the condition into the parser?

lib/surface/construct.ex
```diff
@@ -0,0 +1,45 @@
...
+              body :: list(Surface.AST.t()),
+              sub_blocks :: list(SubBlock.t()),
+              meta :: Surface.AST.Meta.t()
+            ) :: Surface.AST.t()
+
+  defmacro __using__(_opts \\ []) do
+    quote do
+      @behaviour Surface.Construct
+
+      def valid_subblocks(), do: [:default]
```
Malian, line 39: As `case` and `cond` will not have `:default` as a valid subblock, I prefer to be explicit with the `template` and `slot` construct and declare the `valid_subblocks` instead of relying on a default implementation. WDYT?

lib/surface/constructs/if_construct.ex
```diff
@@ -0,0 +1,90 @@
...
+  alias Surface.Construct
+  alias Surface.IOHelper
+
+  def valid_subblocks(), do: [:default, "elseif", "else"]
+
+  def attribute_type(block, attribute, _)
+      when block in [:default, "elseif"] and attribute in [:root, "condition"],
+      do: :boolean
+
+  def attribute_type("else", name, meta) when name in [:root, "condition"] do
```
Malian, line 12: :+1:

Malian (commented):
@lnr0626 Amazing work! I let only minor comments.

lib/surface/compiler/ast_translator.ex
```diff
@@ -0,0 +1,281 @@
...
+  defguardp is_subblock(context) when context.type == Surface.Construct.SubBlock
+  defguardp is_tag(context) when context.type in [AST.Tag, AST.VoidTag]
+
+  def handle_comment(state, _comment), do: {state, :ignore}
+
+  def handle_literal(state, value) do
+    {state, %AST.Literal{value: value}}
+  end
+
+  def handle_end(_state, children) do
```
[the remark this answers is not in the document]
lnr0626: my original thought was that it might make sense to collect information while processing the tree and store that in the state, then this would allow a translator to return some additional information alongside the ast. i'm not sure that'll end up being useful, and if not i'll just remove it :-)

lib/surface/compiler/ast_translator.ex
lnr0626: that's a good idea - it might make sense to turn the `valid_subblocks` callback into a `validate_subblock` or similar to enable construct specific messages without this having to know anything specific about that construct

lib/surface/compiler/ast_translator.ex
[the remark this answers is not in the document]
Malian: Interesting idea. I am wondering if we could plug there some static tool analyser like credo that could give to the end-user some useful information about the complexity of the app, best practices, etc.

lib/surface/compiler/ast_translator.ex
lnr0626: though that would be a bit simpler, the parser doesn't quite have enough knowledge to tell if something's a directive or just an attribute

lib/surface/constructs/for_construct.ex
```diff
@@ -0,0 +1,58 @@
...
+       ...
+     <#else>
+       ...
+     </#if>
+     ```
+     """}
+  end
+
+  def attribute_type(block, attribute, _)
+      when block in [:default, "elseif"] and attribute in [:root, "condition"],
```
Malian, line 22:
```suggestion
      when block in [:default] and attribute in [:root, "condition"],
```

lnr0626:
I'll likely split this into two prs - one to add in the node translator stuff, and another to implement the AST translator to keep the work a bit better contained

msaraiva:
Sounds reasonable to me! 👍

lnr0626:
I'm going to close this and will work on the remaining pieces of it through other PRs

PR #379, opened 2021-05-07
Deprecated unquoted strings

Malian:
Related to #340 

~~My first attempt was to add handle this in the `parser` but I believe it is not the role of the parser and it. I was able to move this warning to the `TypeHandler` or directly to the `Compiler`. From my point of view, it makes more sense to handle it in the compiler, so I decided to catch it in the `attr_value` function when the `attr_value` is not a string. If it is not a string, even if the type is `:string`, it means that the value as been converted before to something else, boolean or integer.~~

Ok, this will teach me not to run all the tests before pushing...

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -672,7 +672,19 @@ defmodule Surface.Compiler do
     }
   end
 
-  defp attr_value(name, type, value, meta) do
+  defp attr_value(name, type, value, meta) when is_boolean(value) or is_integer(value) do
+    message = """
+      Passing attribute values as `#{name}=#{value}` has been deprecated and will be removed in future version.
+
+      Hint: replace `#{name}=#{value}` by `#{name}={#{value}}`
+    """
+
+    IOHelper.warn(message, meta.caller, fn _ -> meta.line end)
+
+    Surface.TypeHandler.literal_to_ast_node!(type, name, value, meta)
+  end
+
+  defp attr_value(name, type, value, meta) when is_binary(value) do
     Surface.TypeHandler.literal_to_ast_node!(type, name, value, meta)
   end
 
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -605,6 +605,7 @@ defmodule Surface.Compiler do
   end
 
   defp process_attributes(mod, [{name, value, attr_meta} | attrs], meta, acc) do
+    unquoted_string? = attr_meta[:unquoted_string?]
     name = String.to_atom(name)
     attr_meta = Helpers.to_meta(attr_meta, meta)
     {type, type_opts} = Surface.TypeHandler.attribute_type_and_opts(mod, name, attr_meta)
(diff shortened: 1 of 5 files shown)


Malian:
Mhhh these tests should not pass. We should force `warning_as_error=true` in `0.5`

Malian:
@msaraiva & @lnr0626 due to the parser we will have to raise a warning from the parser, or modify the way it handles `<div selected>`.

Because of this piece of code in the parser:
https://github.com/surface-ui/surface/blob/2a17bda331022943f1e28e593fce2a9f2ef9308b/lib/surface/compiler/parser.ex#L229-L231

There is no way to differentiate `<div selected>` from `<div selected=true>` in the compiler.

We could emit a warning from the parser, but we do not have access to a `caller` for `IOHelper.warn`. We could also modify the `translate_attr` signature to differentiate both cases.

For example, we could replace the above function with:
```elixir
  defp translate_attr({name, nil, meta}) do
    {name, {:attribute_expr, true, to_meta(meta)}, to_meta(attr_meta)}
  end
```
and keep the current implementation. It means that `<div selected>` === `<div selected={true}>` !== `<div selected=true>`. Then in the compiler we emit a warning when a `:string` type with a boolean value or an integer value are provided.

WDYT? Do you have any alternatives?

lib/surface/compiler.ex
```diff
@@ -665,7 +665,19 @@ defmodule Surface.Compiler do
     }
   end
 
-  defp attr_value(name, type, value, meta) do
+  defp attr_value(name, type, value, meta) when is_boolean(value) or is_integer(value) do
+    message = """
+      Passing attribute values as `#{name}=#{value}` has been deprecated and will be removed in future version.
+
+      Hint: replace `#{name}=#{value}` by `#{name}={#{value}}`
```
msaraiva, line 672:
```suggestion
    passing unquoted attribute values has been deprecated and will be removed in future versions.

    Hint: replace `#{name}=#{value}` with `#{name}={#{value}}`
```

msaraiva (commented):
@Malian what about passing an extra field in the attribute's metadata to differentiate them? Since the attribute's metadata is available in `attr_value`, you could check that info.

Malian:
> what about passing an extra field in the attribute's metadata to differentiate them? Since the attribute's metadata is available in attr_value, you could check that info.

@msaraiva It is a solution, but it implies that we have to modify the `AST.Meta` struct to accept extra metadata about the node. We call everywhere in the code `Helpers.to_meta` that returns an `AST.Meta`. Adding an attribute that is only used for deprecated code and that we will remove in the future seems weird.

https://github.com/surface-ui/surface/blob/f3a6d224fcbf69f34c9a91725cbb5594a6a89638/lib/surface/compiler/helpers.ex#L101-L113

We could add a field that handles all extra meta, but It is maybe an overkill addition at this stage, isn't it?

An alternative could be to catch this extra metadata one function above in the `process_attributes` but I have the feeling that it makes the code more spaghetti. As I said, we will remove this code in the future, so maybe we should not care too much about it?

https://github.com/surface-ui/surface/blob/f3a6d224fcbf69f34c9a91725cbb5594a6a89638/lib/surface/compiler.ex#L607-L611

msaraiva:
> we will remove this code in the future, so maybe we should not care too much about it?

Agreed. We shouldn't care. Those deprecations are temporary so if they work as expected, it's fine. 👍

msaraiva (approved):

Issue #343, opened 2021-04-21
Support recursive components


msaraiva:
Move all compile-time validation to `after_compile` so we can allow a component to inject itself in `render/1`.

PR #382, opened 2021-05-08
Introduce the concept of root properties

Malian:
Related to #340

### DIFF
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -59,6 +59,8 @@ defmodule Surface.API do
       # Any caller component can hold other components with slots
       Module.register_attribute(__MODULE__, :assigned_slots_by_parent, accumulate: false)
 
+      Module.register_attribute(__MODULE__, :root_prop, accumulate: false)
+
       Module.put_attribute(__MODULE__, :use_context?, false)
 
       for func <- unquote(include) do
@@ -114,6 +116,22 @@ defmodule Surface.API do
     }
 
     assigns = Module.get_attribute(caller.module, :assigns) || %{}
+
+    validate_existing_assign!(assign, assigns, caller)
+
+    if Keyword.get(opts, :root, false) do
+      validate_existing_root_prop!(assign, caller)
+      Module.put_attribute(caller.module, :root_prop, assign)
+    end
+
+    name = Keyword.get(assign.opts, :as, assign.name)
+    new_assigns = Map.put(assigns, name, assign)
+
+    Module.put_attribute(caller.module, :assigns, new_assigns)
+    Module.put_attribute(caller.module, assign.func, assign)
+  end
+
+  defp validate_existing_assign!(assign, assigns, caller) do
     name = Keyword.get(assign.opts, :as, assign.name)
     existing_assign = assigns[name]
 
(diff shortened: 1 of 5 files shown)


lib/surface/api.ex
```diff
@@ -59,6 +59,8 @@ defmodule Surface.API do
       # Any caller component can hold other components with slots
       Module.register_attribute(__MODULE__, :assigned_slots_by_parent, accumulate: false)
 
+      Module.register_attribute(__MODULE__, :root_prop, accumulate: false)
```
Malian, line 62: Not a fan of using an attribute, but I wanted to test quickly the concept.

Malian: An alternative could be to traverse all the existing assigns for every new prop. I suspect this is not what we want :sweat_smile:  We have already talked about refactoring this part of the code.

msaraiva: @Malian in order to achieve https://github.com/surface-ui/surface/issues/343, we should start moving every validation done in `put_assign!` to `__after_compile__`. Since this is a new one, you could try to do it already there. The concept is quite simple:

1. Gather whatever information you need in `put_assign!`
2. Use that information to perform the actual validation on `__after_compile__`

In this case, it would make the validation even simpler as you would just allow multiple props defined with `root: true`. Then, on `__after_compile__` you only have to filter all props with `root: true`. If the list has more than one prop, you can raise a single error indicating all related properties, something like:

```
cannot have multiple properties defined as `root: true`. The following properties were already defined as root:

`prop label` at line 4, `prop whatever` as line 6 and `prop other` at line 10.

Hint: choose a single property to be the root prop.
```

This is just the first example message that came to my mind which could probably be implemented using `Helpers.list_to_string/3`. It could definitely be simpler though. I mean, we shouldn't expect users using the root prop as crazy so any message pointing to the problem will probably be enough. 

If you prefer, you don't have to move the validation to `__after_compile__` now. You can leave it to be done in the scope of https://github.com/surface-ui/surface/issues/343. If that's the case, I believe the current implementation is fine.

msaraiva: BTW, since the message must be related to a specific line, we could raise the error on the first prop redefined as root. No need to list all of them (I don't see that happening often). This way we could use a simpler message:

```
cannot define multiple properties as `root: true`. Property `label` at line 4 was already defined as root.

Hint: choose a single property to be the root prop.
```

Malian: @msaraiva moved! I will move all the other validations in separate PR if it is ok for you?

Malian: After a few tests, it seems that it would not be so complicated to move validation to `__after_compile__`. I'll propose a PR with that changes.

msaraiva (commented):
> After a few tests, it seems that it would not be so complicated to move validation to `__after_compile__`. I'll propose a PR with that changes.

@Malian that's great news! If you want to do it in the scope of this PR, it's fine. If not, it's fine too :)

I also left a couple of minor comments.

Malian:
@msaraiva 
> If you want to do it in the scope of this PR, it's fine. If not, it's fine too :)

I'll do this in a new PR :)

msaraiva (approved):
Awesome! ❤️

PR #383, opened 2021-05-09
Add support for "phx-trigger-action" in `<Form>`

miguel-s:


dekdekbaloo:
I've been looking for this. Hope this one gets merged real soon 🙏

miguel-s:
@dekdekbaloo 👍 

In case you didn't know, until this PR is merged you can pass in the raw option like so `<Form opts={{ :"phx-trigger-action": true }}>`

Malian (approved):

msaraiva:
@miguel-s is using `<Form trigger_action={true}>` any different from using `<Form opts={"phx-trigger-action": true}>`? Or it's just about improving DX? 

My only concern to include props that don't do anything extra is to create a false expectation for the users that every `phx-*` attribute is mapped as a `prop`.

Malian:
@msaraiva It is about improving DX. If you would like to have more info about why we proposed this, you can start reading on Slack from here: https://elixir-lang.slack.com/archives/C01220C92QM/p1620475740453600

miguel-s:
@msaraiva it technically doesn't do anything additional or different.

I was aiming for self-documenting components i.e. the props should describe what the component is capable of (especially so with LV events and bindings).

On Slack and the forums I'm seeing more and more that most devs (old and new) don't really know or understand that they can pass the raw LV options to the `opts` prop.

msaraiva:
I see. Ideally, we could have all of them as props to improve DX but that's a risky move. It might be hard to keep them all in sync with LV. At least for a while until LV gets more stable in that regard.

Malian:
It is an interesting point of view. An alternative could be to wait for LV to get more stable and in the meantime, improve the doc about Form about the usage of `phx-trigger-form`.

Malian:
@miguel-s Ok for me to go with `trigger_action`.

However, I am not sure I understand what you mean with
>phx-key: to wrap up events and have nice docs with values option (prop key, :string, values: ["Escape", "Enter", ...])

Could you please elaborate?

miguel-s:
@Malian let me try again, these are the reasons why I would like to add `phx-key`:

1. We already support all the key events (`phx-keydown`, `phx-keyup`, `phx-window-keydown`, `phx-window-keyup`), but we're missing `phx-key`. As a user I would expect it to be supported, too. Especially as a new user I would be confused if it's not there or if I have to pass the raw binding to `opts`.
2. In LV it's not clear what values are accepted by `phx-key`. The documentation mentions some examples and links to 2 external sites. In Surface with the `values` option we can provide some hints for the most common uses like `Escape` and `Enter`.

Hope this is a bit clearer 🙂

miguel-s:
@msaraiva should we go ahead with this PR or are you not a fan of this approach?

msaraiva:
@miguel-s I'm fine with `trigger_action` as it will be specific to `<Form>`, however, for other stuff that applies to many other components like `phx-key`, `phx-debounce` and friends, I'd wait. I'm not convinced it's a good idea to have properties to all of them at the moment.

PR #385, opened 2021-05-10
Introduce <#for>...<#else>...</#for> 

miguel-s:
Related to #340

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -235,6 +235,7 @@ defmodule Surface.Compiler do
   # Raw
   defp node_type({"#raw", _, _, _}), do: :raw
 
+  defp node_type({"#for", _, _, _}), do: :for_else
   defp node_type({"#" <> _, _, _, _}), do: :macro_component
   defp node_type({<<first, _::binary>>, _, _, _}) when first in ?A..?Z, do: :component
   defp node_type({name, _, _, _}) when name in @void_elements, do: :void_tag
@@ -341,6 +342,57 @@ defmodule Surface.Compiler do
        else: to_ast(children, compile_meta),
        meta: meta
      }}
+  end
+
+  defp convert_node_to_ast(
+         :for_else,
+         {_name, attributes, children, node_meta},
+         compile_meta
+       ) do
+    meta = Helpers.to_meta(node_meta, compile_meta)
+    default = %AST.AttributeExpr{value: false, original: "", meta: meta}
+    generator = attribute_value_as_ast(attributes, "each", :generator, default, compile_meta)
+
+    [for_children, else_children] =
+      case children do
+        [{:default, [], default, _}, {"#else", _, _, _} = else_block] ->
+          [default, [else_block]]
+
+        [{:default, [], default, _}] ->
+          [default, []]
+
+        children ->
+          [children, []]
+      end
+
+    for_ast = %AST.For{
+      generator: generator,
+      children: to_ast(for_children, compile_meta),
+      meta: meta
+    }
+
+    if else_children == [] do
+      {:ok, for_ast}
+    else
+      [{_, _, [{_, _, _}, value]}] = generator.value
+
+      condition = %AST.AttributeExpr{
+        original: "",
+        value:
+          quote generated: true do
+            unquote(value) != []
+          end,
+        meta: compile_meta
+      }
+
+      {:ok,
+       %AST.If{
+         condition: condition,
+         children: [for_ast],
+         else: to_ast(else_children, compile_meta),
+         meta: meta
+       }}
+    end
   end
 
   defp convert_node_to_ast(
@@ -587,20 +639,20 @@ defmodule Surface.Compiler do
   defp has_attribute?(attributes, attr_name),
     do: Enum.any?(attributes, &match?({^attr_name, _, _}, &1))
 
-  defp attribute_value_as_ast(attributes, attr_name, default, meta) do
+  defp attribute_value_as_ast(attributes, attr_name, type \\ :integer, default, meta) do
     Enum.find_value(attributes, default, fn
       {^attr_name, {:attribute_expr, value, expr_meta}, _attr_meta} ->
         expr_meta = Helpers.to_meta(expr_meta, meta)
 
         %AST.AttributeExpr{
           original: value,
-          value: Surface.TypeHandler.expr_to_quoted!(value, attr_name, :integer, expr_meta),
+          value: Surface.TypeHandler.expr_to_quoted!(value, attr_name, type, expr_meta),
           meta: expr_meta
         }
 
       {^attr_name, value, attr_meta} ->
         attr_meta = Helpers.to_meta(attr_meta, meta)
-        Surface.TypeHandler.literal_to_ast_node!(:integer, attr_name, value, attr_meta)
+        Surface.TypeHandler.literal_to_ast_node!(type, attr_name, value, attr_meta)
 
       _ ->
         nil
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -374,10 +374,12 @@ defmodule Surface.Compiler do
     if else_children == [] do
       {:ok, for_ast}
     else
+      [else_ast | _] = to_ast(else_children, compile_meta)
+
       value =
         case generator.value do
           [{:<-, _, [_, value]}] -> value
-          _ -> raise_complex_generator(meta)
+          _ -> raise_complex_generator(else_ast.meta)
         end
 
       condition = %AST.AttributeExpr{
@@ -393,7 +395,7 @@ defmodule Surface.Compiler do
        %AST.If{
          condition: condition,
          children: [for_ast],
-         else: to_ast(else_children, compile_meta),
+         else: [else_ast],
          meta: meta
        }}
     end
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -1084,16 +1084,15 @@ defmodule Surface.Compiler do
 
   defp raise_complex_generator(meta) do
     message = """
-    using <#for>..<#else> only supports single generators without filters.
+    using `<#else>` is only supported when the expression in `<#for>` has a single generator and no filters.
 
     Example:
-      ```
+
       <#for each={i <- [1, 2, 3]}>
-      ...
+        ...
       <#else>
-      ...
+        ...
       </#for>
-      ```
     """
 
     IOHelper.compile_error(message, meta.file, meta.line)
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -232,10 +232,12 @@ defmodule Surface.Compiler do
   defp node_type({"#else", _, _, _}), do: :if_elseif_else
   defp node_type({"#unless", _, _, _}), do: :unless
 
+  # For
+  defp node_type({"#for", _, _, _}), do: :for_else
+
   # Raw
   defp node_type({"#raw", _, _, _}), do: :raw
 
-  defp node_type({"#for", _, _, _}), do: :for_else
   defp node_type({"#" <> _, _, _, _}), do: :macro_component
   defp node_type({<<first, _::binary>>, _, _, _}) when first in ?A..?Z, do: :component
   defp node_type({name, _, _, _}) when name in @void_elements, do: :void_tag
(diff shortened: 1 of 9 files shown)


lib/surface/compiler.ex
```diff
@@ -324,6 +325,72 @@ defmodule Surface.Compiler do
...
+    else
+      [{_, _, [{_, _, _}, value]}] = generator.value
+
+      condition = %AST.AttributeExpr{
+        original: "",
+        value:
+          quote generated: true do
+            unquote(value) != []
+          end,
+        meta: compile_meta
```
miguel-s, line 381: So this works but I'm almost ashamed of having written it 😅 

@Malian @lnr0626 @msaraiva I need smarter people than me to enlighten me

Malian: @miguel-s I would introduce a new `else` attribute in the `AST.For` instead of wrapping the `For` in the `If`. The main reason is to have an AST as close as possible of the user code.

We could handle the two cases in the `EExEngine`.

Another solution is to wait that for #378 :sweat_smile:

msaraiva: @miguel-s I don't think that will work when using filters, like:

```
<#for {i <- items, items in [...]}>
  ...
</#for>
```

The condition of the `If` must test against the result of the `for`. We need to generate something like:

```elixir
for {i <- items, items in [...]} do
  ...
end |> case do
  [] ->
    # The else block
  result ->
    result
end
```

However, I'm not sure if creating this variable would mess with diff tracking inside the `for` block.

lib/surface/compiler.ex
```diff
@@ -343,6 +346,62 @@ defmodule Surface.Compiler do
...
+        value:
+          quote do
+            unquote({:for, [generated: true], generator_expr})
+            |> case do
+              [] -> false
+              _ -> true
+            end
+          end,
+        meta: compile_meta
+      }
```
miguel-s, line 393: > The condition of the If must test against the result of the for. We need to generate something like:
> 
> ```elixir
> for {i <- items, i in [...]} do
>   ...
> end |> case do
>   [] ->
>     # The else block
>   result ->
>     result
> end
> ```

@msaraiva new attempt, a bit closer to what you suggest and handles the generator with filters.

Still not too happy with the code, though.

lib/surface/compiler/eex_engine.ex
```diff
@@ -102,6 +102,45 @@ defmodule Surface.Compiler.EExEngine do
...
+    for_expr = {:for, [generated: true], generator_expr}
+
+    case_expr =
+      quote generated: true do
+        unquote(for_expr)
+        |> case do
+          [] -> unquote(else_buffer)
+          result -> result
+        end
+      end
```
miguel-s, line 138: @lnr0626 @Malian `for_else` works fine when passing regular html as children, but breaks when passing other components with error (both as child of the `for` or the `else`):
> ** (ArgumentError) cannot convert component Surface.Constructs.ForTest.SomeComponent with id nil to HTML.

It happens in this highlighted piece of code, any hints?

msaraiva: @miguel-s that's bad news. Does the `<#for>` works as expected with components if you don't use `<#else>`? If so, that means we can't use the pipe :(

Not sure how we can do this without creating a variable, which will mess with change tracking. Maybe we just can't. I'll try to come up with something but there's a big chance we'll have to let the `for...else` go.

msaraiva: Another solution would be to allow `for...else` only when it's using a single generator and no filters. Although it's not an ideal solution, using multiple generators and filters in templates is quite rare so maybe it's worth it. 

From the implementation PoV, all we'd have to do is to generate two pieces of code. For instance, the following code:

```
<#for {i <- whatever}>
  CONTENT
<#else>
  ELSE CONTENT
</#for>
```
would be equivalent to:

```
<#for {i <- whatever}>
  CONTENT
</#for>
<#if {whatever != []}>
  ELSE CONTENT
</#if>
```

The funny thing is that the above code looks exactly how most of the developers would handle this situation when they want to show a fallback-content if the list is empty.

This solution comes with a tiny performance issue if the user decides to use a more expensive expression. For instance:

```
<#for {i <- Enum.map(@items, &some_expensive_calculation/1)}>
  ...
<#else>
  ...
</#for>
```

In that case, `Enum.map/2` would always run twice for each rendering.

We could also try to generate something like:

``` 
case Enum.map(@items, &some_expensive_calculation/1)} do
  result ->
    for {i <- result}>
      ...
    end
end
```

But I'm not sure it would compromise change tracking. I'll try to create some tests so we can evaluate that. Unless, of course, we just decide to remove `for...else` from our list :)

My opinion is that, since we know when there's an `<#else>` and since we can pattern match against the expression to make sure we only accept a single generator so we can raise a nice error message when the expression does not conform to that, it seems to be worth it to give it a try.

Any other ideas?

msaraiva: @miguel-s I played around with `for` using some vanilla EEX and I couldn't see any difference regarding change tracking when wrapping it in a `case`. So, in case we decide to keep `for...else` restricting its use only when passing a single generator without filters, the following should work just fine:

```
<#for {i <- whatever}>
  CONTENT
<#else>
  ELSE CONTENT
</#for>
```

becomes:

```
case whatever do
  [] ->
    ELSE CONTENT
  result ->
    for {i <- result}>
      CONTENT
    end
end
```

lib/surface/compiler.ex
```diff
@@ -232,6 +232,9 @@ defmodule Surface.Compiler do
   defp node_type({"#else", _, _, _}), do: :if_elseif_else
   defp node_type({"#unless", _, _, _}), do: :unless
 
+  # For
+  defp node_type({"#for", _, _, _}), do: :for_else
+
```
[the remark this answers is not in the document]
Malian: @miguel-s As the `else` sub block has the same goal and behaviour for `if/elseif/else` and `for/else`I would go with
```elixir
defp node_type({"#else", _, _, _}), do: :else
```
and
```elixir
defp convert_node_to_ast(:else, {_, _attributes, children, node_meta}, compile_meta) do

  {:ok,
   %AST.Container{
     children: to_ast(children, compile_meta),
     meta: meta,
     directives: []
   }}
end
```

test/constructs/for_test.exs
```diff
@@ -66,4 +85,229 @@ defmodule Surface.Constructs.ForTest do
...
+        compile_surface(code)
+      end)
+    end
+
+    test "compile error message contains the correct line" do
+      code =
+        quote do
+          ~H"""
+          <#for each={fruit <- ["apples", "bananas", "oranges"]}>
+            <ListProp prop="some string" />
```
Malian, line 168:
```suggestion
            <ListProp
              prop="some string" />
```
This way we can check that the error occurs on the correct line

lib/surface/compiler/eex_engine.ex
Malian: What a bad news :cry: 

> Although it's not an ideal solution, using multiple generators and filters in templates is quite rare so maybe it's worth it.

I would say that using multiple generators, filters __and else__ in templates is quite rare. Even with the simple `for/else` use case I believe we create a lot of value for developers.

miguel-s: @msaraiva @Malian I got it to work with the compiler version (nesting `AST.For` inside `AST.If`), but have not been able to get it to work in the eex_engine version.

These are the issues I'm running into:

#### Attempt 1

```elixir
[{:<-, _, [{_, _, _}, value]}] = generator # `value` is the generator e.g. `["apples", "bananas", "oranges"]`

if_expr =
  quote generated: true do
    if unquote(value) == [] do
      unquote(else_buffer)
    else
      unquote(for_expr)
    end
  end
```

Everything inside `if` runs fine, but `else` still raises:
> ** (ArgumentError) cannot convert component Surface.Constructs.ForTest.SomeComponent with id nil to HTML.

#### Attempt 2

```elixir
[{:<-, _, [{var, _, _}, value]}] = generator # `var` is the name of the defined variable e.g. `:fruit` ; `value` is the generator e.g. `["apples", "bananas", "oranges"]`

if_expr =
  quote generated: true do
    if unquote(value) == [] do
      unquote(else_buffer)
    else
      for unquote(var) <- unquote(value) do
        unquote(for_buffer)
      end
    end
  end
```

This raises:
> ** (CompileError) test/constructs/for_test.exs:190: undefined function fruit/0

The `fruit` variable comes from the generator in the test `<#for each={fruit <- ["apples", "bananas", "oranges"]}>`

It feels like I'm getting very close, but I don't know how to solve this last error 😢

Malian: I believe it is
```elixir
[{:<-, _, [{_, _, _} = var, value]}] = generator # `var` is the name of the defined variable e.g. `:fruit` ; `value` is the generator e.g. `["apples", "bananas", "oranges"]`

if_expr =
  quote generated: true do
    if unquote(value) == [] do
      unquote(else_buffer)
    else
      for unquote(var) <- unquote(value) do
        unquote(for_buffer)
      end
    end
  end
```
but from my point of view it is exactly the same as
```elixir
[{:<-, _, [{_, _, _}, value]}] = generator # `value` is the generator e.g. `["apples", "bananas", "oranges"]`

if_expr =
  quote generated: true do
    if unquote(value) == [] do
      unquote(else_buffer)
    else
      unquote(for_expr)
    end
  end
```
or
```elixir
    case_expr =
      quote generated: true do
        case unquote(value) do
          [] -> unquote(else_buffer)
          result -> unquote(for_expr)
        end
      end
```
I am not expert enough with macros to go further.

miguel-s: @Malian you are right, your solution (I must be blind to not have seen it) fixed the CompileError, but we're back to:

> ** (ArgumentError) cannot convert component Surface.Constructs.ForTest.SomeComponent with id nil to HTML.

msaraiva: > I got it to work with the compiler version (nesting `AST.For` inside `AST.If`), but have not been able to get it to work in the eex_engine version.

@miguel-s implementing that in the eex_engine will be much harder as you can't generate the whole quoted expression like you'd normally do when creating a macro. You'll have to emit each piece of code according to what EEX expects. So I'd just do that at the compiler level using `AST.If` + `AST.For` as they should be properly handled by the engine, emitting the correct bits calling the underlying engine's `handle_*` callbacks. 

I'm glad you tried to use it with components and found that issue. We need to make sure all constructs follow that same rule, otherwise, they would work nice with HTML/text content but fail when wrapping a live component. We should add a test for that scenario to every constructwe have.

miguel-s: Pushed the compiler version and added some tests to make sure `#if` accepts components

lib/surface/compiler.ex
```diff
@@ -1026,6 +1082,23 @@ defmodule Surface.Compiler do
...
+    using <#for>..<#else> only supports single generators without filters.
+
+    Example:
+      ```
+      <#for each={i <- [1, 2, 3]}>
+      ...
+      <#else>
+      ...
+      </#for>
+      ```
```
msaraiva, line 1096: Assuming the error will be in the `<#else>` line, we could have:

```suggestion
    using `<#else>` is only supported when the expression in `<#for>` has a single generator and no filters.

    Example:

      <#for each={i <- [1, 2, 3]}>
        ...
      <#else>
        ...
      </#for>
```

However, it's hard to know if the user will prefer to keep the `else` changing the `for` expression or keep the generators/filters getting rid of the `else`. The user might not even know the concept of generators and filters. 

I'm not sure we're able to provide a better message than that.

test/constructs/for_test.exs
```diff
@@ -55,15 +74,243 @@ defmodule Surface.Constructs.ForTest do
...
+          <#for each={x <- @list1, y <- @list2, x in @range, y in @range}>
+          <span>x: {x}, y: {y}</span>
+          <#else>
+          <span>The else content</span>
+          <span>with multiple tags</span>
+          </#for>
+          """
+        end
+
+      message = ~r/using <#for>..<#else> only supports single generators without filters./
```
msaraiva, line 271: We should include the `file:line` part in the regex to make sure the error is raised at the expected line.

msaraiva (commented):
@miguel-s fantastic work! ❤️

lib/surface/compiler.ex
miguel-s: > Assuming the error will be at the <#else> line

It was at the `<#for>` line, but changed it and added the line number to the test.

You write good error messages 👏

msaraiva (approved):

Malian (approved):

Issue #334, opened 2021-04-15
Recursive Component doesn't render properly


Menkir:
Hey guys,

I've wrote a Component and want it to build  the entries in a recursive function. I collapse/expand some table entries by some click event. My problem is that the state changes, but the template is not re-rendered. Another strange thing is when i add some `IO.inspect` call within the recursive template it works correctly. 

I've created a Table Component like this: 
```elixir
defmodule Bug1Web.Surface.Table do
  use Surface.Component

  defmodule __MODULE__.Row do
    use Surface.Component

    defmodule __MODULE__.Data do
      use Surface.Component
      slot default
      def render(assigns) do
        ~H"""
        <td>
          <slot/>
        </td>
        """
      end

    end

    slot default
    def render(assigns) do
      ~H"""
      <tr>
        <slot/>
      </tr>
      """
    end
  end

  prop columns, :list, default: []
  slot default
  def render(assigns) do
    ~H"""
    <table>
      <thead>
        <tr>
          <th :for={{ column <- @columns }}>
            {{ column }}
          </th>
        </tr>
      </thead>
      <tbody>
        <slot/>
      </tbody>
    </table>
    """
  end

end

```

Then i rende the table in my live_view:
```html
<Table columns={{ [:col1, :col2, :col3] }}>
{{ render assigns, @data, 0 }}
</Table>
```

the `render/3` function is defined in the live_view:
```elixir
defmodule Bug1Web.PageLive do
  use Surface.LiveView

  alias Bug1Web.Surface.Table
  alias Table.Row
  alias Row.Data


  data data, :list, default: [
    %{
      visible: false,
      data: [1,2,3]
    },
    %{
      visible: false,
      data: [4,5,6]
    },
    %{
      visible: false,
      data: [7,8,9]
    },


  ]

  def render(_, [], _), do: nil
  def render(assigns, [%{visible: visible, data: data}|rest], index) do
    ~H"""
    <Row>
      <Data>
        <button :on-click="click" phx-value-index={{ index }}>click</button>
      </Data>
    </Row>
    <Row :if={{ visible }}>
      <Data :for={{ record <- data }}>{{ record }}</Data>
    </Row>

    {{ render assigns, rest, index + 1 }}
    """
  end

  def handle_event("click", %{"index" => index_str}, socket) do
    index = String.to_integer(index_str)
    new_socket = socket |> update(:data, & List.update_at(&1, index,
    fn entry ->
[...]

PR #386, opened 2021-05-11
Move validations to after compile

Malian:
Related to #334

### DIFF
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -68,7 +68,6 @@ defmodule Surface.API do
   end
 
   defmacro __before_compile__(env) do
-    validate_assigns!(env)
     generate_docs(env)
 
     [
@@ -80,6 +79,7 @@ defmodule Surface.API do
   end
 
   def __after_compile__(env, _) do
+    validate_assigns!(env)
     validate_duplicated_assigns!(env)
     validate_slot_props_bindings!(env)
     validate_duplicate_root_props!(env)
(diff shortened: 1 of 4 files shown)


Malian:
@msaraiva @lnr0626 @miguel-s The goal of this PR is to move validations from `put_assign!` to `__after_compile__`, not to improve the validations themselves. So it means, they are still rooms for improvements like:

- creating an `Surface.Assign` struct.
- Pass `assign` to `validate_opts` and `validate_opt` instead of passing 99% of the attributes.
- reduce the number of pass through the `assigns`
- introduce `__assigns__()` to be consistent with `__props__` and company
- ...

Before going further I would like to have your opinion/advice especially on the following points:

- One test, still not pass :( As a said on Slack, it is due to the `opts` value that is not a keyword list. When we generate the doc, in `__before_compile__` we haven't performed the validations yet and so the documentation generator fails as it assumes that the `opts` is a keyword list. I have no idea how to solve this as we cannot generate the doc in `__after_compile__` and I do not completely understand the impact of keeping some validation before `__after_compile__`.
- I tried to understand the issue with recursive components, #334 and https://elixirforum.com/t/recursive-surface-component/39019, but I am not able to write a minimum example that reproduces the error. Could you help me with this too?

msaraiva:
@Malian having validation before `__after_compile__` is only a problem if the validation assumes a child already exists, i.e. whenever the module being validated needs to access a child's public interface. e.g. `child.__props__()`, `mod.module_info()`, etc. When that happens, since the parent and the child module are the same, the module can't be compiled as it must wait for itself to be compiled to move on, which can never happen. 

Most validations in that situation are probably in the `Surface.Compiler` module, not in `put_assign!`. I believe only `validate_slot_props_bindings!` has that requirement, however, moving all validations to `__after_compile__` ensures any current or future validation can use the public interface of the module. This makes the API much more consistent as we can implement it assuming the module already exists, avoiding a bunch of conditionals like this: https://github.com/surface-ui/surface/blob/197207f2f3e77af8ad9fa73049a54dd69006952b/lib/surface/api.ex#L182-L193

Generating the docs breaks this as it must be done before compilation. So we have two options:

1. Remove the docs generation
2. Leave the minimum necessary to implement it.

If the minimum for `#2` ends up being a lot, I'd say we should consider removing it.

Malian:
I finally decided to move the validation to `validate_opts_ast` like what we do with the `validate_name_ast`. To be consistent, I added a `validate_type_ast` function.

msaraiva (approved):
❤️

Malian:
Related to #343

PR #387, opened 2021-05-11
Support dynamic attributes shorthand notation

msaraiva:
This is minimal implementation to accept `{...@attrs}` as a shorthand notation to `:attrs` and `:props`. As you can see, moving the implementation to the parser level makes everything extremely simple as it just converts the `:root` node into the respective `:attrs` and `:props` directives. However, going in this direction assumes we should keep the directives and consider the new syntax as a shorthand notation, not the only notation. I'm not sure yet this is a good idea or not.

I also implemented the root prop for `<#if>`, which was also extremely easy, however, as I said on Slack, having two different ways to handle root props (one for components, one for constructs) feels a bit odd. It would be nice to be able to define the root prop for the constructs, the same way we do for components using the same API. That would also avoid duplication of common validation. For instance, required root prop:

```elixir

demodule Surface.Constructs.If do
  use Surface.Construct

  prop condition, :boolean, root: true, required: true

  ...
end
```

I believe the current implementation of `<#if>` does not validate that yet and I wonder if we should do it already or try to come up with a more unified API for constructs before we move on. The old `<If>` component would be able to do that just fine so it seems we took a step back in that regard.

@lnr0626 I'm not sure if it would be better to wait for #378 to be finished to move on with this PR. Would the way it's been implemented change a lot with the new approach? Would it also in any way help with the concerns I just presented?

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -299,7 +299,10 @@ defmodule Surface.Compiler do
        ) do
     meta = Helpers.to_meta(node_meta, compile_meta)
     default = %AST.AttributeExpr{value: false, original: "", meta: node_meta}
-    condition = attribute_value_as_ast(attributes, "condition", default, compile_meta)
+
+    condition =
+      attribute_value_as_ast(attributes, :root, nil, compile_meta) ||
+        attribute_value_as_ast(attributes, "condition", default, compile_meta)
 
     [if_children, else_children] =
       case children do
--- a/lib/surface/compiler/parser.ex
+++ b/lib/surface/compiler/parser.ex
@@ -220,7 +220,24 @@ defmodule Surface.Compiler.Parser do
     end
   end
 
+  defp translate_attr({:root, {:expr, "..." <> value, expr_meta}, _attr_meta}) do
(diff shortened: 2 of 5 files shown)


msaraiva:
@lnr0626 so in your opinion, would be better to put this on hold until we have #378 merged?

> I really like your idea of defining props in the same way for constructs/components.

I don't think we have to do that now in the context of `v0.5`. We can choose this simpler implementation forwarding the logic to the existing `:attrs`/`:props`  directives and having the tailor-made implementation for the constructs. After `v0.5` is out, we can then focus on improving the API, maybe taking into account one of the other tagged expressions like the `{=@class}`. This way we'll have a better picture of the best way to do it.

Trying to do everything in #378 might be too much ;)

lnr0626:
> Trying to do everything in #378 might be too much ;)

Agreed 😁

> so in your opinion, would be better to put this on hold until we have #378 merged?

I think that makes sense, I also think it'll make sense to implement this with the other tagged expressions in mind like you mentioned

msaraiva:
Replaced by #416.

PR #389, opened 2021-05-16
Improve warning with duplicated html attributes

### LINKED ISSUE #381: Confusing warning message for duplicated attribute if node is not a component
From version 4.0, when there are multiple attributes found on components a warning message appears to warn the developer.

Example:
```
warning: The prop `height` has been passed multiple times. Considering only the last value.
Hint: Either remove all redundant definitions or set option `accumulate` to `true`:
```

However, the current implementation of the parser detects the duplicated attributes on non-component nodes too.
Although I found the warning for non-component node is useful, the message is rather confusing because it says about the attribute/property for a component based node.

I think there are two options to fix this issue:
- Fix the parser so it will detect only for component node
- Change the message to be more proper

Malian:
FIxes #381

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -705,39 +705,56 @@ defmodule Surface.Compiler do
     attr_meta = Helpers.to_meta(attr_meta, meta)
     {type, type_opts} = Surface.TypeHandler.attribute_type_and_opts(mod, name, attr_meta)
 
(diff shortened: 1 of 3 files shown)


Malian:
Unfortunately, I was not able to add the `tag` name to the warning message as this information is not present in the `meta` :(

msaraiva (approved):
❤️

PR #391, opened 2021-05-18
Treat embedded Elixir like a Comment

chadfennell:
In the process of converting LiveView components and views into Surface, developers can run into somewhat confusing parsing errors if they fail to convert their Embedded Elixir `<%= foo = :bar %>` into the Surface Elixir interpolation syntax `{{ foo = :bar }}`.

I bumped into this a couple of times and the "missing closing tag foo" errors had me initially scratching my head until I realized that the parser died due to some embedded Elixir that I had not converted to `{{ }}`.  My first thought was to raise a more descriptive error but then realized that just printing the code to the screen might be better - it should be fairly obvious at that point to the developer what to change.

In any case, thanks for this library, love it!

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -235,6 +235,7 @@ defmodule Surface.Compiler do
   defp node_type({_, _, _, _}), do: :tag
   defp node_type({:interpolation, _, _}), do: :interpolation
   defp node_type({:comment, _}), do: :comment
+  defp node_type({:embedded_elixir, _}), do: :embedded_elixir
   defp node_type(_), do: :text
 
   defp process_directives(%{directives: directives} = node) do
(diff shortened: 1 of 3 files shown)


Malian:
@chadfennell I already have bumped into this issue a couple of times. I like the idea to not raise an error, or at least we need a better error message.

We are working on the future of Surface on the surface-next branch. In that branch, we removed the nibleparsec parser, and we replaced it with a new tokenizer. So it means, even if we accept your change, we will have to reimplement it in the next version. I suggest to directly add your proposition in the next parser/tokenizer. @msaraiva could we easily catch "eex" expressions with the new tokenizer? If so, it could be a good improvement for users who switch from Phoenix to Surface. We will be able to give them proper message/warnings.

@chadfennell I am not sure treating `<%= %>` as plain text is the better idea, but I am not against it. I would at least emit a warning + your proposition.

chadfennell:
Closing in favor of a new PR against the upcoming version - I'm going to raise a descriptive error in the new PR.

PR #392, opened 2021-05-19
New notation for constructs

msaraiva:
These are the minimum changes necessary to apply the new syntax without updating the tokenizer and keeping the same structure of the nodes, except for the addition of the `:block` atom to differentiate blocks from tags. It even keeps the representation of the construct expression as a `:root` prop. However, we should change this to a more appropriate format after #378 gets merged.

@lnr0626, @Malian and @miguel-s there are still a couple of things to decide:

1. Are we keeping `<#slot>`, `<#template>` and `<#raw>` or should we go back to `<Slot>`, `<Template>` and `<#Raw>`?
2. Should we use `{/...}` or `{#end}`. The current version is using `{/...}` but it's extremely easy to change. 

Here are some examples using both notations:

### Using `{/...}`

image

### Using `{#end}`

image

Using `{#end}` somehow looks better to me. I'm not sure why :)

### DIFF
--- a/lib/surface/compiler/parser.ex
+++ b/lib/surface/compiler/parser.ex
@@ -69,8 +69,8 @@ defmodule Surface.Compiler.Parser do
     handle_token(rest, buffers, state)
   end
 
-  defp handle_token([{:comment, comment} | rest], buffers, state) do
-    buffers = push_node_to_current_buffer({:comment, comment}, buffers)
+  defp handle_token([{:comment, comment, meta} | rest], buffers, state) do
+    buffers = push_node_to_current_buffer({:comment, comment, meta}, buffers)
     handle_token(rest, buffers, state)
   end
 
--- a/lib/surface/compiler/parser.ex
+++ b/lib/surface/compiler/parser.ex
@@ -78,7 +78,11 @@ defmodule Surface.Compiler.Parser do
   # should be moved to the tokenizer or not. We could also hold the expression
   # directly instead of using the :root attribute.
   for block <- @blocks do
-    defp handle_token([{:interpolation, "#" <> unquote(block) <> " " <> expr, meta} | rest], buffers, state) do
+    defp handle_token(
+           [{:interpolation, "#" <> unquote(block) <> " " <> expr, meta} | rest],
+           buffers,
+           state
+         ) do
       expr_meta = %{meta | column: meta.column + String.length(unquote(block)) + 2}
       attrs = [{:root, {:expr, expr, expr_meta}, expr_meta}]
       handle_token([{:block_open, unquote(block), attrs, meta} | rest], buffers, state)
@@ -169,7 +173,11 @@ defmodule Surface.Compiler.Parser do
   end
 
   # If there's a previous sub-block defined. Close it.
-  defp close_sub_block(_token, buffers, %{tags: [{:block_open, name, attrs, meta} | tags]} = state)
+  defp close_sub_block(
+         _token,
+         buffers,
+         %{tags: [{:block_open, name, attrs, meta} | tags]} = state
+       )
        when name in @sub_blocks do
     # pop the current buffer and use it as children for the sub-block node
     [buffer | buffers] = buffers
@@ -284,11 +292,17 @@ defmodule Surface.Compiler.Parser do
     %{state | tags: [token | state.tags]}
   end
 
-  defp pop_matching_tag(%{tags: [{:tag_open, tag_name, _, _} = tag | tags]} = state, {:tag_close, tag_name, _}) do
+  defp pop_matching_tag(
+         %{tags: [{:tag_open, tag_name, _, _} = tag | tags]} = state,
+         {:tag_close, tag_name, _}
+       ) do
     {tag, %{state | tags: tags}}
   end
 
-  defp pop_matching_tag(%{tags: [{:block_open, name, _, _} = tag | tags]} = state, {:block_close, name, _}) do
+  defp pop_matching_tag(
+         %{tags: [{:block_open, name, _, _} = tag | tags]} = state,
+         {:block_close, name, _}
+       ) do
     {tag, %{state | tags: tags}}
   end
 
--- a/lib/surface/compiler/parser.ex
+++ b/lib/surface/compiler/parser.ex
@@ -25,10 +25,7 @@ defmodule Surface.Compiler.Parser do
     "if",
     "unless",
     "for",
-    "case",
-    "else",
-    "elseif",
-    "match"
+    "case"
   ]
 
   @sub_blocks [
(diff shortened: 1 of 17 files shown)


lnr0626:
Though `end` feels a bit more natural (likely because of elixir's syntax), I actually prefer the `{/...}` variant - almost purely because it'll make it easier for me to figure out which closing thing matches up with which opening thing, ie:

```
{#for i <- @items}
  {#if mod(i, 2) == 1}
    even!
  {#else}
    odd!
  {/if}
{/for}
```
vs:
```
{#for i <- @items}
  {#if mod(i, 2) == 1}
    even!
  {#else}
    odd!
  {#end}
{#end}
```

It'll also allow us to indicate which construct is missing it's closing tag:

```
{#for i <- @items}
  {#if mod(i, 2) == 1}
    even!
  {#else}
    odd!
{/for}
```
we'd be able to say the `#if` is missing it's closing, and would be able have some better suggestions as to where the missing closing tag could go - in this case the `#else` block could be associated with either the `#for` or the `#if`, depending on what the user intended

compared to:

```
{#for i <- @items}
  {#if mod(i, 2) == 1}
    even!
  {#else}
    odd!
  {#end}
```

we'd assume the missing `#end` would be for the `#for` construct

miguel-s:
> Though end feels a bit more natural (likely because of elixir's syntax), I actually prefer the {/...} variant - almost purely because it'll make it easier for me to figure out which closing thing matches up with which opening thing

+1 on using `{/...}`! Html makes me blind and I alway miss the closing tags, the easier and clearer we make it the better for me 😄

msaraiva:
Another advantage of using `{/...}` is that it provides a more consistent way to define self-closing constructs (in case we ever need it). For instance, suppose we would like to convert the `<#slot>` tag into a construct in the future, we could use:

```
 {#slot :footer, item: item /}
```

So we have this implicit symmetry between `<tag>`|`</tag>`|`<tag/>` and `{#contruct}`|`{/contruct}`|`{contruct/}`. 

We could even make that symmetry more explicit when defining our private comment notation. We could have:

```
<!--
This comment hits the browser
--}

{!--
This comment does not hit the browser
--}
```

Malian:
I also vote for `{\...}` for the same reasons as my companions.

About `<#slot>`,  `<#template>` and `<#raw>`, could we convert them to `construct`? If so, should we?

- `<#template>` and `<#raw>` will always have an end tag, so I suppose it is easy to treat them like `case` and `for`.
- `<#template slot="header">` will be replaced with `<:header>` anyway
- What about a self closing construct ? `<#slot name="header" />`. I could imagine a `{#deprecated message /}` construct into sface code
- I am fine with the syntax: `{#slot header}...{/slot}` or `{#slot name="header"}...{/slot}` or `{#slot /}`

If we cannot, or if we would not, I would like to use the new syntax `<#slot>` as I think `slot` and `template` are not "normal" component. They are special component like the `#raw` macro component.

msaraiva:
Only one thing left to do at the tokenizer level which is to handle tagged expressions. Example:

```
{=@class} {...@attrs}
```

should be tokenized as:

```
[
  {:expr, "@class", %{marker: "=", line: 1, ...}},
  {:expr, "@attrs", %{marker: "...", line: 1, ...}}
]
```

 I'll try to finish it up tomorrow morning so it could be ready for review.

@lnr0626 do you think we should merge `surface-next` into `master` after this change or do you think it's better to wait for #378?

msaraiva:
Guys, I decided to go a bit further in the scope of this PR to do the following:

1. Handle `{:tagged_expr, ...}` in attributes so we implement `{...@attrs}` and `{=@class}`
2. Normalize all expressions/interpolation to use either `{:expr, ...}` or `{:tagged_expr}` (no need to have a separate `{:interpolation, ...}`)

In case this PR is holding anyone on any related issue/PR, please let me know so we can consider merging it before.

msaraiva:
Ok. I think this is ready for a first review. After this gets merged, we'll be able to implement `{...@attrs}` and `{=@class}` using the new `{:tagged_expr, marker, expr, meta}` node.

I also improved the error messages for unclosed curly braces on all types of expressions. Some of them would just say:

```
expected closing `}`
```

at the end of the file, without giving any clue about which and where the related opening brace is.

There are other more specific error messages we'll need to add at the parser level but we can do this in another PR. One example is to validate which tagged expression are accepted in which context (body, root expression or attribute value):

  * `{...@attrs}` and `{=...}` -> only as root attribute of a comonent/tag
  * `{~...}` and friends -> only in the body or as attribute value

lib/surface/compiler/parser.ex
```diff
@@ -102,32 +96,94 @@ defmodule Surface.Compiler.Parser do
...
-    node = {name, transtate_attrs(attrs), Enum.reverse(buffer), to_meta(meta)}
+    node = {:block, name, transtate_attrs(attrs), Enum.reverse(buffer), to_meta(meta)}
     buffers = push_node_to_current_buffer(node, buffers)
     handle_token(rest, buffers, state)
   end
 
-  # IF there's a previous sub-block defined. Close it.
-  defp close_sub_block(_token, buffers, %{tags: [{:tag_open, name, attrs, meta} | tags]} = state)
+  defp handle_token([{:block_close, name, meta} | _], _buffers, _state) do
+    blocks = Helpers.list_to_string("block is", "blocks are", @blocks ++ @sub_blocks)
```
Malian, line 173:
```suggestion
    blocks = Helpers.list_to_string("block is", "blocks are", @blocks)
```
I would remove `@sub_blocks` as the user can't close a sub block with `{/sub_block_name}`, he must use the parent block name to close the whole block.

We can go further and if the user uses a subblock name to "close" a block we could give him a proper hint. For example, with the `else` sub block:
> You can only close a sub block by closing the whole block. Did you mean `{/if}` or `{/for}`?

test/parser_test.exs
```diff
@@ -774,189 +751,226 @@ defmodule Surface.Compiler.ParserTest do
...
+      {#if}
+        1
+      {/iff}
+      """
+
+      exception = assert_raise ParseError, fn -> parse!(code) end
+
+      message = """
+      unknown `{/iff}` block. Available blocks are \
+      "if", "unless", "for", "case", "else", "elseif" and "match"\
```
Malian, line 970: Propose `case`, `else`, `elseif` and `match` may mislead the user as these constructs cannot be directly closed with their equivalent: `{/case}`, `{/else}`, `{/elseif}` and `{/match}` I suggest to remove them.

Malian:
@msaraiva We should also update the `converter` to transform:

- `<If condition={{ my_condition }}>` to `{#if my_condition}`.
- `</If>` to `{/if}`
- `<For each={{ item <- [1, 2, 3] }}>` to `{#for item <- [1, 2, 3]}`
- `</For>` to `{/for}`

lib/surface/compiler/parser.ex
msaraiva: Good catch!

PR #393, opened 2021-05-20
Support Descriptive EEx Delimiter Errors

chadfennell:
In the process of converting LiveView components and views into Surface, developers can run into somewhat confusing parsing errors if they fail to convert their Embedded Elixir `<%= foo = :bar %>` into the Surface Elixir interpolation syntax `{{ foo = :bar }}`.

I bumped into this a couple of times and the "missing closing tag foo" errors had me initially scratching my head until I realized that the parser died due to some embedded Elixir that I had not converted to `{{ }}`. 

This commit raises a more descriptive error alerting developers to the need to change their syntax. I didn't bother capturing the interstitial code between the EEx delimiters or ensure a closing EEx delimiter since most of the time we just want to tip off developers that they have some EEx code in their Surface template.

### DIFF
--- a/lib/surface/compiler/tokenizer.ex
+++ b/lib/surface/compiler/tokenizer.ex
@@ -22,7 +22,7 @@ defmodule Surface.Compiler.Tokenizer do
 
   defp handle_text("<%=" <> _rest, line, column, _buffer, _acc, state) do
     raise parse_error(
-            "EEx syntax `<%=` not allowed. Please use the Surface interpolation syntax `{{ foo = :bar }}`",
+            "EEx syntax `<%= foo = :bar %>` not allowed. Please use the Surface interpolation syntax `{{ foo = :bar }}`",
             line,
             column + 4,
             state
--- a/lib/surface/compiler/tokenizer.ex
+++ b/lib/surface/compiler/tokenizer.ex
@@ -20,6 +20,15 @@ defmodule Surface.Compiler.Tokenizer do
 
   ## handle_text
 
+  defp handle_text("<%=" <> _rest, line, column, _buffer, _acc, state) do
+    raise parse_error(
+            "EEx syntax `<%=` not allowed. Please use the Surface interpolation syntax `{{ foo = :bar }}`",
+            line,
+            column + 4,
+            state
+          )
+  end
+
   defp handle_text("\r\n" <> rest, line, _column, buffer, acc, state) do
     handle_text(rest, line + 1, state.column_offset, ["\r\n" | buffer], acc, state)
   end
--- a/lib/surface/compiler/tokenizer.ex
+++ b/lib/surface/compiler/tokenizer.ex
@@ -22,7 +22,7 @@ defmodule Surface.Compiler.Tokenizer do
 
   defp handle_text("<%=" <> _rest, line, column, _buffer, _acc, state) do
     raise parse_error(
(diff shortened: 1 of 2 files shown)


chadfennell:
I'm not 100% on this approach given the moving parts of the upcoming version, so I am converting it to a draft for now.

msaraiva:
Hi @chadfennell. Thanks for the PR!

> most of the time we just want to tip off developers that they have some EEx code in their Surface template.

In this case, maybe it would be better to have two distinct behaviours:

1. If the `<%= %>` is found inside the node's body, emit a warning instead of an error. This way the user could see every occurrence and fix them all at once instead of performing multiples `run -> error -> fix -> run`. The warning message could be the same you already have, like: *"EEx syntax `<%= ... %>` not allowed. Please use the Surface interpolation syntax `{...}` instead"*

2. If the `<%= %>` is found inside the attributes area then we have no choice but to throw an error, however, 
 the hint can be more useful in this case. Something like *"EEx syntax `<%= ... %>` not allowed. Please use the Surface interpolation syntax `{...}` instead. If you're trying to generate dynamic attributes, please use the `{...@attrs}` notation"*.

I believe `#2` could be easily handled in `handle_maybe_tag_open_end`.

WDYT?

Malian:
If we are going in that direction I believe we should handle all `eex` tags. From the doc, these are the supported tags:

```eex
<% Elixir expression - inline with output %>
<%= Elixir expression - replace with result %>
<%% EEx quotation - returns the contents inside %>
<%# Comments - they are discarded from source %>
```
I would like to have different messages that handle different situations.

For example with the following code, the ideal case I have in mind is:
```eex
<%= if true do %>
  <%= for item <- list do %>
    <p><%= item %></p>
  <% end %>
<% else %>
  <%# TODO %>
  <%%= x + 3 %>
<% end %>
```

- "EEx syntax `<%= if true do %> ...<% end %>` is not allowed. Please use the Surface construct syntax `{#if true} ... {\if}` instead"
- "EEx syntax `<%= for item <- list do %> ...<% end %>` is not allowed. Please use the Surface construct syntax `{#for item <- list} ... {\if}` instead"
- "EEX syntax `<%= item %>` is not allowed. Please use the Surface interpolation syntax `{ item }` instead"
- "EEx syntax `<%# TODO %>` is not allowed. Please use the Surface comment syntax `{!-- TODO --}` instead"
- "EEx syntax `<%%= x + 3 %>` is not allowed. Please use the Surface raw macro component: `<#raw><%= x + 3 %></raw>`
- If the user forgets to replace the <% end %>` with its equivalent, a message could be: "EEX syntax `<% end %>` is not allowed. Did you forget to close a `{\if}` or `{\for}` construct?

From a newcomer point of view, I believe it is really valuable to have that kind of message.

We can start with a more generic message that handles "all" cases: 

EEX syntax `<%= ... %>` and derivatives (`<%`, `<%#`, `<%%`) are not allowed. Please use the Surface equivalents: 

- `<%= if condition do %> ... <% end %>`: `{#if condition}...{\if}`
- `<%= for generator do %> ... <% end %>`: `{#for generator}...{\for}`
- `<%= foo %>`: `{foo}`
- `<%# comment %>`: `{!-- comment --}`
- `<%%= x + 3 %>`: `<#raw><%= x + 3 %></#raw>`
- `<% end %>`: `{\if}` or `{\for}`

WDYT?

chadfennell:
Thank you for the feedback, this all sounds very reasonable to me - I will adjust my PR to incorporate these ideas. Thanks!

PR #395, opened 2021-05-22
Update warning message according to the new syntax

Malian:


lib/surface/type_handler/atom.ex
```diff
@@ -11,7 +11,7 @@ defmodule Surface.TypeHandler.Atom do
     automatic conversion of string literals into atoms is deprecated \
     and will be removed in v0.5.0.
 
-    Hint: replace `#{name}="#{value}"` with `#{name}={{ :#{value} }}`
+    Hint: replace `#{name}="#{value}"` with `#{name}={:#{value}}`
```
msaraiva, line 14: @Malian since this PR is already for `v0.5`, we actually have to remove this implementation along with its tests and let the `literal_to_ast_node ` return `:error`.

Malian: I thought V0.5.0 would do soft depreciation.

msaraiva: We could postpone the removal of the feature and change the message to `...will be removed in v0.6.0` but since the warning was already emitted in `v0.4`, I guess we can just move on ;)

Malian: Oh sorry, I had not seen that the warning message as introduced in `v0.4`. I will update the PR to remove it and remove the implementation.

msaraiva (approved):
❤️

PR #397, opened 2021-05-24
add node translator to parser

lnr0626:
this is essentially purely the node translator from #378


lnr0626:
I'm currently installing elixir 1.12/erlang 24 to run the correct formatter, but beyond that this is ready for a review :-)

msaraiva:
@lnr0626 fantastic! 🎉  I'll review it tomorrow. 

Make sure you're on Elixir `v1.12.0` when running `mix format`. There were some tiny changes introduced in the formatting rules, so the final code is a bit different when running on previous versions.

lib/surface/compiler/node_translator.ex
```diff
@@ -0,0 +1,71 @@
+defmodule Surface.Compiler.NodeTranslator do
+  @type parse_metadata :: %{line: non_neg_integer(), column: non_neg_integer(), file: binary()}
+
+  @type tag_info :: {:tag_open, binary, list(), parse_metadata(), context()}
```
msaraiva, line 4: This can also be `{:block_open, ...}`. 

We need a name that can represent both, `tag` and `block` and rename the state's `tags` field accordingly. Technically, they are still tokens that came from the Tokenizer. Maybe `parent_tokens`, `tokens_stack` or `last_tokens`. Not sure. All terrible names :(

We could also call them just `nodes` but then it might be confusing since we already use `node` for translated nodes in the buffers.

Man, naming is hard :)

lib/surface/compiler/node_translator.ex
```diff
@@ -0,0 +1,71 @@
...
+              state :: state(),
+              context :: context(),
+              name :: binary() | atom(),
+              value :: binary() | {:expr, binary(), parse_metadata()},
+              attr_meta :: parse_metadata()
+            ) :: any()
+
+  @callback handle_init(state :: state()) :: state()
+
+  @callback handle_literal(state :: state(), value :: binary()) :: {state(), any()}
```
msaraiva, line 35: Since we'll no longer handle any value other than string, I believe we should use `handle_text` here. We should also rename the `Ast.Literal` in the future after we make sure we don't expect integer nor booleans anymore.

lib/surface/compiler/tokenizer.ex
```diff
@@ -283,7 +299,8 @@ defmodule Surface.Compiler.Tokenizer do
           column: column,
           line_end: line,
           column_end: new_column,
-          file: state.file
+          file: state.file,
+          void_tag?: name in @void_elements
```
msaraiva, line 303: Good call moving this here 👍

lib/surface/compiler/parse_tree_translator.ex
```diff
@@ -0,0 +1,88 @@
...
+
+  def handle_expression(state, expression, meta) do
+    {state, {:expr, expression, to_meta(meta)}}
+  end
+
+  def handle_comment(state, comment, meta) do
+    {state, {:comment, comment, meta}}
+  end
+
+  def handle_node(state, context, "template", attributes, body, meta) do
```
msaraiva, line 16: I wonder if it would be better to have the node info at the beginning and move `state` and `context` to be the last arguments.

msaraiva (commented):
@lnr0626 great stuff! ❤️

Having the node translator abstraction makes the whole thing extremely flexible. I have no doubt we'll be able to generate the AST directly using this approach.

I left some minor comments for your consideration.

lib/surface/ast.ex
```diff
@@ -45,11 +46,12 @@ defmodule Surface.AST.Expr do
       * `:value` - a quoted expression
       * `:meta` - compile meta
   """
-  defstruct [:value, :meta]
+  defstruct [:value, :meta, directives: []]
```
Malian, line 49: The property `directives` under `## Properties` should be documented.

Same for all structs that newly accepts the property `directives`

lib/surface/ast.ex
```diff
@@ -385,6 +392,31 @@ defmodule Surface.AST.Component do
...
 
+defmodule Surface.AST.MacroComponent do
+  @moduledoc """
+  An AST node representing a macro component
+
+  ## Properties
+      * `:module` - the component module
+      * `:attributes` - the specified attributes
+      * `:directives` - any directives to be applied to this tag
+      * `:children` - the tag children
```
Malian, line 403:
```suggestion
      * `:name` - the macro component name
      *`:body` - the content of the macro component
```

lib/surface/compiler/node_translator.ex
lnr0626: I went for `token_stack`, though `parent_stack` would also be reasonable to me - that indicates it's a hierarchy a bit better... agreed that naming is hard 😂

lnr0626 (commented):
> Having the node translator abstraction makes the whole thing extremely flexible

I haven't explored it (and will likely leave it to others that are more familiar with that codebase), but I suspect that we could re-implement the formatter using a node translator too

> I have no doubt we'll be able to generate the AST directly using this approach.

it might even be feasible to do the entire thing in one pass, though even just generating the AST directly is a nice step in the right direction

msaraiva:
> I haven't explored it (and will likely leave it to others that are more familiar with that codebase), but I suspect that we could re-implement the formatter using a node translator too
> 
> >

Not only the formatter but many other tools. For instance, we could generate tokens for `makeup`. See https://github.com/elixir-makeup/makeup/blob/master/CONTRIBUTING.md#supported-tokens-. This way we could have code highlighting for Surface at compile-time with no JS dependency.

Malian (approved):

lib/surface/ast.ex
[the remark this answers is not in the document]
lnr0626: that makes a lot of sense

lib/surface/ast.ex
lnr0626: i think i got them all

lib/surface/compiler/parse_tree_translator.ex
lnr0626: I also flipped `{state, node}` to `{node, state}`; lmk what you think

msaraiva (approved):
Looks great to me! ❤️ 🎉 🚀

Issue #398, opened 2021-05-25
Compile warning on latest live view


totaltrash:
I've just updated deps (:surface, "0.4.0", :phoenix_live_view, "0.15.7") and I'm getting compile warnings for all of my components `warning: passing the @socket to live_component is no longer necessary, please remove the socket argument`.

The warning points to the `~H` sigil in my render function. I'm guessing it's the new 'socketless' live_component signature.

totaltrash:
fixed by #399

PR #399, opened 2021-05-25
Support Phoenix Live View >= 1.15.6

maennchen:
Fix for #398


Malian:
@maennchen Thanks for opening this PR.

After discussion, we believe we should support  `15.0 <=  phoenix live view <= 15.6` and `phoenix live view >= 15.7`. The main reason is that if we just start using the new phoenix live view version, a lot of people using older versions will get in trouble and will start open issues.

We could release a `v0.4.1` of Surface that can work for both. By just copying/pasting that code and conditionally passing the socket based on the version would do the trick. Then, on `v0.5.0`, we can simply remove the condition and accept only the new version.

WDYT? If it's ok for you, could you please adapt the PR to take these change into account?

maennchen:
@Malian Sure, I'll have a look.

maennchen:
@Malian Is this acceptable?

msaraiva:
@maennchen thanks! ❤️

PR #403, opened 2021-05-27
Move typespecs into appropriate modules

lnr0626:
I added typespecs for the tokenizer, and updated some of the metadata that we generate. Most of the metadata changes were to be consistent (e.g. adding `self_close: false`, updating the comment metadata to include line and column info, always adding the node line/column end values).

The only addition is `macro?: boolean()` to the tag metadata - this allows us to centralize our definition of macro so we don't have to have `when first in ?A..Z? or name == "#raw"` in a few different places. I have another change (https://github.com/lnr0626/surface/compare/lr-translator-types...lnr0626:lr-macro-body-end) that removes the `</#raw>` and `</# <> ...` signatures

### DIFF
--- a/lib/surface/compiler/node_translator.ex
+++ b/lib/surface/compiler/node_translator.ex
@@ -1,93 +1,84 @@
 defmodule Surface.Compiler.NodeTranslator do
-  @type parse_metadata :: %{line: non_neg_integer(), column: non_neg_integer(), file: binary()}
-
-  @typedoc """
-  The token representing the open node for a block.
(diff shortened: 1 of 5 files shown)


msaraiva (approved):
Looks good to me!

I'm just not sure why you needed to handle `#raw` again. Did I miss something in #401?

miguel-s (approved):

lnr0626:
Oh whoops, that's my bad - I'll remove that

PR #404, opened 2021-05-27
Compatibility mode

lnr0626:
to give this a try:

* update your surface dependency to `{:surface, git: "https://github.com/lnr0626/surface.git", ref: "lr-compat-mode"}`
* add the following config to your `config.exs`:
```
config :surface, syntax_compatibility_mode: %{
  Path.expand("lib/") => 4
}
```
* run the converter
* remove that config


msaraiva (approved):
@lnr0626 great job!

I run it against `surface_site` and although there are a few issues to solve, they are mostly related to things that already have been fixed in `surface-next` or things we still need to address, like changing the tokenizer to not parse inside `<style>` and `<script>`.

So I think we can merge this and improve it on `surface-next`.

Thanks for working on this! ❤️

msaraiva:
@lnr0626 are you ok with the following changes:

config.exs:

```
config :surface, syntax_compatibility_mode: %{
  Path.expand("lib/") => 4,
  Path.expand("deps/surface_bulma/lib/") => 4
}
```

compiler.ex:

```elixir
  defp find_syntax_version(file) do
    Enum.find_value(@syntax_versions, @default_syntax_version, fn
      {path, version} -> String.starts_with?(file, path) && version
    end)
  end
```

That would make the converter work with any kind of dependency, including the ones using `path: ...`. We could even validate the config entries using `Path.type/1` so we can validate they are all `:absolute`. 

Let me know if you still have any pending changes to commit, otherwise I'll merge it and we can continue the work in `surface-next`, including the changes above.

Malian:
@lnr0626 I probably miss something but I do not need this code in my projects to convert it to the new syntax:

This is the process I follow:

- Upgrade surface in the mix file (here I set the "surface-next" branch)
- Run mix deps.get && mix deps.compile
- Run mix Surface.ConvertSyntax
- It just works

What do I miss?

msaraiva:
@lnr0626 I believe @Malian suggestion may work.

From slack:

@msaraiva:

> One thing to keep in mind is that it only works because it seems surface itself doesn't have any component using `<For>` or any other component/tag that evaluates properties/attributes at compile-time, however, other components belonging to other dependencies might have. If that's the case, the compilation will fail after conversion. If that happens, the user has to fix the dependency before fixing the project itself, which might be the best approach anyway.
>
> I mean, it's safer to make sure the users have all dependencies migrated before being able to compile the project

lnr0626:
> Let me know if you still have any pending changes to commit

I don't have anything pending, and I'm good with that suggested change

> I probably miss something but I do not need this code in my projects to convert it to the new syntax

That does sound like it'd work, and it's a bit simpler than this. Let's try it on a few other projects and see?

miguel-s:
Just tried @Malian's suggestion:

> Upgrade surface in the mix file (here I set the "surface-next" branch)
> Run mix deps.get && mix deps.compile
> Run mix Surface.ConvertSyntax

Worked without any issues 👍 

Maybe relevant: I only use `~H` and don't have any `.sface` files.

lnr0626:
cool beans, that's a lot simpler. great idea @Malian!

lnr0626:
closing this as it's not needed :-)

PR #408, opened 2021-05-30
New v0.5 syntax

msaraiva:


lib/mix/tasks/surface/convert_syntax.ex
```diff
@@ -0,0 +1,211 @@
+defmodule Mix.Tasks.Surface.ConvertSyntax do
+  @shortdoc "Converts .sface files and ~H sigils from pre-v0.5 to v0.5 syntax"
+
+  @moduledoc """
+  Converts .sface files and ~H sigils from pre-v0.5 to v0.5 syntax.
+
+      mix surface.convert_syntax "lib/**/*.{ex,exs}" "test/**/*.{ex,exs}"
```
Malian, line 7: On my computer, if I run `mix surface.convert_syntax`, the `.sface` files are not converted. I have to force it with a file pattern `mix surface.convert_syntax "lib/**/*.sface"`. I believe we should handle sface files by default.

Malian: The converter only runs against paths configured in `.formatter`.

miguel-s (approved):
LGTM! Didn't encounter any issues in my projects :)

Very excited for v0.5.0!

lib/surface/api.ex
```diff
@@ -100,9 +101,34 @@ defmodule Surface.API do
   end
 
   @doc false
-  def put_assign!(caller, func, name, type, opts, opts_ast, line) do
```
lnr0626, line 103: would it make sense to keep this signature around for now, deprecate it, and remove it in `0.6`?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.5.0 (WIP)
+
+  * [components] ErrorTag: Renamed prop `phx_trigger_action` to `trigger_action`
```
lnr0626, line 5: looks like we forgot to update this haha

lnr0626 (commented):
looking good, i'm also excited for `0.5` :tada:

lib/surface/catalogue/example.ex
```diff
@@ -41,13 +41,13 @@ defmodule Surface.Catalogue.Example do
       @config unquote(opts)
       @before_compile unquote(__MODULE__)
 
-      import Surface, except: [sigil_H: 2]
+      import Surface, except: [sigil_F: 2]
 
-      defmacrop sigil_H({:<<>>, meta, [string]} = ast, opts) do
+      defmacrop sigil_F({:<<>>, meta, [string]} = ast, opts) do
```
lnr0626, line 46:
```suggestion
      defmacrop sigil_F({:<<>>, _meta, [string]} = ast, opts) do
```

it doesn't look like we actually use this variable :-)

msaraiva: I wonder why there's no warning 🤔

lnr0626: could be because it's quoted? though i'd potentially expect warnings on `use Surface.Catalogue.Example` if that were the case

MIGRATING.md
```diff
@@ -0,0 +1,87 @@
...
+      {:surface, github: "surface-ui/surface"},
+      ...
+    ]
+  end
+```
+
+Compile the dependencies:
+
+```
+mix clean && mix deps.get && mix deps.compile
```
Malian, line 55: I am wondering if we shouldn't add the flag `--deps` to `mix clean`? I always did a `rm -rf deps _build` during my test instead of `mix clean`

PR #415, opened 2021-05-31
Replace ~H with ~F to avoid conflicts with HEEx

msaraiva:


miguel-s (approved):
Didn’t review :sweat_smile: But the converter worked fine on my project, turned all `~H` into `~F`

PR #416, opened 2021-06-01
Handle tagged expressions `{... }` and `{= }`

msaraiva:
* Add shorthand notation for dynamic attrs/props with `{... }`

* Add shorthand notation for self-assign attrs/prop with `{= }`

### DIFF
--- a/lib/surface/compiler/parse_tree_translator.ex
+++ b/lib/surface/compiler/parse_tree_translator.ex
@@ -101,7 +101,7 @@ defmodule Surface.Compiler.ParseTreeTranslator do
 
     message = """
     cannot assign `{...#{value}}` to attribute `#{name}`. \
-    Tagged expression `{... }` can only be defined as root attribute/property.
+    The tagged expression `{... }` can only be used on a root attribute/property.
 
     Example: <div {...@attrs}>
     """
@@ -144,7 +144,7 @@ defmodule Surface.Compiler.ParseTreeTranslator do
 
     message = """
     cannot assign `{=#{value}}` to attribute `#{name}`. \
-    Tagged expression `{= }` can only be defined as root attribute/property.
+    The tagged expression `{= }` can only be used on a root attribute/property.
 
     Example: <div {=@class}>
     """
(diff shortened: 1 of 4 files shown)


lib/surface/compiler/parse_tree_translator.ex
```diff
@@ -68,6 +68,90 @@ defmodule Surface.Compiler.ParseTreeTranslator do
...
+        {:tagged_expr, "...", expr, marker_meta},
+        _attr_meta,
+        _state,
+        _context
+      ) do
+    {:expr, value, _expr_meta} = expr
+
+    message = """
+    cannot assign `{...#{value}}` to attribute `#{name}`. \
+    Tagged expression `{... }` can only be defined as root attribute/property.
```
lnr0626, line 104:
```suggestion
    The tagged expression `{... }` can only be used on a root attribute/property.
```

lib/surface/compiler/parse_tree_translator.ex
```diff
@@ -68,6 +68,90 @@ defmodule Surface.Compiler.ParseTreeTranslator do
...
+        {:tagged_expr, "=", expr, marker_meta},
+        _attr_meta,
+        _state,
+        _context
+      ) do
+    {:expr, value, _expr_meta} = expr
+
+    message = """
+    cannot assign `{=#{value}}` to attribute `#{name}`. \
+    Tagged expression `{= }` can only be defined as root attribute/property.
```
lnr0626, line 147:
```suggestion
    The tagged expression `{= }` can only be used on a root attribute/property.
```

lib/surface/compiler/parse_tree_translator.ex
```diff
@@ -109,4 +193,24 @@ defmodule Surface.Compiler.ParseTreeTranslator do
       :ignored_body?
     ])
   end
+
+  defp strip_name_from_tagged_expr_equals!(value, expr_meta) do
+    case Code.string_to_quoted(value) do
```
lnr0626, line 198: since we have such a limited set of expressions we support `{= ...}` to be used on (and a fairly small set of supported characters for elixir identifiers), would it make sense to just use a regex with a capture group? or is that going to just be harder to maintain with very little benefit?

lnr0626 (commented):
looking good, just had a few small suggestions for your consideration

lib/surface/compiler/parse_tree_translator.ex
msaraiva: I thought about doing that but for some reason I find it more intuitive using `string_to_quoted` than using regex. I'm not sure why. Maybe it's because I'm always forgetting about some edge case when dealing with regular expressions :)

PR #417, opened 2021-06-02
Add case/match block

msaraiva:
I decided to create generic `AST.Block` and `AST.SubBlock` nodes to represent `case/match`. I think it simplifies the code a lot. And now that we have all constructs standardized with the same shape, I don't think it's necessary to have a specific node for each one of them.

I believe all other blocks (if/elseif/else, for/else, ...) could be much easier to implement using this approach as they would share the implementation in `Surface.Compiler`. The code in `Surface.Compiler.EExEngine` also gets much simpler, IMO. Here's the `to_expression/3` implementation for `case`:

```elixir
  defp to_expression(%AST.Block{name: "case"} = block, buffer, state) do
    %AST.Block{expression: case_expr, sub_blocks: sub_blocks} = block

    state = %{state | depth: state.depth + 1, context: [:case | state.context]}

    match_blocks =
      Enum.flat_map(sub_blocks, fn %AST.SubBlock{children: children, expression: expr} ->
        match_body = handle_nested_block(children, buffer, state)

        quote do
          unquote(expr) -> unquote(match_body)
        end
      end)

    quote do
      case unquote(case_expr) do
        unquote(match_blocks)
      end
    end
    |> maybe_print_expression(block)
  end
 ```

I also implemented some of the validation in `Surface.Compiler.ParseTreeTranslator` as it contains information about the parent block. All validations should be unified later when we start generating the AST directly in the node translator.

**Quick general note:** as you can see, I'm starting to avoid pattern matching everything I can in the head of the functions. If something is not required to match in order to distinguish the different function clauses, I prefer to have an extra line right below the head just to extract the other information. This makes it much easier to read the function head as well as preventing it from be broken by the formatter. However, if the function head is small enough, I don't see a problem extracting everything there.

### DIFF
--- a/lib/surface/compiler/helpers.ex
+++ b/lib/surface/compiler/helpers.ex
@@ -26,8 +26,8 @@ defmodule Surface.Compiler.Helpers do
     @builtin_assigns_by_type[type]
   end
 
-  def interpolation_to_quoted!(text, meta) do
-    case Code.string_to_quoted(text, file: meta.file, line: meta.line) do
+  def expression_to_quoted!(text, meta) do
+    case Code.string_to_quoted(text, file: meta.file, line: meta.line, column: meta.column) do
       {:ok, expr} ->
         expr
 
@@ -102,11 +102,11 @@ defmodule Surface.Compiler.Helpers do
     assigns
   end
 
-  def to_meta(tree_meta, %CompileMeta{caller: caller, checks: checks}) do
+  def to_meta(tree_meta, %CompileMeta{caller: caller, checks: checks} = compile_meta) do
     %AST.Meta{
-      line: tree_meta.line,
-      column: tree_meta.column,
-      file: tree_meta.file,
+      line: tree_meta[:line] || compile_meta.line,
+      column: tree_meta[:column],
+      file: tree_meta[:file] || compile_meta.file,
       caller: caller,
       checks: checks
     }
--- a/lib/surface/compiler/helpers.ex
+++ b/lib/surface/compiler/helpers.ex
@@ -102,11 +102,11 @@ defmodule Surface.Compiler.Helpers do
     assigns
   end
 
-  def to_meta(tree_meta, %CompileMeta{caller: caller, checks: checks} = compile_meta) do
+  def to_meta(tree_meta, %CompileMeta{caller: caller, checks: checks}) do
     %AST.Meta{
-      line: tree_meta[:line] || compile_meta.line,
-      column: tree_meta[:column],
-      file: tree_meta[:file] || compile_meta.file,
+      line: tree_meta.line,
+      column: tree_meta.column,
+      file: tree_meta.file,
       caller: caller,
       checks: checks
     }
--- a/lib/surface/compiler/parse_tree_translator.ex
+++ b/lib/surface/compiler/parse_tree_translator.ex
@@ -2,6 +2,7 @@ defmodule Surface.Compiler.ParseTreeTranslator do
   @behaviour Surface.Compiler.NodeTranslator
 
   alias Surface.IOHelper
+  alias Surface.Compiler.Helpers
 
   def handle_init(state), do: state
 
@@ -44,6 +45,15 @@ defmodule Surface.Compiler.ParseTreeTranslator do
 
   def handle_block(name, expr, body, meta, state, _context) do
     {{:block, name, expr, body, to_meta(meta)}, state}
+  end
+
+  def handle_subblock(:default, expr, children, meta, state, %{parent_block: "case"}) do
+    if !Helpers.blank?(children) do
+      message = "cannot have content between {#case ...} and {#match ...}"
+      IOHelper.compile_error(message, meta.file, meta.line)
+    end
+
+    {{:block, :default, expr, [], %{}}, state}
   end
 
   def handle_subblock(:default, expr, children, _meta, state, _context) do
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: lib/surface/compiler/parse_tree_translator.ex [hunk 1]
```diff
@@ -164,12 +174,12 @@ defmodule Surface.Compiler.ParseTreeTranslator do
...
-    nil
+  def context_for_subblock(_name, _meta, _state, parent_context) do
+    parent_context
   end
 
-  def context_for_block(_name, _meta, _state) do
-    nil
+  def context_for_block(name, _meta, _state) do
+    %{parent_block: name}
   end
```
lnr0626, line 183: a slight alternative:

```
def context_for_subblock(name, _meta, _state, parent_context) do
  %{sub_block: name, parent_block: Map.get(parent_context, :block_name)}
end

def context_for_block(name, _meta, _state_ do
  %{block_name: name}
end
```

my reasoning is that I think the attribute `parent_block` makes sense in the subblock context, I'm not sure it does for the block context

lib/surface/compiler/helpers.ex
```diff
@@ -102,11 +102,11 @@ defmodule Surface.Compiler.Helpers do
     assigns
   end
 
-  def to_meta(tree_meta, %CompileMeta{caller: caller, checks: checks}) do
+  def to_meta(tree_meta, %CompileMeta{caller: caller, checks: checks} = compile_meta) do
     %AST.Meta{
-      line: tree_meta.line,
-      column: tree_meta.column,
-      file: tree_meta.file,
+      line: tree_meta[:line] || compile_meta.line,
```
lnr0626, line 107: is this change because the default sub block has empty meta? If that's the case, we could also update https://github.com/surface-ui/surface/blob/05d8cfbb1a8b2e70547ab550852faaab62ccebcc/lib/surface/compiler/parse_tree_translator.ex#L60 and https://github.com/surface-ui/surface/blob/05d8cfbb1a8b2e70547ab550852faaab62ccebcc/lib/surface/compiler/parse_tree_translator.ex#L56 to pass `to_meta(meta)` instead of `%{}`

lnr0626 (commented):
> as you can see, I'm starting to avoid pattern matching everything I can in the head of the functions

i love it!

> I decided to create generic AST.Block and AST.SubBlock nodes to represent case/match. I think it simplifies the code a lot

It definitely simplifies the code, though I think there are some other approaches we could take and so I'm not fully convinced this is the way to go yet.

I think my main hesitance with a common, generic `AST.Block`and `AST.SubBlock` is that having specific nodes for the different block expressions makes it easier to know that the blocks are actually valid (more important when creating the nodes by hand, but somewhat relevant imo), and to give better names. For instance, the `Surface.Compiler.EExEngine ` would be very similar with an `AST.Case` and `AST.Match`:

```elixir
  defp to_expression(%AST.Case{} = block, buffer, state) do
    state = %{state | depth: state.depth + 1, context: [:case | state.context]}

    match_blocks =
      Enum.flat_map(block.matches, fn %AST.Match{children: children, expression: expr} ->
        match_body = handle_nested_block(children, buffer, state)

        quote do
          unquote(expr) -> unquote(match_body)
        end
      end)

    quote do
      case unquote(block.expression) do
        unquote(match_blocks)
      end
    end
    |> maybe_print_expression(block)
  end
```

Doing something similar for `{#if}` might look like 
```
%AST.If{
  branches: [
    %AST.Branch{condition: ..., children: [...]},
    %AST.Branch{condition: ..., children: [...]},
    %AST.Else{children: [...]}
  ]
}
# though the above has similar issues with assuming that If.branches is well formed, so another
# option might be:
%AST.If{
  branches: [
    %AST.Branch{condition: ..., children: [...]},
    %AST.Branch{condition: ..., children: [...]}
  ],
  else: %AST.Else{children: [...]}
}
```

and for `{#for}`:
```
%AST.For{
  generator: ...,
  children: [...],
  else: %AST.Else{...}
}
```

that being said, I haven't fully thought through either of those suggestions yet 🙂 

A somewhat hybrid approach might be to keep `AST.SubBlock` around, but have specific nodes for each of the sub blocks (i.e. `match`, `else`, `elseif`, ...)

Any way we go, we'll likely still need separate cases for most if not all of the subblock/block pairs that are supported, and ideally most of the current code in `Surface.Compiler` would no longer be needed once we're generating the ast from a node translator

msaraiva:
> I think my main hesitance with a common, generic `AST.Block`and `AST.SubBlock` is that having specific nodes for the different block expressions makes it easier to know that the blocks are actually valid (more important when creating the nodes by hand, but somewhat relevant imo), and to give better names. For instance, the `Surface.Compiler.EExEngine ` would be very similar with an `AST.Case` and `AST.Match`

> makes it easier to know that the blocks are actually valid (more important when creating the nodes by hand, but somewhat relevant imo)

We should make sure the AST is always valid when it hits the `EExEngine`. We actually shouldn't allow building it manually outside the compiler. The only reason we still allow that is because we haven't implemented a way build the AST injecting custom nodes, similar to what `quote/unquote` do. Take `Elixir` itself, it doesn't have a specific nodes for each statement, buy it provides a nice way to built and validate the structure. We need our equivalent `quote/unquote` to make sure the AST is validated by the same rules as the regular template.

msaraiva:
> A somewhat hybrid approach might be to keep `AST.SubBlock` around, but have specific nodes for each of the sub blocks (i.e. `match`, `else`, `elseif`, ...)
> 
> Any way we go, we'll likely still need separate cases for most if not all of the subblock/block pairs that are supported, and ideally most of the current code in `Surface.Compiler` would no longer be needed once we're generating the ast from a node translator

We can postpone this decision until we start working on generating the AST directly.

lnr0626 (approved):
lgtm!

PR #422, opened 2021-06-04
Fix 411

### LINKED ISSUE #411: Boolean attributes not handled correctly
I have some buttons that I'd like to disable based on changeset being invalid etc, currently this is more difficult than it should be while using Surface components (such as the inbuilt `<Submit>` button), passing a boolean value for the disabled attribute doesn't work as the HTML spec requires, and also not as it works in Phoenix.

https://hexdocs.pm/phoenix_html/Phoenix.HTML.Tag.html#tag/1-boolean-values

I made a change to `opts_to_attrs` in the `utils.ex` file, but it doesn't seem to update  when the state changes, I most likely have done it in the wrong place, and `opts_to_attrs` is a one time thing only.

Below is my modified `opts_to_attrs/1`

```
def opts_to_attrs(opts) do
    for {key, value} <- opts do
      case {key, value} do
        {:phx_capture_click, _} -> {:"phx-capture-click", value}
        {:phx_click, _} -> {:"phx-click", value}
        {:phx_window_focus, _} -> {:"phx-window-focus", value}
        {:phx_window_blur, _} -> {:"phx-window-blur", value}
        {:phx_focus, _} -> {:"phx-focus", value}
        {:phx_blur, _} -> {:"phx-blur", value}
        {:phx_window_keyup, _} -> {:"phx-window-keyup", value}
        {:phx_window_keydown, _} -> {:"phx-window-keydown", value}
        {:phx_keyup, _} -> {:"phx-keyup", value}
        {:phx_keydown, _} -> {:"phx-keydown", value}
        {:phx_target, _} -> {:"phx-target", value}
        {:data, _} -> data_to_attrs(value)
        {_, true} -> {:"#{key}", key}
        {_, false} -> nil
        _ -> {key, value}
      end
    end
    |> List.flatten()
    |> Enum.reject(&is_nil/1)
  end
```

miguel-s:
Fixes #411 

Change tracking is broken for a couple of the builtin-in components.

This PR should fix it. Credit for finding the cause goes to @Malian!

#### Components
- [x] Submit
- [x] Link
- [x] Link.Button


lib/surface/components/form/submit.ex
```diff
@@ -23,11 +23,13 @@ defmodule Surface.Components.Form.Submit do
...
-    opts = prop_to_attr_opts(assigns.class, :class) ++ assigns.opts ++ events_to_opts(assigns)
-    attrs = opts_to_attrs(opts)
-
     ~F"""
-    <button type="submit" :attrs={attrs}><#slot>{@label}</#slot></button>
+    <button type="submit" :attrs={to_attrs(assigns)}><#slot>{@label}</#slot></button>
     """
   end
+
+  def to_attrs(assigns) do
```
Malian, line 31: I believe we could use a private function, WDYT?

lib/surface/components/link.ex
```diff
@@ -89,4 +83,9 @@ defmodule Surface.Components.Link do
       [data: [method: method, to: to] ++ csrf_data] ++ opts
     end
   end
+
+  def to_attrs(assigns, to) do
```
Malian, line 87: Same

lib/surface/components/link/button.ex
```diff
@@ -91,4 +85,9 @@ defmodule Surface.Components.Link.Button do
       [data: [method: method, to: to] ++ csrf_data] ++ opts
     end
   end
+
+  def to_attrs(assigns, to) do
```
Malian, line 89: Same

Malian (approved):
We should take a look at surface-bootstrap to see how @olivermt handles this case.

lib/surface/components/form/submit.ex
miguel-s: Definitely, that was a silly oversight.

miguel-s:
@Malian looks like surface-bootstrap does not allow arbitrary attributes with a generic `opts` prop. Each attribute gets its own prop.

PR #425, opened 2021-06-07
evaluate literal expressions at compile time

lnr0626:
This came about because I was noticing a lot of `class="..."` and similar in the websocket diff payloads for literal values, and I was curious how much bandwidth we could save if we evaluated these literal expressions at compile time instead of runtime (which would be more in line with what would happen using pure live view).

This should only effect expressions which are essentially quoted literals, and should not effect any expression that calls a function or uses a variable/assign.

I tested it out on a project of mine locally that uses tailwind extensively, and it resulted in ~10% less data over the wire for diff payloads. I didn't do extensive testing on the savings though, so that number should be taken with a grain of salt.


msaraiva (approved):
@lnr0626 very nice optimization 👍 ❤️

PR #426, opened 2021-06-07
Add surface_quote/2 macro to build AST from templates

msaraiva:


lib/surface/ast.ex
```diff
@@ -540,3 +522,28 @@ defmodule Surface.AST.SlotableComponent do
...
+          | Surface.AST.MacroComponent.t()
+          | Surface.AST.SlotableComponent.t()
+          | Surface.AST.Error.t()
+
+  def find_attribute_value(attributes, name) do
+    Enum.find_value(attributes, fn
+      %Surface.AST.Attribute{name: ^name, value: value} -> value
+      _ -> nil
+    end)
+  end
```
msaraiva, line 548: Not sure if it's the best place to add this function, however, wherever we decide to put it, we should start  moving similar functions to this module too.

I also think we should start considering all AST structs as private API and provide functions to manipulate them. My intention with the new `surface_quote` macro is to avoid any kind of manual build of AST.

lib/surface/compiler.ex
```diff
@@ -94,7 +95,8 @@ defmodule Surface.Compiler do
       line: line,
       file: file,
       caller: caller,
-      checks: opts[:checks] || []
+      checks: opts[:checks] || [],
+      variables: opts[:variables] || []
```
msaraiva, line 99: I'm not happy adding this field to `CompileMeta` but I couldn't find another way to pass the list of existing variables to be unquoted by `surface_quote`.

lib/surface/ast.ex
lnr0626: I think that makes a lot of sense. `Surface.AST` seems like a reasonable place to hold the ast manipulation functions.

Would you envision doing something like I did with `Surface.AST.AttributeExpr.new/3` for the other nodes? Or stick to building the structs with `%Struct{}`? If so, would you suggest moving that type of function to this module as well? (i.e. `Surface.AST.attribute_expr/3` - I debated about doing this originally)

lnr0626: It would probably make sense to split this file up (not necessarily in this pr) - either into a `surface/ast.ex` that has this module and one `surface/ast/{meta,literal,attribute_expr, ...}` for each node or a `surface/ast.ex` and a single  `surface/ast/nodes.ex` that has all of the nodes

lib/surface/compiler.ex
lnr0626: I'm going to spend a bit of time messing around with some things to see if I can come up with an alternative, but I'm not positive currently

lnr0626 (approved):
We should probably also add a test for using `{^ ...}` outside of `surface_quote` - i.e.

```elixir
def render(assigns) do
  ~F"""
    <div class={^something}></div>
    """
end
```

just to make sure we cover that case

lib/surface/compiler.ex
```diff
@@ -284,6 +288,11 @@ defmodule Surface.Compiler do
      }}
   end
 
+  defp convert_node_to_ast(:ast, {_, variable, expr_meta}, compile_meta) do
+    ast = fetch_quoted_value!(compile_meta, variable, expr_meta)
```
lnr0626, line 292:
```suggestion
    meta = Helpers.to_meta(expr_meta, compile_meta)
    ast = fetch_quoted_value!(compile_meta, variable, meta)
```

lib/surface/compiler.ex
```diff
@@ -791,21 +801,25 @@ defmodule Surface.Compiler do
...
+  defp attr_value(name, type, {:attribute_expr, value, expr_meta}, attr_meta, _compile_meta) do
     expr_meta = Helpers.to_meta(expr_meta, attr_meta)
     expr = Surface.TypeHandler.expr_to_quoted!(value, name, type, expr_meta)
 
     AST.AttributeExpr.new(expr, value, expr_meta)
   end
 
-  defp attr_value(name, type, value, meta) do
+  defp attr_value(_name, _type, {:ast, variable, expr_meta}, _attr_meta, compile_meta) do
+    fetch_quoted_value!(compile_meta, variable, expr_meta)
```
lnr0626, line 819:
```suggestion
    meta = Helpers.to_meta(expr_meta, compile_meta)
    fetch_quoted_value!(compile_meta, variable, meta)
```

lnr0626 (commented):
a couple suggestions so we're consistent with parse meta vs `ast.meta`

lnr0626:
I tried a couple things, but couldn't come up with anything better than adding `variables` to `CompileMeta`

lib/surface/ast.ex
msaraiva: I think those are good suggestions and yes, it's probably a good idea to implement them on a separate PR along with other changes we can discuss to simplify the AST, including the use of the `Block` and `SubBlock` nodes to replace other nodes. My feeling is that `surface_quote` was be the missing tool for building AST nodes easier and safer. The goal is to use it not only in macro components but also in the compiler itself.

lib/surface/compiler.ex
```diff
@@ -1153,4 +1167,58 @@ defmodule Surface.Compiler do
       IOHelper.warn(message, meta.caller, fn _ -> meta.line end)
     end
   end
+
+  defp fetch_quoted_value!(compile_meta, variable, expr_meta) do
+    variable =
+      if Regex.match?(~r/^[a-z][a-zA-Z_\d]*$/, variable) do
```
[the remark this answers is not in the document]
msaraiva: For now, it's probably better to start with a more strict approach as we don't accept arbitrary expressions, only variables. We can change it later, if we need to.

PR #428, opened 2021-06-09
Rearrange tests

msaraiva:
Just moving tests directly related to a specific module to its counterpart `test/surface/...` folder, which is the convention. All other tests went to `test/surface/integrations/`.

I also adjusted some of the test module names that were not matching their test files.

> **Note**: some tests in `integrations` should probably be moved to some of the component's own tests but I don't think we should do this in this PR.


miguel-s (approved):

miguel-s:
Should make finding the relevant test easier 👍

lnr0626 (approved):
Lgtm

PR #429, opened 2021-06-10
Add static option for prop

msaraiva:
As we can now use dynamic expressions's in macro components, we need a way to differentiate props that are meant to be used at compile-time and props that will be forwarded to the generated AST and evaluated at runtime. 

This PR adds a `:static` option to `prop` to achieve that goal. This option is used by `MacroComponent.eval_static_props!/3` so it bypass the runtime props while evaluating/validating the values.

### Usage

```
@doc "The value of this prop will be used at compile-time"
prop unwrap, boolean, static: true
```

### DIFF
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -404,7 +404,7 @@ defmodule Surface.API do
   end
 
   defp get_valid_opts(:prop, _type, _opts) do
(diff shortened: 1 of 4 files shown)


lnr0626 (approved):
This is nice!

PR #430, opened 2021-06-10
remove deprecations that are automatically handled by the converter

lnr0626:


lib/surface/compiler/parser.ex
```diff
@@ -288,26 +288,16 @@ defmodule Surface.Compiler.Parser do
...
+    raise parse_error(
+            """
+            Surface does not support passing non-string attribute values as literals \
+            (i.e. `selected=true` or `tabindex=3`).
+
+            Hint: replace `#{name}=#{value}` with `#{name}={#{value}}`
+            """,
+            meta
+          )
   end
```
msaraiva, line 301: One thing just crossed my mind. 

Since unquoted attribute values are accepted in HTML, maybe we should keep this as a warning and the conversion to string, which will keep the rendering consistent with the original HTML. This way copied/pasted HTML would still run and the user will have the opportunity to see all occurrences of this problem instead of a single error pointing to the first one.

We could also try go give a bit of context about the motivation. Something like:

```
the use of unquoted attribute values is not recommended as they will always be converted to strings.

For intance, `selected=true` and `tabindex=3` will be translated to `selected="true"` and `tabindex="3"` respectively.

Hint: if you want to pass a literal boolean or integer, replace `#{name}=#{value}` with `#{name}={#{value}}`
```

WDYT?

lnr0626: that makes sense - I'll make that change

msaraiva (approved):
Awesome! ❤️

miguel-s (approved):
lgtm!

lib/surface/compiler/tokenizer.ex
```diff
@@ -660,6 +660,7 @@ defmodule Surface.Compiler.Tokenizer do
 
   ## handle_attr_value_unquoted
 
+  # TODO: remove after 0.5 release
```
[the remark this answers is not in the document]
lnr0626: 👍 good catch

PR #433, opened 2021-06-11
Rename slot props to args

msaraiva:


msaraiva:
@lnr0626 I just renamed things around as needed. You can just take a quick look to make sure I didn't forget anything but it should be fine to merge.

PR #435, opened 2021-06-16
Replace slot option :props with :args

msaraiva:
* [converter] Rename :props to :args 
  * Fix error message when no argument is defined for the slot
  * Update migration guide


miguel-s (approved):
lgtm!

Issue #436, opened 2021-06-17
ErrorTag crashing when an error has an opt that can't be converted to a string


kfajdsl:
I found this out trying to use Changeset.unsafe_validate_unique/4, which has an opt with an array value in its error. This causes an ArgumentError when ErrorTag tries to interpolate the error opts into the message in ErrorTag.translate_error/1. 

I'll open a PR with my fix, but I'm not sure if it's the best solution (I'm pretty new to Elixir!).

Malian:
@kfajdsl  Thanks for reporting this. Could you try to provide the default translator for `ErrorTag` as explained here? https://surface-ui.org/builtincomponents/ErrorTag

```elixir
config :surface, :components, [
  {Surface.Components.Form.ErrorTag, default_translator: {MyAppWeb.ErrorHelpers, :translate_error}}
]
```

If it does not fix your issue, could you please provide us a simple example that reproduces the bug you encounter?

kfajdsl:
Aha, that does fix my issue! Can't believe I missed that. So is the `translate_error` found in the component itself just a minimal fallback then?

Thanks for the help, I'll close the issue and PR.

Malian:
@kfajdsl exactly. This is a c/p from phoenix. Glad that solution works for you!

Issue #276, opened 2021-03-01
ErrorTag problem with unsafe_unique changeset validation error


hok:
It looks like the ErrorTag component doesn't handle the case of receiving a changeset that contains a validation error for uniqueness validation. In this case, the validations fields value is a list and not a value that can be converted to a string.

```
errors: [
    email: {"has already been taken",
     [validation: :unsafe_unique, fields: [:email]]}
  ]
```

The error is the following:

```
** (ArgumentError) cannot convert the given list to a string.

To be converted to a string, a list must either be empty or only
contain the following elements:

  * strings
  * integers representing Unicode code points
  * a list containing one of these three elements

Please check the given list or call inspect/1 to get the list representation, got:

[:email]

    (elixir 1.11.3) lib/list.ex:925: List.to_string/1
    (surface 0.3.0) lib/surface/components/form/error_tag.ex:127: anonymous fn/2 in Surface.Components.Form.ErrorTag.translate_error/1
    (elixir 1.11.3) lib/enum.ex:2193: Enum."-reduce/3-lists^foldl/2-0-"/3
    (surface 0.3.0) lib/surface/components/form/error_tag.ex:112: anonymous fn/7 in Surface.Components.Form.ErrorTag.render/1
    (elixir 1.11.3) lib/enum.ex:2193: Enum."-reduce/3-lists^foldl/2-0-"/3
    (surface 0.3.0) lib/surface/components/form/error_tag.ex:109: anonymous fn/6 in Surface.Components.Form.ErrorTag.render/1
    (phoenix_live_view 0.15.4) lib/phoenix_live_view/diff.ex:342: Phoenix.LiveView.Diff.traverse/6
    (phoenix_live_view 0.15.4) lib/phoenix_live_view/diff.ex:430: anonymous fn/4 in Phoenix.LiveView.Diff.traverse_dynamic/6
```

Malian:
Hello @hok ! Thanks you for reporting this.

When using the Surface ErrorTag component, you have to specify a translator in the `Surface` config. The documentation of the `Surface.Components.Form.ErrorTag` module says:

> When using Gettext, you can use configuration to route all errors through the `MyAppWeb.ErrorHelpers.translate_error/1` function generated by Phoenix, which utilizes `Gettext`. You need to provide a tuple with the module and the name of the function (as an atom) as follows

  ```elixir
  config :surface, :components, [
[...]

PR #449, opened 2021-07-04
Improve error message when default_translator is not configured for ErrorTag

### LINKED ISSUE #450: Bug: ErrorTag crashes on Ecto.Enum cast error.
## Describe the bug

ErrorTag crashes LV when there is cast error on a schema field being Ecto.Enum type.

```
[error] GenServer #PID<0.562.0> terminating
** (Protocol.UndefinedError) protocol String.Chars not implemented for {:parameterized, Ecto.Enum, %{on_dump: %{B2B: "B2B", EoC: "EoC"}, on_load: %{"B2B" => :B2B, "EoC" => :EoC}, type: :string, values: [:B2B, :EoC]}} of type Tuple
    (elixir 1.12.1) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir 1.12.1) lib/string/chars.ex:22: String.Chars.to_string/1
    (surface 0.5.0) lib/surface/components/form/error_tag.ex:127: anonymous fn/2 in Surface.Components.Form.ErrorTag.translate_error/1
    (elixir 1.12.1) lib/enum.ex:2356: Enum."-reduce/3-lists^foldl/2-0-"/3
    (surface 0.5.0) lib/surface/components/form/error_tag.ex:112: anonymous fn/7 in Surface.Components.Form.ErrorTag.render/1
```

Probably `translate_error` function doesn't handle Ecto.Enum type properly:
```
def translate_error({msg, opts}) do
    # Because the error messages we show in our forms and APIs
    # are defined inside Ecto, we need to translate them dynamically.
    Enum.reduce(opts, msg, fn {key, value}, acc ->
      String.replace(acc, "%{#{key}}", to_string(value))
    end)
  end
```

`to_string` function receives and cannot handle below tuple which is a part of meta keyword list under errors in a changeset:
```
{:parameterized, Ecto.Enum,
        %{
          on_dump: %{B2B: "B2B", EoC: "EoC"},
          on_load: %{"B2B" => :B2B, "EoC" => :EoC},
          type: :string,
          values: [:B2B, :EoC]
        }
```

Errors key under Changeset:
```
 errors: [
    company_id: {"can't be blank", [validation: :required]},
    type: {"is invalid",
     [
       type: {:parameterized, Ecto.Enum,
        %{
          on_dump: %{B2B: "B2B", EoC: "EoC"},
          on_load: %{"B2B" => :B2B, "EoC" => :EoC},
          type: :string,
          values: [:B2B, :EoC]
        }},
       validation: :cast
     ]},
    country_id: {"is invalid", [type: :id, validation: :cast]}
  ],
  data: #Flip.Schemas.DB.Employee<>,
  valid?: false
```

## How to reproduce it

Employee.ex
```
defmodule Example.Schemas.DB.Employee do
  use Ecto.Schema
  alias Ecto.Changeset

  schema "employee" do
    field :type, Ecto.Enum, values: [:B2B, :EoC]

  def changeset(%__MODULE__{} = employee, params) do
    employee
    |> Changeset.cast(params, all_fields())
 end

  defp all_fields do
    __MODULE__.__schema__(:fields)
  end
```

Form.ex (only render function)

```
  @impl true
  def render(assigns) do
    ~F"""

    <div class="container is-fullhd">

        <h2 class="title">
          New consultant form:
        </h2>

      <div class="tile">
        <div class="column is-half">

      <!-- Form begins here -->
        <Form for={ @changeset } submit="create_consultant" change="validate_consultant">

          <Field name={ :type } class="field">
            <Label class="label" />
            <TextInput class="input is-normal" />
            <ErrorTag />
          </Field>

          <!-- Form submit button -->
          <Submit class="button is-success">
            Submit
          </Submit>
        </Form>

        <!-- Form ends here -->
      </div>
    </div>
    </div>
    """
  end
```

## The behavior you expected
'Is invalid' error message displayed under a field of Ecto.Enum type when TextInput provided with a not allowed value - smoething else than :B2B or :CoE  `field :type, Ecto.Enum, values: [:B2B, :EoC]`

## Your Environment
Surface: v0.5.0
LiveView: v0.15.1
Elixir: v1.12.1

Malian:
Related to #276, #436 and #450. 
Fixes #450

When users use `unsafe_unique` validation, it results in an ambiguous error message that confuses the user. Most of the time this issue raises because the user hasn't set the `default_translator` for the `ErrorTag` component.

The users believe it is a Surface issue and report it on Slack or here on github. I would like to avoid this overhead and help the users upfront.

Here is a proposition to improve the default `phoenix` error message with our own hint that suggests to set the `default_translator`

@miguel-s @msaraiva WDYT?

### DIFF
--- a/lib/surface/components/form/error_tag.ex
+++ b/lib/surface/components/form/error_tag.ex
@@ -144,7 +144,7 @@ defmodule Surface.Components.Form.ErrorTag do
               ]
 
             Given value: #{inspect(value)}
-            
+
             Exception: #{Exception.message(e)}
             """,
             __STACKTRACE__
--- a/lib/surface/components/form/error_tag.ex
+++ b/lib/surface/components/form/error_tag.ex
@@ -21,6 +21,10 @@ defmodule Surface.Components.Form.ErrorTag do
   ]
   ```
 
+  Surface provides a fallback error message if the error cannot be converted to a string. If it
+  happens, Surface will raise a warning to invite you to configure a `default_translator` that
+  handles such a case.
+
   There is also a `translator` prop which can be used on a case-by-case basis.
   It overrides the configuration.
 
@@ -45,7 +49,7 @@ defmodule Surface.Components.Form.ErrorTag do
   ```
 
   ```surface
-  <ErrorTag translator={{ &CustomTranslationLib.translate_error/1 }} />
+  <ErrorTag translator={&CustomTranslationLib.translate_error/1} />
   ```
   """
 
@@ -127,21 +131,23 @@ defmodule Surface.Components.Form.ErrorTag do
       try do
         String.replace(acc, "%{#{key}}", to_string(value))
       rescue
-        e in ArgumentError ->
-          if e.message =~ "cannot convert the given list to a string" do
-            message = """
-            Hint: In Surface, this error often happens because no `default_translator` has been set
-            for the `ErrorTag` component. You can point the default translator to your application helpers:
+        e ->
+          IO.warn(
+            """
+            the fallback message translator for the `ErrorTag` component cannot handle the given message.
+
+            Hint: you can set up the `default_translator` to route all errors to your application helpers:
 
               config :surface, :components, [
                 {Surface.Components.Form.ErrorTag, default_translator: {MyAppWeb.ErrorHelpers, :translate_error}}
               ]
-            """
 
-            reraise %{e | message: e.message <> message}, __STACKTRACE__
-          else
-            reraise e, __STACKTRACE__
-          end
+            Original error: #{Exception.message(e)}
+            """,
+            __STACKTRACE__
+          )
+
+          "invalid value"
       end
     end)
   end
--- a/lib/surface/components/form/error_tag.ex
+++ b/lib/surface/components/form/error_tag.ex
@@ -135,7 +135,7 @@ defmodule Surface.Components.Form.ErrorTag do
         e ->
           IO.warn(
             """
-            the fallback message translator for the `ErrorTag` component cannot handle the given message.
+            the fallback message translator for the `ErrorTag` component cannot handle the given value.
 
             Hint: you can set up the `default_translator` to route all errors to your application helpers:
 
--- a/lib/surface/components/form/error_tag.ex
+++ b/lib/surface/components/form/error_tag.ex
@@ -143,7 +143,9 @@ defmodule Surface.Components.Form.ErrorTag do
                 {Surface.Components.Form.ErrorTag, default_translator: {MyAppWeb.ErrorHelpers, :translate_error}}
               ]
 
-            Original error: #{Exception.message(e)}
+            Given value: #{inspect(value)}
+            
+            Exception: #{Exception.message(e)}
             """,
             __STACKTRACE__
           )
--- a/lib/surface/components/form/error_tag.ex
+++ b/lib/surface/components/form/error_tag.ex
@@ -124,7 +124,25 @@ defmodule Surface.Components.Form.ErrorTag do
     # Because the error messages we show in our forms and APIs
     # are defined inside Ecto, we need to translate them dynamically.
     Enum.reduce(opts, msg, fn {key, value}, acc ->
-      String.replace(acc, "%{#{key}}", to_string(value))
+      try do
+        String.replace(acc, "%{#{key}}", to_string(value))
+      rescue
+        e in ArgumentError ->
+          if e.message =~ "cannot convert the given list to a string" do
+            message = """
+            Hint: In Surface, this error often happens because no `default_translator` has been set
+            for the `ErrorTag` component. You can point the default translator to your application helpers:
+
+              config :surface, :components, [
+                {Surface.Components.Form.ErrorTag, default_translator: {MyAppWeb.ErrorHelpers, :translate_error}}
+              ]
+            """
+
+            reraise %{e | message: e.message <> message}, __STACKTRACE__
+          else
+            reraise e, __STACKTRACE__
+          end
+      end
     end)
   end
 
--- a/lib/surface/components/form/error_tag.ex
+++ b/lib/surface/components/form/error_tag.ex
@@ -21,9 +21,10 @@ defmodule Surface.Components.Form.ErrorTag do
   ]
   ```
 
-  Surface provides a fallback error message if the error cannot be converted to a string. If it
-  happens, Surface will raise a warning to invite you to configure a `default_translator` that
-  handles such a case.
+  > **Note:** If you don't configure a `default_translator`, Surface will try to translate errors using 
+  a built-in message translator which may not cover all types of errors. If the error cannot
+  be translated, a generic `"invalid value"` will be returned and a warning will be emitted,
+  reminding the user to set up a proper `default_translator` that can handle such cases.
 
   There is also a `translator` prop which can be used on a case-by-case basis.
   It overrides the configuration.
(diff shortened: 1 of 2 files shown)


Malian:
It seems that this issue also happens with when `Ecto.Enum` is used.

lib/surface/components/form/error_tag.ex
```diff
@@ -124,7 +124,25 @@ defmodule Surface.Components.Form.ErrorTag do
...
+      rescue
+        e in ArgumentError ->
+          if e.message =~ "cannot convert the given list to a string" do
+            message = """
+            Hint: In Surface, this error often happens because no `default_translator` has been set
+            for the `ErrorTag` component. You can point the default translator to your application helpers:
+
+              config :surface, :components, [
+                {Surface.Components.Form.ErrorTag, default_translator: {MyAppWeb.ErrorHelpers, :translate_error}}
+              ]
```
miguel-s, line 138: Should we also add a [link](https://surface-ui.org/builtincomponents/ErrorTag) to the docs, Elm style? Or is that frowned upon in Elixir?

miguel-s (approved):
Always a fan of giving hintd in error messages!

lib/surface/components/form/error_tag.ex
```diff
@@ -124,7 +124,25 @@ defmodule Surface.Components.Form.ErrorTag do
     # Because the error messages we show in our forms and APIs
     # are defined inside Ecto, we need to translate them dynamically.
     Enum.reduce(opts, msg, fn {key, value}, acc ->
-      String.replace(acc, "%{#{key}}", to_string(value))
+      try do
+        String.replace(acc, "%{#{key}}", to_string(value))
+      rescue
+        e in ArgumentError ->
+          if e.message =~ "cannot convert the given list to a string" do
```
Malian, line 131: Unfortunalty, this condition will not catch the error from #450

> ** (Protocol.UndefinedError) protocol String.Chars not implemented for {:parameterized, Ecto.Enum, %{on_dump: %{B2B: "B2B", EoC: "EoC"}, on_load: %{"B2B" => :B2B, "EoC" => :EoC}, type: :string, values: [:B2B, :EoC]}} of type Tuple

I do not like the idea to have a list of strings that will produce this error, but I don't know how we can solve this in a better way.

davydog187: Could we provide a more general error message instead, then we can append to all `ArgumentError` raised?

Malian: The error type from #450 is not an `ArgumentError` but `Protocol.UndifinedError` :(

msaraiva: What if we instead of catching the error and re-raising it, we test the value, show a fallback message and warn with the hint? 

Something like:

```elixir
    Enum.reduce(opts, msg, fn {key, value}, acc ->
      message = 
        if String.Chars.impl_for(value) do
          to_string(value)
        else
          # TODO: show_warn_with_hint(...)
          "invalid value"
        end
      String.replace(acc, "%{#{key}}", message)
    end)
```

msaraiva (commented):
I left a suggestion of a possible (hopefully) solution.

Additionally, there are a couple things we could/should do to avoid this kind of issue:

1. Add instructions to the "Getting Started" page to set up the default translator, when needed. I believe that as suggested by @Malian a while ago.
2. Roll our own `mix surface.init` task which already sets this whole thing up along with any other custom stuff we might have. Assuming most people will use that task when trying out Surface, this could make a hell of a difference as we'll no longer depend on those manual adjustments by the user. 

We have a working solution for an `init` task [here](https://github.com/lastobelus/sfc_gen_live#sfcgeninit) (`v0.4`). Maybe it's time to bring that to the core, update it to `v0.5` and add any custom stuff we need to avoid this kind of issues.

WDYT?

lib/surface/components/form/error_tag.ex
Malian: It won't work for the `unsafe_unique` validator as it returns a list and `List` implements the `String.Chars` protocol, so the condition will be valid and the `to_string` function will return the plain old `ArgumentError` :cry: 

https://github.com/elixir-lang/elixir/blob/7e97d08b03129f367a6578ebacb17861b6726e4f/lib/elixir/lib/list.ex#L967-L996

msaraiva: @Malian in this case, we have a few options:

1. Catch any error raised (not only ArgumentError), return the generic "invalid value" message as suggested and emit a warning. Something like:

    ```
    the fallback message translator for the `ErrorTag` component cannot handle the given message.

    Hint: If you're using `GetText`, set up the default translator to your application helpers:

        {Surface.Components.Form.ErrorTag, default_translator: {MyAppWeb.ErrorHelpers, :translate_error}}
      ]

    Original error: #{Exception.message(exception)
    ```

2. Do the above if an error occurs, but additionally emit another warning earlier even if we can handle the message but the default_translator is not set. This way users will knows upfront that they need to either set up `GetText` or roll their own translator, otherwise the fallback may not handle the message properly. Something like:

    ```
    the default message translator for `ErrorTag` has not been set. Using a fallback translator.

    Pay attention that the fallback translator may not be able to translate all given messages.

    Hint: If you're using `GetText`, set up the default translator to your application helpers with:

        {Surface.Components.Form.ErrorTag, default_translator: {MyAppWeb.ErrorHelpers, :translate_error}}
      ]

    If you want to suppress this warning and keep using the fallback translator, please explicitly set it 
    as the default translator  with: 

        {Surface.Components.Form.ErrorTag, default_translator: {Surface.Components.Form.ErrorTag, :translate_error}}
      ]
    ```

msaraiva: Actually, I changed my mind and vote for `#1`. I believe if the docs already mention that the fallback translator may fail, we don't have to always warn, only when it fails.

Malian: I confess that I also prefer the first solution. I would not like to emit a warning if the `default_translator` has not been set for the `ErrorTag` component, as some user would not use the built-in form components.

I'll update the doc and the PR according to your proposition.

lib/surface/components/form/error_tag.ex
```diff
@@ -124,7 +128,27 @@ defmodule Surface.Components.Form.ErrorTag do
...
     # are defined inside Ecto, we need to translate them dynamically.
     Enum.reduce(opts, msg, fn {key, value}, acc ->
-      String.replace(acc, "%{#{key}}", to_string(value))
+      try do
+        String.replace(acc, "%{#{key}}", to_string(value))
+      rescue
+        e ->
+          IO.warn(
+            """
+            the fallback message translator for the `ErrorTag` component cannot handle the given message.
```
msaraiva, line 137:
```suggestion
            the fallback message translator for the `ErrorTag` component cannot handle the given value.
```

lib/surface/components/form/error_tag.ex
```diff
@@ -124,7 +128,27 @@ defmodule Surface.Components.Form.ErrorTag do
...
+
+            Hint: you can set up the `default_translator` to route all errors to your application helpers:
+
+              config :surface, :components, [
+                {Surface.Components.Form.ErrorTag, default_translator: {MyAppWeb.ErrorHelpers, :translate_error}}
+              ]
+
+            Original error: #{Exception.message(e)}
```
msaraiva, line 145: Let's also print the `value`. It might be helpful.

```suggestion
            Given value: #{inspect(value)}
            
            Exception: #{Exception.message(e)}
```

msaraiva (approved):
@Malian looks good to me. I just left a couple of suggestions for the messages. I believe you can merge it after updating them.

msaraiva:
❤️

PR #453, opened 2021-07-07
Use component/3 macro in Surface.Component

msaraiva:


test/surface/component_test.exs
```diff
@@ -105,18 +105,14 @@ defmodule Surface.ComponentTest do
...
 
-    @impl true
-    def update(assigns, socket) do
-      socket =
-        socket
-        |> assign(assigns)
-        |> assign(:id_copy, assigns.id)
-
-      {:ok, socket}
+    def update(assigns) do
```
Malian, line 108:
```suggestion
    defp update(assigns) do
```
I suggest to make this function private in order to avoid that users believe this is part of the `Surface` api neither `PhoenixLiveView` api when looking at the tests.

msaraiva:
@lnr0626, @Malian and @miguel-s I think this is ready to be merged. I'll implement support for LV's function components using `<.func>` in a separate PR.

test/surface/integrations/component_change_tracking_test.exs
```diff
@@ -0,0 +1,78 @@
...
+
+  test "change tracking" do
+    # Initial values
+
+    {:ok, view, html} = live_isolated(build_conn(), View)
+    result_1 = parse_result(html)
+    assert result_1["comp_1"].value == 0
+    assert result_1["comp_2"].value == 0
+
+    # Don't rerender components if their props haven't change
```
miguel-s, line 49: Mini nitpick: typo in "change" -> "changed"

lib/surface/compiler/eex_engine.ex
```diff
@@ -318,42 +288,29 @@ defmodule Surface.Compiler.EExEngine do
...
-           context_expr,
-           props_expr,
-           dynamic_props_expr,
-           slot_props,
-           slot_meta,
-           module,
-           node_alias,
-           do_block
-         ) do
+    else
```
miguel-s, line 296: Probably a nitpick: I believe it would help to be more explicit in this `else`. Either change it for `if component_type == Surface.Component` or use a case statement.

miguel-s (approved):
Lgtm! Added 2 nitpicks, feel free to ignore if too pedantic.

Excited for components in deadviews!

This is a breaking change for `Surface.Components` that currently use `mount` or `update`, right?

lnr0626:
Lgtm!

msaraiva:
> This is a breaking change for `Surface.Components` that currently use `mount` or `update`, right?

@miguel-s yes, it is. You must replace them with `assign`, `assing_new` or `update` inside `render/1`.

lib/surface/compiler/eex_engine.ex
msaraiva: @miguel-s that conditional will probably get a third branch with the functional component case. I'll change that in the other PR. Thanks!

PR #454, opened 2021-07-08
Handle doctype as text

msaraiva:


msaraiva:
@lnr0626 decided to merge this as it's mostly a verbatim copy of heex's implementation

PR #458, opened 2021-07-13
All Input components should inherit :default_class from Form.Input

thorsten-de:
Only some of the Input components (e.g. `TextInput`) make use of `get_default_class()` to inherit the default_class configured for `Form.Input`.
This mechanism should be used consequently on every component that make use of Form.Input. It may be discussed if `Checkbox` and `FileInpu`t should be exceptions, because they often require different markup anyway.

### DIFF
--- a/lib/surface/components/form/checkbox.ex
+++ b/lib/surface/components/form/checkbox.ex
@@ -39,7 +39,7 @@ defmodule Surface.Components.Form.Checkbox do
         :value
       ])
 
-    attr_opts = props_to_attr_opts(assigns, class: get_config(:default_class))
+    attr_opts = props_to_attr_opts(assigns, class: get_default_class())
     event_attrs = events_to_attrs(assigns)
 
     ~F"""
--- a/lib/surface/components/form/color_input.ex
+++ b/lib/surface/components/form/color_input.ex
@@ -24,7 +24,7 @@ defmodule Surface.Components.Form.ColorInput do
 
   def render(assigns) do
     helper_opts = props_to_opts(assigns)
-    attr_opts = props_to_attr_opts(assigns, [:value, class: get_config(:default_class)])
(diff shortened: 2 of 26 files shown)


Malian:
Hi @thorsten-de and thanks for your first Surface contribution!

I agree we should be consistent across the `Input` components.

Could you please add tests for components that use `get_default_class()`? Something like this should do the trick:
```elixir
  test "component inherits :default_class from Form.Input" do
    using_config Surface.Components.Form.Input, default_class: "inherited_default_class" do
      html =
        render_surface do
          ~F"""
          <ColorInput/>
          """
        end

      assert html =~ ~r/class="inherited_default_class"/
    end
  end
```
We should also decide what we do if both `Form.Input` and `ColorInput` default class has been set. My point of view is that `default_class` from the "child" input overrides the `default_class` from `Form.Input`. Another solution could be to merge classes, but I do not like this solution without being able to explain why...
```elixir
  test ":default_class config overrides inherited :default_class from Form.Input" do
    using_config ColorInput, default_class: "default_class" do
      using_config Surface.Components.Form.Input, default_class: "inherited_default_class" do
        html =
          render_surface do
            ~F"""
            <ColorInput/>
            """
          end

      end
    end
  end
```

@msaraiva Actually, the `using_config` macro does not allow to define config for several components. Does it make sense to refactoring it to handle this case or is it overengineering?

thorsten-de:
Hi @Malian, 

I added the tests you proposed to every component that should use `get_default_class()`.

As it is, a `default_class` defined for a specialized input overrides the one defined for `Form.Input`, which feels fine for me. I just didn't want to define `default_class: "input"` for all input types to use Bulma properly.

Btw,  writing the test felt a bit copy-paste-ish for all the input types, so I initially thought about putting these in a loop for testing. But I found no obvious way to dynamically define the tag inside the template.

Malian:
❤️

PR #459, opened 2021-07-15
Migrate to Phoenix.HTML.Tag.attributes_escape/1

miguel-s:
Use the newly introduced [`Phoenix.HTML.Tag.attributes_escape/1`](https://github.com/phoenixframework/phoenix_html/pull/331) instead of handrolling our own solution.

@msaraiva @lnr0626 @Malian I tested the new approach with the `Link` and `Link.Button` components, if you think this is the way to go I'll continue cleaning up the rest. The main changes are in [type_handler.ex](https://github.com/surface-ui/surface/compare/master...miguel-s:ms-tag-attributes-escape?expand=1#diff-266fdbcf268ed223050000921d1b6946b9a028b4f7e17c64fee8d3e0a41a2991R183) and [eex_engine.ex](https://github.com/surface-ui/surface/compare/master...miguel-s:ms-tag-attributes-escape?expand=1#diff-96328b37fbe00256a44961d185c56d2b6812ccf64fd5108cbc8cf9b4a1b2f2ecR76). wdyt?

### DIFF
--- a/lib/surface/components/link.ex
+++ b/lib/surface/components/link.ex
@@ -60,32 +60,27 @@ defmodule Surface.Components.Link do
   slot default
 
   def render(assigns) do
-    valid_label!(assigns)
-    to = valid_destination!(assigns.to, "<Link />")
-
-    ~F"""
-    <a id={@id} class={@class} href={to} :attrs={to_attrs(assigns, to)}><#slot>{@label}</#slot></a>
-    """
-  end
-
-  defp valid_label!(assigns) do
     unless assigns[:default] || assigns[:label] || Keyword.get(assigns.opts, :label) do
       raise ArgumentError, "<Link /> requires a label prop or contents in the default slot"
     end
+
+    to = valid_destination!(assigns.to, "<Link />")
+    events = events_to_opts(assigns)
+    opts = link_method(assigns.method, to, assigns.opts)
+    assigns = assign(assigns, to: to, opts: events ++ opts)
+
+    ~F"""
+    <a id={@id} class={@class} href={@to} :attrs={@opts}><#slot>{@label}</#slot></a>
+    """
   end
 
-  defp apply_method(to, method, opts) do
+  defp link_method(method, to, opts) do
     if method == :get do
       skip_csrf(opts)
     else
+      {data, opts} = Keyword.pop(opts, :data, [])
       {csrf_data, opts} = csrf_data(to, opts)
-      opts = Keyword.put_new(opts, :rel, "nofollow")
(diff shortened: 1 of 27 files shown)


lib/surface/components/link.ex
```diff
@@ -60,32 +60,27 @@ defmodule Surface.Components.Link do
...
-  defp apply_method(to, method, opts) do
+  defp link_method(method, to, opts) do
     if method == :get do
       skip_csrf(opts)
     else
+      {data, opts} = Keyword.pop(opts, :data, [])
       {csrf_data, opts} = csrf_data(to, opts)
-      opts = Keyword.put_new(opts, :rel, "nofollow")
-      [data: [method: method, to: to] ++ csrf_data] ++ opts
+      [data: data ++ csrf_data ++ [method: method, to: to], rel: "nofollow"] ++ opts
```
Malian, line 83: If data contains the `method` or the `to` keys, the `++` operator will create a keyword list with duplicated keys and it will result in invalid HTML. `Keyword.merge` will replace the previous values and solve this "potential" issue, but I admit it is less readable. WDYT?

miguel-s: It turns out it's safer AND more readable 👍

miguel-s: Just wanted to add that the `:attrs` directive has a safety net and only uses the first occurence if there's duplicate keys. But still good to clean it up at the component level :)

lib/surface/components/link.ex
```diff
@@ -60,32 +60,33 @@ defmodule Surface.Components.Link do
...
   def render(assigns) do
-    valid_label!(assigns)
+    unless assigns[:default] || assigns[:label] || Keyword.get(assigns.opts, :label) do
+      raise ArgumentError, "<Link /> requires a label prop or contents in the default slot"
+    end
+
     to = valid_destination!(assigns.to, "<Link />")
+    events = events_to_opts(assigns)
+    opts = link_method(assigns.method, to, assigns.opts)
+    assigns = assign(assigns, to: to, opts: events ++ opts)
```
msaraiva, line 70: 👍

msaraiva (commented):
Looks good to me!

lib/surface/components/link.ex
msaraiva: We need to remember to use `assign/2..3` and friends in all our components. No need to do that in this PR though. Only the ones you already have to touch.

miguel-s: I'll open a different PR to migrate all remaining components to the new `assign/2..3` way.

msaraiva (approved):
❤️ ❤️ ❤️

PR #461, opened 2021-07-17
Improve change tracking

msaraiva:
This PR fix change tracking when not using contexts and adds a bunch of optimizations when using contexts.

### DIFF
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -59,7 +59,6 @@ defmodule Surface.API do
       # Any caller component can hold other components with slots
       Module.register_attribute(__MODULE__, :assigned_slots_by_parent, accumulate: false)
 
-      Module.put_attribute(__MODULE__, :use_context?, false)
       Module.put_attribute(__MODULE__, :changes_context?, false)
       Module.put_attribute(__MODULE__, :gets_context?, false)
 
--- a/lib/surface/api.ex
+++ b/lib/surface/api.ex
@@ -60,6 +60,8 @@ defmodule Surface.API do
       Module.register_attribute(__MODULE__, :assigned_slots_by_parent, accumulate: false)
 
       Module.put_attribute(__MODULE__, :use_context?, false)
+      Module.put_attribute(__MODULE__, :changes_context?, false)
+      Module.put_attribute(__MODULE__, :gets_context?, false)
 
       for func <- unquote(include) do
         Module.register_attribute(__MODULE__, func, accumulate: true)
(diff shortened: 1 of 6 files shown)


paulstatezny:
@msaraiva I tested this out in the context where I first discovered the issue.

Previously: About **45kb** was sent per event.
Using this PR: About **17kb** is sent per event.

It appears that this PR has a huge impact! Awesome. 🙂

lib/surface/api.ex
```diff
@@ -60,6 +60,8 @@ defmodule Surface.API do
       Module.register_attribute(__MODULE__, :assigned_slots_by_parent, accumulate: false)
 
       Module.put_attribute(__MODULE__, :use_context?, false)
```
lnr0626, line 62: are we still using this attribute?

lib/surface/compiler/eex_engine.ex
```diff
@@ -93,7 +95,7 @@ defmodule Surface.Compiler.EExEngine do
       handle_nested_block(children, buffer, %{
         state
         | depth: state.depth + 1,
-          context: [:for | state.context]
+          scope: [:for | state.scope]
```
lnr0626, line 98: 👍

lib/surface/compiler/eex_engine.ex
```diff
@@ -477,9 +506,15 @@ defmodule Surface.Compiler.EExEngine do
...
+      state
+      | depth: state.depth + 1,
+        context_vars: %{state.context_vars | count: state.context_vars.count + 1}
+    }
+
     [
       {add_default_bindings(component, name, let), Keyword.merge(default_props, props),
-       handle_nested_block(template, buffer, %{state | depth: state.depth + 1})}
+       handle_nested_block(template, buffer, state)}
       | handle_templates(component, tail, buffer, state)
```
lnr0626, line 518: i think this will cause issues when multiple templates are provided as we're now passing the updated state to both `handle_nested_block` and the call to `handle_templates` for the next template in the list

we could either rename the updated state so it's only passed to `handle_nested_block`, or pull the update logic into a helper method and pass the result directly into `handle_nested_block`

lib/surface/compiler/eex_engine.ex
```diff
@@ -434,9 +458,14 @@ defmodule Surface.Compiler.EExEngine do
...
+      state
+      | depth: state.depth + 1,
+        context_vars: %{state.context_vars | count: state.context_vars.count + 1}
+    }
+
     [
-      {add_default_bindings(component, name, let), [],
-       handle_nested_block(children, buffer, %{state | depth: state.depth + 1})}
+      {add_default_bindings(component, name, let), [], handle_nested_block(children, buffer, state)}
       | handle_templates(component, tail, buffer, state)
```
lnr0626, line 469: this has the same potential issue as below

lnr0626 (commented):
one potential issue, but other than that lgtm!

msaraiva:
@lnr0626 good catch! That'd certainly be a problem. I believe it's fine now.

lnr0626 (approved):
Lgtm, nice work!

PR #462, opened 2021-07-18
Better error message for misplaced closing delimiter

davydog187:
Improves the error message for space in a closing tag like `<div / >`

I also feel that `<div = >` and `<div ="value no name">` could have better error messages too. What do you think?

### DIFF
--- a/lib/surface/compiler/tokenizer.ex
+++ b/lib/surface/compiler/tokenizer.ex
@@ -530,6 +530,9 @@ defmodule Surface.Compiler.Tokenizer do
   end
 
   ## handle_attr_name
+  defp handle_attr_name(<<"/", _rest::binary>>, _column, []) do
+    {:error, "unexpected closing tag delimiter `/`"}
+  end
 
   defp handle_attr_name(<<c::utf8, _rest::binary>>, _column, []) when c in @name_stop_chars do
     {:error, "expected attribute name"}
--- a/lib/surface/compiler/tokenizer.ex
+++ b/lib/surface/compiler/tokenizer.ex
@@ -534,8 +534,8 @@ defmodule Surface.Compiler.Tokenizer do
     {:error, "unexpected closing tag delimiter `/`"}
   end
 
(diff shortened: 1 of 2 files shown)


msaraiva (approved):
It's always nice to provide better messages! ❤️

> I also feel that <div = > and <div ="value no name"> could have better error messages too. What do you think?

Not sure if they are as common as the one you've addressed but maybe we could add more info to the fallback message, like:

```
expected attribute name, got: "="
```

davydog187:
Thanks @msaraiva, I pushed up a better message for those as well :)

lib/surface/compiler/tokenizer.ex
```diff
@@ -530,9 +530,12 @@ defmodule Surface.Compiler.Tokenizer do
...
 
   ## handle_attr_name
+  defp handle_attr_name(<<"/", _rest::binary>>, _column, []) do
+    {:error, "unexpected closing tag delimiter `/`"}
+  end
 
-  defp handle_attr_name(<<c::utf8, _rest::binary>>, _column, []) when c in @name_stop_chars do
-    {:error, "expected attribute name"}
+  defp handle_attr_name(<<c::utf8, _rest::binary>> = text, _column, []) when c in @name_stop_chars do
+    {:error, "expected attribute name, got: `#{text}`"}
```
msaraiva, line 538: I wonder if  it wouldn't be better to use just:

```
{:error, "expected attribute name, got: `#{c}`"}
```

After all `text` will be the rest of the whole template which might pollute the error message a lot.

davydog187: We definitely could, but it would always be a single character, in this case "=". I'm happy to change to that.

Btw I'll also need to convert from the ascii code to a string

davydog187: walked it back to your suggestion @msaraiva

msaraiva:
@davydog187 thanks! ❤️

Issue #463, opened 2021-07-19
Compile error if slot index property is listed before name


totaltrash:
## Describe the bug

When binding slots to generators, if i configure the slot like so:

```
<#slot index={index} name="cols" :args={item: item} />
```

i get 

```
** (CompileError) lib/i_fix/components/table.ex:38: invalid attribute `index` for <#slot>.
Slots only accept `name`, `:args`, `:if` and `:for`.
```

but, if i reorder the slot props, so the index comes after the name:

```
<#slot name="cols" index={index} :args={item: item} />
```

everything is cool

## How to reproduce it

as above

## The behavior you expected

to either compile ok, or to warn about the order of the slot props if there is a requirement

## Your Environment
Surface: v0.5.1
LiveView: v0.15.7
Elixir: v1.12.1

miguel-s:
In the compiler we force the `name` attribute to always be first, else we raise a compile error: https://github.com/surface-ui/surface/blob/2f951acf277126ef2a33f09f1050898ce2c384d5/lib/surface/compiler.ex#L1195
Is there a reason for this @msaraiva ? Else this should be an easy fix :)

msaraiva:
@miguel-s the implementation is completely broken. This function was supposed to be recursive, traversing all attributes and it should consider `name` and `index` as valid attributes, not just `name`. 

Jesus, where was my mind I when I wrote that 🤦‍♂️

PR #464, opened 2021-07-19
Fix slot attribute validation

msaraiva:
Fixes https://github.com/surface-ui/surface/issues/463.

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -47,6 +47,8 @@ defmodule Surface.Compiler do
     Surface.Directive.For
   ]
 
+  @valid_slot_props ["name", "index"]
+
   @directive_prefixes [":", "s-"]
 
   @void_elements [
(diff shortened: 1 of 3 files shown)


lib/surface/compiler.ex
```diff
@@ -1192,7 +1192,11 @@ defmodule Surface.Compiler do
     end
   end
 
-  defp validate_slot_attrs!([{name, _, %{file: file, line: line}} | _]) when name != "name" do
+  defp validate_slot_attrs!(attrs) do
+    Enum.each(attrs, &validate_slot_attr!/1)
+  end
+
+  defp validate_slot_attr!({name, _, %{file: file, line: line}}) when name not in ["name", "index"] do
```
miguel-s, line 1199: I think it would read more naturally if we avoid the negation and switch it around:
- when name in ["name", "index"] -> :ok
- rest -> raise compile error

PR #466, opened 2021-07-22
Cleanup usage of `{{ ... }}`

davydog187:
Its a shame that we can't doctest components, otherwise we would have caught all these. Maybe I can pair with @wojtekmach on a creative solution ;)

### DIFF
--- a/lib/surface.ex
+++ b/lib/surface.ex
@@ -85,7 +85,7 @@ defmodule Surface do
 
         def render(assigns) do
           ~F"\""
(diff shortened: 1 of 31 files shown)


msaraiva:
@davydog187 thanks! ❤️

msaraiva:
It would be nice to have those changes applied to the branch `v0.5` too so we could have the website updated. However, since Phoenix `v1.6` is right around the corner, I think we can just wait :)

PR #469, opened 2021-07-29
docs: Adds changelog and license as extra pages

Blond11516:
Adds the changelog and license to ex_doc's `extras` config to make them accessible from the official hexdocs.pm page.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -59,6 +59,10 @@ defmodule Surface.MixProject do
         Surface.Compiler,
         Surface.Components,
         Surface.Directive
+      ],
+      extras: [
+        "CHANGELOG.md",
+        "LICENSE.md"
       ]
     ]
   end


Malian:
Hi @Blond11516 ! Thanks for using Surface.

I never have seen a project that adds changelog and license to the `extras`.

We have defined the license with the `license` keyword:
https://github.com/surface-ui/surface/blob/514bbbda4575378d45e3788a3f0391e8d0581a60/mix.exs#L68-L69

We could add the `changelog` to the `links` section. 

What is the purpose of adding it to the documentation? Could you explain us more why you would like to have these content accessible from Hexdoc?

Blond11516:
Hi @Malian!

The goal of this change is mostly to make the changelog as easily accessible as possible. Adding it to the links section would also be an acceptable solution, but I feel like having it right in the docs is more intuitive (both are not mutually exclusive though). That's the first place I usually look for a changelog, be it for Elixir packages or any other language (this is obviously subjective, I have no idea how common this is).

Examples of projects that do this include:
- Elixir: https://github.com/elixir-lang/elixir/blob/58ca6524f88bfe7e963deafe0e7d6aa07a568003/lib/elixir/docs.exs#L6
- ExDoc: https://github.com/elixir-lang/ex_doc/blob/38c672f061a2303a312caa806ad1ad5cb1d446da/mix.exs#L89

As for the license, I added it more as a "might as well do it while I'm here" kind of thing. I don't really have a strong use case for it, but felt it makes sense to have it in the docs too.

Malian:
@Blond11516 Thanks for the clear explanation and the examples. You convinced me.

Just one last thing. Could you please run `mix format` to fix the CI? :wink:

Blond11516:
@Malian Done!

Malian:
❤️

PR #471, opened 2021-07-30
Remove automatic punctuation from generated docs

oskarkook:
Current implementation can have some unintended effects, e.g. I have an example section for a prop, which has a dangly dot after it:
image

I also prefer to end sentences with proper punctuation myself, which leaves another dangly dot:
image

I think it would be sensible if punctuation were left to the person documenting the module/properties.


Malian:
❤️

Issue #24, opened 2020-01-29
Rendering dynamic component


andreasknoepfle:
Hi 👋 

Is there some way of rendering a component dynamically? Like if I had an array with a `Button` and a `Link` component and I want to render them in a for loop:

```
def render(assigns) 
  content = [Button, Link]
  ~H"""
  <div :for={{component <- content}}>
     ???
  </div>
  """
end
```

I tried already some things like rendering with `live_component()` like : 
```
{{ live_component, @socket, component }}
```
That works on an older state, but errors with the current master: `BadMapError: expected a map, got: nil` in `Surface.ContentHandler.init_contents/1`.

Also trying to interpolate the component somehow does not work, but would probably be a nice syntax: 
```
<{{component}} />
```

Do you have any plans for this feature already (or am I missing something). 
Since this is easily possible in plain LiveView it would sad if we couldn't also do this with surface. 
I also imagine having the interpolation on a html element level like in react would be handy: 
```
<h{{level}}>A headline</h{{level}}>
```

Also I would like to thank you for creating this project since I am already a huge fan of the idea 💚.

I would also volunteer to look into this and maybe create a PR if you give me some directions :) 

Cheers
Andi

msaraiva:
Hi @andreasknoepfle!

> Is there some way of rendering a component dynamically?

Not yet, but this is definitely on the roadmap and there's no way we'll ship the first version without it.

The plan is to have a built-in `<Dynamic>` component that can handle those cases. Something that you could use like this:

```elixir
def render(assigns) do
  props = %{label: "Ok", class: "button"}
  component = Button

  ~H""" 
  <Dynamic component={{ component }} props={{ props }}/>
  """
end
```
A first naive implementation that would probably work for the most simple cases (like stateless components without nested children) would look like:

```elixir
  defmodule Dynamic do
    use Surface.Component

    property component, :module, required: true
    property props, :map, default: %{}

    def render(assigns) do
      props =
        assigns
        |> Map.get(:props)
        |> Map.merge(%{__surface__: %{groups: %{__default__: %{binding: false, size: 0}}}}) # Don't worry about this for now :)

      ~H"""
      {{ live_component(@socket, @component, props) }}
      """
    end
  end
```

I'm currently working on the [new event API](https://github.com/msaraiva/surface/issues/20) due to some breaking changes in LiveView so, unfortunately, I wasn't able to validate the above solution but feel free to give it a try if you want and let me know if it works :)

Cheers.

[...]

PR #473, opened 2021-08-03
Support function components

msaraiva:
Also closes https://github.com/surface-ui/surface/issues/343 and https://github.com/surface-ui/surface/issues/24.


lib/surface/compiler.ex
```diff
@@ -1223,4 +1294,17 @@ defmodule Surface.Compiler do
 
     IOHelper.compile_error(message, file, line)
   end
+
+  defp handle_convert_node_to_ast_error(name, error, meta) do
```
lnr0626, line 1298: ❤️

lib/surface/compiler/eex_engine.ex
```diff
@@ -216,124 +216,188 @@ defmodule Surface.Compiler.EExEngine do
...
 
-    dynamic_props_expr = handle_dynamic_props(dynamic_props)
+    {props_expr, dynamic_props_expr} = build_props_expressions(nil, component)
+    {context_expr, context_var, state} = process_context(module, fun, props, meta.caller, state)
+    {do_block, slot_meta, slot_props} = collect_slot_meta(component, buffer, state, context_var)
 
-    caller_component_type = Module.get_attribute(meta.caller.module, :component_type)
+    quote generated: true do
+      component(
+        &(unquote(Macro.var(fun, __MODULE__)) / 1),
```
lnr0626, line 266: The spacing here looks odd, but I assume that's not an actual issue?

lib/surface/components/dynamic_component.ex
```diff
@@ -0,0 +1,43 @@
+defmodule Surface.Components.Component do
```
lnr0626, line 1: Should we add a 'props' prop to allow passing a map of props? My thinking is that otherwise users would get warnings about unknown props for this component

lnr0626 (approved):
This looks great, good work!

I had a few small comments/questions, but no blockers

test/surface/components/dynamic_component_test.exs
```diff
@@ -0,0 +1,298 @@
+defmodule Surface.Components.ComponentComponentTest do
```
Malian, line 1:
```suggestion
defmodule Surface.Components.DynamicComponentTest do
```

lib/surface/compiler/eex_engine.ex
msaraiva: Yeah, that's not an issue. I find that ugly too but that's the way the formatter handles it so there's not much we can do about it, I guess :)

lib/surface/components/dynamic_component.ex
msaraiva: They shouldn't get any warning but they can also use `{...}`. However, I'm not sure yet if that API is the best one. I'll create some tests for those cases and think how we can improve it.

test/surface/components/dynamic_component_test.exs
msaraiva: @Malian good catch! I was also thinking about changing the namespace from `Surface.Components.Component` to `Surface.Components.Dynamic.Component` as we still gonna have to implement a `Surface.Components.Dynamic.LiveComponent` in the near future. Since those built-in components are automatically aliased, I don't mind the larger name. WDYT?

lib/surface/components/dynamic_component.ex
miguel-s, line 1: Going by the file name:

```suggestion
defmodule Surface.Components.DynamicComponent do
```

miguel-s (approved):
Amazing work 👏 

Can't add it to the review, but we might be able to remove the `import Phoenix.HTML.Form` from the `Form` component.

lib/surface/components/dynamic_component.ex
miguel-s: But in the tests it seems you named it `Component` on purpose, it does read nicer 🤔

Issue #482, opened 2021-08-21
Custom hooks extension support


yusufyildirim:
Hi everyone!

I was playing around with esbuild and encountered a slight problem with its Surface compatibility. I use Preact on my project to render some react components and I trigger that rendering inside my hooks. The problem is, esbuild requires `.jsx` extension to support JSX and Surface strictly looks for `.hooks.js` extension. I think adding a configuration option for hooks extension would be great to solve this case, just like you [did for output dir](https://surface-ui.org/js_interop#choosing-a-different-output-dir).

yusufyildirim:
PR Ready! https://github.com/surface-ui/surface/pull/483

Issue #444, opened 2021-06-25
Feature: Add Typescript Support


chevinbrown:
We don't currently support Typescript. I likely won't have time to get to this, unfortunately, but want to tag the issue to kick off discussion.

codechips:
Interesting! What current (or future) problem will it solve?

chevinbrown:
@codechips TS is fairly pervasive in the js community...and for some projects, nonstarter if the project isn't in TS. I would think we would want to support that part of the JS community.

codechips:
Hear hear .. I am curious where JS in Surface is used. Thought it was hidden (abstracted away) by the LiveView since Surface is more or less a compiler. Care to point me in the right direction @chevinbrown? Where do I start? I know my way around TS so I might be able to help :)

chevinbrown:
@codechips nice! Yeah, it'd only be for hooks AFAIK. I'm actually not familiar with how it does/should work. I'd start here: https://github.com/surface-ui/surface/blob/master/lib/mix/tasks/compile/surface.ex#L9

justin-m-morgan:
One solution would be to write a types declaration and publish it to the @types NPM organization. 

[Typescript - Publishing](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)

I'd be happy to undertake this project.

simonmcconnell:
@justin-m-morgan I think it is more around allowing people to write their hooks in typescript.  PR #483 has some info around it.

Not sure if the clojurescript usage in https://github.com/code-shoily/covid19 would be of any inspiration?

onelesd:
Phoenix 1.6 uses esbuild which supports typescript out of the box. You rename the *.js files to *.ts and it Just Works. See my PR #574.

onelesd:
Phoenix 1.6 uses esbuild which supports typescript out of the box. You rename the *.js files to *.ts and it Just Works. See my PR #574.

VictorGaiva:
I could see some gain in writing hook type definitions, so that writing them become less painful, but this work would be done on the Phoenix Live View project first.

I started a refactor of the JS on the Phoenix Live View project a while ago, seen [here](https://github.com/VictorGaiva/phoenix_live_view), strongly typing stuff all around. Any help moving forward is welcome.

chevinbrown:
@onelesd very nice! Hopefully that makes it upstream and we can close this out!

PR #483, opened 2021-08-21
Custom hooks extension support

yusufyildirim:
This PR resolves issue #482.

* Implemented the feature as simple as possible, guide me if there is a better way 😄 
* Added a basic test to make sure it works.
* I should probably add this to documentation too but I can deal with that later 😄 

I hope it helps!

### DIFF
--- a/lib/mix/tasks/compile/surface.ex
+++ b/lib/mix/tasks/compile/surface.ex
@@ -7,7 +7,7 @@ defmodule Mix.Tasks.Compile.Surface do
   @recursive true
 
   @default_hooks_output_dir "assets/js/_hooks"
-  @hooks_extension ".hooks.js"
+  @default_hooks_extension ".hooks.js"
 
   @doc false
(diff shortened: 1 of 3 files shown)


msaraiva:
Hi @yusufyildirim. Thanks for the PR!

Since we already have https://github.com/surface-ui/surface/issues/444, maybe we should consider a more generic solution that could work not only with either `.jsx` or `.js` but with all of them, including `.ts` files. Something like `hooks_extensions: [".hooks.js", ".hooks.ts", ".hooks.jsx"]`.

WDYT?

yusufyildirim:
Hi @msaraiva!

Yeah, that was my first approach but I acted lazily and developed the simplest solution 😂

That requires a bit more work. We should do that. Let me check what I can do

Malian:
@yusufyildirim Sorry for the late reply. Is this RP ready to be reviewed following the latest comments?

PR #490, opened 2021-09-17
[#481] Disable build timestamp, replace header

### LINKED ISSUE #481: Can we disable build time timestamp?
I would like to add a config option to Surface that would disable adding "Build time: " line. The code in question is here:

https://github.com/surface-ui/surface/blob/6164224a7b156dd41327db7142af7eae182e86bb/lib/mix/tasks/compile/surface.ex#L176

The reasoning is that it annoys me when switching branches and is being added to commits by developers by mistake, even if nothing changes, resulting in merge conflicts that have to be resolved.

I can provide a PR.

That is if you want to keep this. I am not sure this timestamp is needed at all - the file on the filesystem can be also queried for modification date, so maybe we don't need it at all and can just rely on that? If that's preferred I'd remove the line completely.

hubertlepicki:
This change removes the build timestamp from headers of the generated
files, and replaces it with a helpful message pointing to the original
source files that should be modified instead of the current file.

This change also improves the tests by generating the hook files in the
tests themselves, making sure their content changes and not just rely on
the timestamps / build times.

Closes #481


msaraiva:
@hubertlepicki thanks! ❤️

PR #492, opened 2021-09-20
Allow Inputs component inside the Field component

Malian:
Currently, the API does not allow to use the `Inputs` without explicitly mention the `for` property inside the `Field` component. Moreover, the `for` property is not mandatory so it results with a useless error when it is used like this:

```JSX
<Form for={@changeset}>
  <Field name={:settings}>
    <Inputs>
      <TextInput field={:allow_javascript}>
    </Inputs>
  </Field>
</Form>
```
Error:
> could not generate inputs for nil from MyAppWeb.MyEntity. Check the field exists and it is one of embeds_one, embeds_many, has_one, has_many, belongs_to or many_to_many

I believe it is counterintuitive to not allow the `Inputs` component to retrieve the field from the `Field` context, as we allow it for all the other types of input. This PR fixes it.

Alternative:

- Make the `for` property mandatory with `required: true` as the phoenix helper does not allow the field argument to be `nil`

Drawback:
If the user forgets to define the `field` for an input inside the `Inputs`, the component will try to use the `field` from the `Field` context.
```jsx
<Form for={:parent}>
  <Field name={:children}>
    <Inputs>
      <TextInput /> <-- It will use :children for the field property
    </Inputs>
  </Field>
</Form>
``` 

I am not sure how to solve this. I was thinking about "overriding" the field in the context to set it to `nil` inside the `Inputs` context but I am not sure it will work. @msaraiva @miguel-s do you have an idea?

### DIFF
--- a/test/surface/components/form/inputs_test.exs
+++ b/test/surface/components/form/inputs_test.exs
@@ -134,8 +134,8 @@ defmodule Surface.Components.Form.InputsTest do
       end
 
     assert html =~ """
-           <form action="#" method="post">\
-           <input name="_csrf_token" type="hidden" value="test">
+           <form action="#" method="post">
+               <input name="_csrf_token" type="hidden" value="test">
              <div>
              <input id="parent_children_name" name="parent[children][name]" type="text">
              <input id="parent_children_email" name="parent[children][email]" type="text">


Malian:
Finally, I am wondering if we shouldn't warn at runtime/compile time if an `Input` is used without a `field` from the context or explicitly defined.

Malian:
I do not understand why the test fail. Locally, I do not have these issues. I tested with OTP 23/25 and Elixir 1.11/1.12.

msaraiva (approved):
> I do not understand why the test fail

I can't understand that either. I just pulled the branch and run the tests locally. All passed!

@miguel-s the changes look good to me. Can you see any potential issue? If not, let's merge it and see we keep getting this error.

miguel-s (approved):
lgtm! I agree with @Malian that we should keep the api consistent, I beieve in this case it was just an oversight. 

I checked the output of the failed test and it seems the cause is a missing newline:
```elixir
left:  "<form action=\"#\" method=\"post\">\n    <input ...
right: "<form action=\"#\" method=\"post\"><input ..."
```

miguel-s:
> Finally, I am wondering if we shouldn't warn at runtime/compile time if an Input is used without a field from the context or explicitly defined.

Not sure I follow, could you elaborate on this?

Malian:
> > Finally, I am wondering if we shouldn't warn at runtime/compile time if an Input is used without a field from the context or explicitly defined.
> 
> Not sure I follow, could you elaborate on this?

@miguel-s  Today, you can do this without a warning
```
<TextInput />
```

According to the underlying Phoenix helper, https://github.com/phoenixframework/phoenix_html/blob/v3.0.3/lib/phoenix_html/form.ex#L616, the user should pass a `form` and a `field`. In order to improve DX, we could emit a warning if `form` or `field` is equal to `nil`. In other words, for example, if we do not find the `field` in the context or as a prop, we could emit a warning. Am I clearer?

This could be done in another PR.

Malian:
> I checked the output of the failed test and it seems the cause is a missing newline
Why does this happen with the CI but not locally? I tried the different combination of elixir/erlang/otp with asdf and it works.

Issue #516, opened 2021-10-29
Support dynamic stateful components


tiberiuc:
Are there any plans to add support for dynamic stateful components similar to Dynamic.Component (maybe Dynamic.LiveComponent ) ?

cybernetlab:
+1

msaraiva:
Yes, that's certainly on the roadmap 👍

tiberiuc:
@msaraiva I created a PR #518 with that one implemented , let me know what you think , thank you

tiberiuc:
This issues is done in PR #518

PR #518, opened 2021-11-03
Add Surface.Components.Dynamic.LiveComponent 

tiberiuc:
This PR is related to issue #516 and #519

* added new component to support dynamic stateful components

### DIFF
--- a/lib/surface/components/dynamic/live_component.ex
+++ b/lib/surface/components/dynamic/live_component.ex
@@ -1,0 +1,40 @@
+defmodule Surface.Components.Dynamic.LiveComponent do
+  @moduledoc """
+  A built-in component that allows users to inject dynamic live components into
+  a Surface template.
+  """
+
+  use Surface.LiveComponent
+
+  alias Surface.AST
+
+  @doc """
+  The module of the component
+  """
+  prop module, :module, required: true
+
+  @doc """
+  The default slot
+  """
+  slot default
+
+  def transform(node) do
+    %AST.Component{props: props, directives: directives, templates: templates, meta: meta} = node
+
+    {%{module: mod}, other_props} = AST.pop_attributes_values_as_map(props, [:module])
+
+    %AST.Component{
+      module: mod,
+      type: :dynamic_live,
+      props: other_props,
+      directives: directives,
+      templates: templates,
+      meta: meta
+    }
+  end
+
+  def render(assigns),
+    do: ~F"""
+      <div/>
+    """
+end
--- a/lib/surface/components/dynamic/live_component.ex
+++ b/lib/surface/components/dynamic/live_component.ex
@@ -33,8 +33,5 @@ defmodule Surface.Components.Dynamic.LiveComponent do
     }
   end
 
-  def render(assigns),
-    do: ~F"""
-      <div/>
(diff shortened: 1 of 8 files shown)


test/surface/components/dynamic/live_component_test.exs
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  defmodule StatefulComponent do
+    use Surface.LiveComponent
+
+    data label, :string, default: "Initial stateful"
+    data assigned_in_update, :any
+
+    def update(_assigns, socket) do
+      {:ok, assign(socket, assigned_in_update: "Assinged in update/2")}
```
Malian, line 23:
```suggestion
      {:ok, assign(socket, assigned_in_update: "Assigned in update/2")}
```

lib/surface/components/dynamic/live_component.ex
```diff
@@ -0,0 +1,40 @@
...
+      props: other_props,
+      directives: directives,
+      templates: templates,
+      meta: meta
+    }
+  end
+
+  def render(assigns),
+    do: ~F"""
+      <div/>
```
Malian, line 38: Should we add an empty `div` because it is a live component?

test/surface/components/dynamic/live_component_test.exs
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  test "render LiveComponent" do
+    {:ok, _view, html} = live_isolated(build_conn(), View)
+    assert html =~ "Initial stateful"
+  end
+
+  test "render data assigned in update/2" do
+    {:ok, _view, html} = live_isolated(build_conn(), View)
+    assert html =~ "Assinged in update/2"
```
Malian, line 74:
```suggestion
    assert html =~ "Assigned in update/2"
```

Malian (approved):
Except the minor changes, LGTM.

@msaraiva I do not fully understand the ins and outs of having dynamics components implemented like this. I think it's safest if you take a look at it ;)

lib/surface/components/dynamic/live_component.ex
tiberiuc: yes, without adding an empty html tag the compiler throws an error as a live component must have a root tag. I don't know if an empty div is the right choice, I'm open to any suggestions

msaraiva (commented):
Hi @tiberiuc!

Thank you for the PR. Great work! I left just a couple of comments for your consideration.

lib/surface/components/dynamic/live_component.ex
tiberiuc: good point , changed

test/surface/components/dynamic/live_component_test.exs
```diff
@@ -0,0 +1,81 @@
...
+    use Surface.Component
+
+    prop label, :string
+
+    def render(assigns) do
+      ~F"""
+      <div phx-click="click">{@label}</div>
+      """
+    end
+  end
```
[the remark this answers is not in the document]
tiberiuc: Removed StatelessComponent from tests and added tests for default slot

Malian:
:heart:

PR #522, opened 2021-11-06
Add migration guide from v0.5 to v0.6

### LINKED ISSUE #489: Improve documentation on Surface.Component vs Surface.LiveComponent for master branch
It would be great to emphasize the difference between `Surface.Component` and `Surface.LiveComponent` in the docs in more detail.

Documentation says: 
> Note: Stateless components cannot handle Phoenix LiveView events. If you need to handle them, please use a Surface.LiveComponent instead.

As I experienced, there's acutally more about this: Stateless components *don't handle any lifetime callbacks* like `mount/1` or `update/2`. This may be obvious when starting out now, but they *did handle them in the past*.

I migrated a LiveView app to Surface, and as data preprocessing inside the render callback is discouraged, there were some components using update for this task. These didn't have any event handling, so I thought going with `Surface.Component` is the key. There's no state management to be done, except data preparation for display. 

I used the published hex version until now and it worked perfectly well. Now I switched to the master branch, my components don't work anymore and I had a hard time figuring out the reason - it could be the bump to Phoenix 1.6.0-rc, or live_view 0.16 or something else. Just by chance I rememberd I read somewhere that `Surface.Component` used `Phoenix.LiveComponent` under the hood, and this changed in the master branch of surface. 

So I have rethink the usage of `mount` and `update` callbacks, and perhaps use a `LiveComponent` instead. This is perfectly fine, but that should be stated clearly somewhere.

Malian:
Closes #489

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -1,7 +1,8 @@
 # Migrating from `v0.5.x` to `v0.6.x`
 
-The `mount` and `update` callbacks are no longer available as `Surface.Component` now is built on top of function components instead of stateless live components.
+`Surface.Component` now is built on top of function components instead of stateless live components. This decision implies some breaking changes described below with solutions that allow you updade your code smoothly.
 
+## The `mount` and `update` callbacks are no longer available
 Basically, any data preparation that was done inside those callbacks must be moved to `render/1`. The Phoenix Live View API has been updated so you can use [`assign`](https://hexdocs.pm/phoenix_live_view/0.16.4/Phoenix.LiveView.html#assign/2), [`assign_new`](https://hexdocs.pm/phoenix_live_view/0.16.4/Phoenix.LiveView.html#assign/2), etc. in any function component.
 
 Before:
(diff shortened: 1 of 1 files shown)


Malian:
@msaraiva not sure if the example is relevant enough. WDYT?

miguel-s (commented):
Overall nice guide 👍 

I believe keeping the `mount` and `update` naming for the private functions in the "After" section can be confusing, it somewhat hides the fact that thos callbacks are gone, wdyt?

Malian:
@miguel-s I agree. The first time I saw this change in the test, I didn't understand what was going on. The idea of keeping the function is to show how to implement those "callback". That being said, I am fine with both.

```elixir
defmodule StatelessComponent do
  use Surface.Component
  
  prop count, :string
  data count_mount, :string
  data count_updated, :string
  
  @impl true
  def render(assigns) do
    assigns =
      assigns
      |> assign_new(assigns, :count_mount, fn -> assigns.count + 1 end)
      |> assign(assigns, :count_updated, assigns.count + 2

    ~F"""
    <div>{@count} - {@count_mount} - {@count_updated}</div>
    """
  end
end
```

msaraiva:
> I believe keeping the `mount` and `update` naming for the private functions in the "After" section can be confusing, it somewhat hides the fact that thos callbacks are gone, wdyt?

I agree. The other thing we forgot was to mention that since `Surface.Component` now is built on top of function components, the `@socket` assign is no longer available so if the users need the socket to render routes, they should use the endpoint instead.

msaraiva (approved):
❤️

PR #523, opened 2021-11-06
Fix tests that test nothing

Malian:
@msaraiva maybe I am missing something with macro, but I believe the tests test nothing ^^


msaraiva (approved):
Yup. Good catch!

PR #524, opened 2021-11-07
Fix typo in error message

NduatiK:


Malian:
@NduatiK good catch! Thanks for the PR. It seems that some other tests should be updated to take into account your changes. After fixing this minor issues, we will be able to merge it :+1:

NduatiK:
Done.

Malian:
❤️

PR #530, opened 2021-11-10
Fix FileInput docs

miguel-s:
The example in the FileInput moduledocs was not updated and would not work, leading to confusion.

@msaraiva would it be possible to regenerate the docs with this change?


Malian (approved):

msaraiva:
@miguel-s sure. I’ll publish it. Thanks!

PR #535, opened 2021-11-23
Merge surface_formatter into Surface

paulstatezny:
This merges in the newly released `surface_formatter 0.7.0`, including the `Surface.Formatter.Plugin` Formatter Plugin for Elixir 1.13. 😄 

To do:

- Add Elixir 1.13 into matrix in `.github/workflows/ci.yml` and ensure `Surface.Formatter.PluginTest` only runs for that version of Elixir.
- Update README, which points to separate `surface_formatter` Hex package.
- Add any information from [surface_formatter's README](https://github.com/surface-ui/surface_formatter/blob/master/README.md)? It is not included in this PR, but I think all of the information there is redundant, repeating information inside of other modules included in this PR.

Are there any other architectural or organizational code changes we want to include?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
           - elixir: 1.12.1
             otp: 24.0
             check_formatted: true
-          - elixir: 1.13.0-rc.1
+          - elixir: 1.13.0
             otp: 24.0
             run_plugin_tests: true
             warnings-as-errors: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,13 +16,14 @@ jobs:
         include:
           - elixir: 1.11.4
             otp: 22.3
-            warnings-as-errors: true
           - elixir: 1.11.4
             otp: 23.3
-            warnings-as-errors: true
           - elixir: 1.12.1
             otp: 24.0
             check_formatted: true
+          - elixir: 1.13.0-rc.1
+            otp: 24.0
+            run_plugin_tests: true
             warnings-as-errors: true
     env:
       MIX_ENV: test
@@ -42,3 +43,6 @@ jobs:
       - run: mix compile --warnings-as-errors
         if: matrix.warnings_as_errors
       - run: mix test
+        if: matrix.run_plugin_tests
+      - run: mix test --exclude plugin
+        if: ${{ !matrix.run_plugin_tests }}
--- a/lib/mix/tasks/surface/surface.format.ex
+++ b/lib/mix/tasks/surface/surface.format.ex
@@ -1,0 +1,500 @@
+defmodule Mix.Tasks.Surface.Format do
+  @shortdoc "Formats Surface ~F sigils and .sface files in the given files/patterns"
+
+  @moduledoc """
+  **To format Surface code using Elixir 1.13 or later, use
+  `Surface.Formatter.Plugin`.**
+
+  Formats Surface `~F` sigils and `.sface` files in the given files and patterns.
+
+  ```bash
+  $ mix surface.format "lib/**/*.{ex,exs}" "test/**/*.{ex,exs}"
+  $ cat path/to/file.ex | mix surface.format -
+  ```
+
+  Takes the same options as `mix format` except for `--check-equivalent`.
+
+  ## Formatting options
+
+  Like `mix format`, the Surface formatter reads a `.formatter.exs` file in the
+  current directory for formatter configuration. The Surface formatter accepts
+  the same options as `mix format`. Read more about the expected format of
+  `.formatter.exs` and the shared configuration options
+  [documented here](https://hexdocs.pm/mix/master/Mix.Tasks.Format.html#module-formatting-options).
+
+  The Surface formatter also takes the following two additional options
+  specified in `.formatter.exs`:
+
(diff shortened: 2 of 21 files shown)


lib/mix/tasks/surface/surface.format.ex
```diff
@@ -0,0 +1,499 @@
+defmodule Mix.Tasks.Surface.Format do
```
paulstatezny, line 1: I'm not sure if we want to pull in this mix task. It's a _lot_ of code copied over from `Mix.Tasks.Format`, so it feels a bit hacky to keep it around now that Formatter Plugins are available. However, not everyone is going to immediately switch over to Elixir 1.13 as soon as it's available, so it may be best to keep this.

msaraiva: Agreed. Let's keep it for for a while.

.github/workflows/ci.yml
```diff
@@ -16,13 +16,14 @@ jobs:
...
           - elixir: 1.11.4
             otp: 23.3
-            warnings-as-errors: true
           - elixir: 1.12.1
             otp: 24.0
             check_formatted: true
+          - elixir: 1.13.0-rc.1
+            otp: 24.0
+            run_plugin_tests: true
             warnings-as-errors: true
```
paulstatezny, line 27: @msaraiva The plugin test will fail before Elixir 1.13 because it's dependent upon the Mix Formatter Plugin feature. So here, I'm adding Elixir 1.13(-rc.1) to the CI matrix and only allowing that test to run for that version.

It turns out that the following two tests fail on Elixir 1.13. This happens on `master` as well; it's unrelated to this PR.

```
  1) test nested if/elseif/else parser error message contains the correct line with nested struct (Surface.Constructs.IfTest)
Error:      test/surface/constructs/if_test.exs:321
     Expected exception Surface.Compiler.ParseError but got CompileError (code: invalid module name: :"Elixir.Surface.Constructs.IfTest.(test nested if/elseif/else parser error message contains the correct line with nested struct) at line 349")
     code: assert_raise(Surface.Compiler.ParseError, message, fn ->
     stacktrace:
       (elixir 1.13.0-rc.1) src/elixir_module.erl:97: :elixir_module.compile/5
Warning:        (elixir 1.13.0-rc.1) src/elixir_lexical.erl:15: :elixir_lexical.run/3
       (stdlib 3.16.1) erl_eval.erl:685: :erl_eval.do_apply/6
Warning:        (elixir 1.13.0-rc.1) src/elixir.erl:289: :elixir.recur_eval/3
       (elixir 1.13.0-rc.1) src/elixir.erl:274: :elixir.eval_forms/3
Warning:        (elixir 1.13.0-rc.1) lib/code.ex:797: Code.eval_quoted/3
       test/surface/constructs/if_test.exs:349: anonymous fn/0 in Surface.Constructs.IfTest."test nested if/elseif/else parser error message contains the correct line with nested struct"/1
Warning:        test/surface/constructs/if_test.exs:348: (test)


  2) test nested if/elseif/else compile error message contains the correct line with nested struct (Surface.Constructs.IfTest)
Error:      test/surface/constructs/if_test.exs:290
     Wrong message for CompileError
     expected:
       "code:12: invalid value for property \"prop\". Expected a :list, got: \"some string\"."
Warning:      actual:
       "code: invalid module name: :\"Elixir.Surface.Constructs.IfTest.(test nested if/elseif/else compile error message contains the correct line with nested struct) at line 317\""
     code: assert_raise(CompileError, message, fn ->
     stacktrace:
       test/surface/constructs/if_test.exs:316: (test)
```

msaraiva (approved):
Hey @paulstatezny. Great work!

It looks good to me.

I fixed the tests. The issue was that Elixir v1.13 does not accept module names containing `/` anymore so since `compile_surface` defines modules based on the test's full name, it failed for those two.

The one thing we need to check is if we need to change the converter to check if the version is `>= 1.13` so it can configure `.formatter.exs` properly. We can do this on a separate PR if you like.

msaraiva:
@paulstatezny sounds good to me!

paulstatezny:
This is almost ready. `mix surface.init` is only currently able to add `Surface.Formatter.Plugin` to `:plugins` in `.formatter.exs` if `:plugins` is already defined.

msaraiva:
@paulstatezny I'll update the patch to handle that case.

.github/workflows/ci.yml
```diff
@@ -16,13 +16,14 @@ jobs:
...
           - elixir: 1.11.4
             otp: 22.3
-            warnings-as-errors: true
           - elixir: 1.11.4
             otp: 23.3
-            warnings-as-errors: true
           - elixir: 1.12.1
             otp: 24.0
             check_formatted: true
+          - elixir: 1.13.0-rc.1
```
msaraiva, line 24:
```suggestion
          - elixir: 1.13.0
```

msaraiva:
@paulstatezny I've just pushed the changes to make it work when there's no `:plugins`. Please let me know if you find any edge case I missed.

paulstatezny:
@msaraiva Perfect! I don't think there are any other scenarios regarding `.formatter.exs` and `:plugins`.

So I think this PR is ready to go!

msaraiva:
@paulstatezny please, go ahead and thank you for the PR ❤️

paulstatezny:
Awesome! Thanks for all your help! 🙂

Issue #532, opened 2021-11-14
mix surface.init failed to patch mix.exs


davydog187:
## Describe the bug
After trying to introduce surface to a (nearly brand new) Phoenix project, the `surface.init` task was not able to patch my `mix.exs` file.

## How to reproduce it
1. Clone my repo at this commit https://github.com/davydog187/website/tree/f4b76f64356a68c7a86fb802b38a5bc3832fd031
2. Run `mix deps.get`
3. Run `mix surface.init`
4. Observe the following error

```elixir
* patching .formatter.exs
* patching .gitignore
* patching assets/js/app.js
* patching config/dev.exs
* patching lib/website_web.ex
* patching mix.exs
** (FunctionClauseError) no function clause matching in Sourceror.Range.do_get_range/1

    The following arguments were given to Sourceror.Range.do_get_range/1:

        # 1
        :compilers

    Attempted function clauses (showing 10 out of 20):

        defp do_get_range({:__aliases__, meta, segments})
        defp do_get_range({:__block__, meta, [string]}) when is_binary(string)
        defp do_get_range({:__block__, meta, [number]}) when is_integer(number) or is_float(number)
        defp do_get_range({:__block__, meta, [atom]}) when is_atom(atom)
        defp do_get_range({:__block__, _, args} = quoted)
        defp do_get_range({form, meta, context}) when is_atom(form) and is_atom(context)
        defp do_get_range({left, right})
        defp do_get_range(list) when is_list(list)
        defp do_get_range({:->, _, [left_args, right]})
        defp do_get_range({{:., _, [Access, :get]}, _, _} = quoted)
        ...
        (10 clauses not shown)
```

## The behavior you expected
If it can't patch, give me a warning.

## Your Environment
Surface: v0.6.1
LiveView: v0.16.4
Elixir: v1.12.2

PR #537, opened 2021-12-04
Fix compiler injection patches

doorgan:
This addresses #532 

I left a comment in the code, but the reason for this bug is that when we have a qualified call, like `Mix.compilers()`, the AST will look like this:
```elixir
{{:., inner_metadata, [[:__aliases__, _, [:Mix]}, :compilers]}, outer_metadata, []}
```
Note that `:compilers`(ie the identifier) is a "naked" atom, ie it's not wrapped in a block, so it doesn't have metadata. The line and column information for the identifier lives in the `outer_metadata`, while the dot position information lives in the `inner_metadata`. This means that we can't know the range of the inner `:.` call if we don't have the parent too.

When navigating down such tree, the first child will be the `:.` call, then the arguments follow. If we use `Z.rightmost` to get the last child, but the call doesn't have any arguments like in this case, then the only child we'll get is the `:.` call.

The code generating this patch was trying to generate a patch on this `:.` call, but since that call alone doesn't have enough information to generate a range to patch, it crashed. The solution is basically to check if we landed in such call and if so then work with the parent node.

### DIFF
--- a/lib/mix/tasks/surface/surface.init/ex_patcher.ex
+++ b/lib/mix/tasks/surface/surface.init/ex_patcher.ex
@@ -207,7 +207,7 @@ defmodule Mix.Tasks.Surface.Init.ExPatcher do
   def replace_code(%__MODULE__{code: code} = patcher, fun) when is_function(fun) do
     patch(patcher, [preserve_indentation: false], fn zipper ->
       node = Z.node(zipper)
-      range = Sourceror.get_range(node)
+      range = Sourceror.get_range(node, include_comments: true)
       code_to_replace = get_code_by_range(code, range)
       fun.(code_to_replace)
     end)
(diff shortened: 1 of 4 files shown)


msaraiva:
Hey @doorgan!

Thank you for the PR. As you could probably see, I tried to implement a higher level API that allows applying groups of patches to elixir projects. I'm not fully satisfied with the current API but it works as expected and it has all I need to create precise patches for `mix surface.init` and eventually for `mix surface.convert`. 

There were a few issues while designing it, especially due to problems that forced me to update the parent node, which should be avoided as it may format more code than it should. Anyway, I think a similar API would be great to have in Sourceror itself. I remember we've discussed this a while ago and I told I'd give it a try so here's my initial attempt :)

Let me know if you're interested in something similar. I'd be glad to contribute back.

doorgan:
@msaraiva Yup, I'll spend more time checking out ExPatcher
What I saw so far is quite promising already, something similar would be definitely a welcome addition to Sourceror :)

So far most of my focus on my similar focus was on studying the ruby [https://github.com/whitequark/parser](parser)'s Rewriter, but I didn't do much progress. Marc-André had some insights on this [here](https://github.com/whitequark/parser/issues/401), but again, I dind't do much progress as to come with a good api for Sourceror yet, and I'm yet not sure about how applicable these ideas are

PR #539, opened 2021-12-06
props can be marked as deprecated

### LINKED ISSUE #513: Proposal: add `deprecated` option to props
It would be nice to have an easy way to deprecate a `prop` and show a warning if the prop is still used. 
The option would be: `[deprecated: "deprecation message shown in the warning at compile time"]`
This is an example of what I would have in mind:
```
prop old_prop, :any, deprecated: "This is an old prop that will be deprecated and is being replaced by other_prop."
```
Obviously this would not work if the prop is used dynamically, but I still think that this can be useful.

Zurga:
closes #513 

One of the tests is still failing due to the output of the terminal concerning coloring, I could use a little help because the regex seems similar to other tests...

Thanks!


test/surface/integrations/properties_test.exs
```diff
@@ -879,4 +879,57 @@ defmodule Surface.PropertiesSyncTest do
...
+    output =
+      capture_io(:standard_error, fn ->
+        compile_surface(code, %{deprecated: "anst"})
+      end)
+
+    assert output =~ ~r"""
+           the prop `deprecated` is marked as deprecated.
+           Hint: This is prop is deprecated
+
+             code.exs:1:\
```
RudolfMan, line 932:
```suggestion
           Hint: This prop is deprecated

             code:1:\
```

Zurga:
@RudolfMan Thanks for the fix, I should go and have my eyes checked :p.

tiagoefmoraes:
> One of the tests is still failing due to the output of the terminal concerning coloring, I could use a little help because the regex seems similar to other tests...

The tests seems to be fine now. We have an [error with code formatting](https://github.com/surface-ui/surface/runs/5448772689?check_suite_focus=true).

```bash
mix format --check-formatted
** (Mix) mix format failed due to --check-formatted.
The following files are not formatted:

  * lib/surface/api.ex
  * test/surface/integrations/properties_test.exs
  * lib/surface/compiler.ex
```

Make sure you format it with Elixir 1.12.1 as that is the version we [check formatting against](https://github.com/surface-ui/surface/blob/main/.github/workflows/ci.yml#L21-L23).

Zurga:
Thanks, for pointing that out. I did not have a lot of time recently. I will fix the formatting asap.

PR #541, opened 2021-12-18
Merge bugfixes from surface_formatter 0.7.3 and 0.7.4

paulstatezny:
Fixes formatting of multi-line strings in attribute expressions, e.g.

```html
<div
  class="test"
  x-data={"{
    foo: '#{@bar}',
    baz: '#{@qux}'
  }"}
/>
```

Prior to the patch, every time the formatter runs it will add an indentation level to the inner contents of the string. For example:

```html
x-data={"{
    foo: '#{@bar}',
    baz: '#{@qux}'
  }"}
```

```html
x-data={"{
      foo: '#{@bar}',
      baz: '#{@qux}'
    }"}
```

```html
x-data={"{
        foo: '#{@bar}',
        baz: '#{@qux}'
      }"}
```

The patched code is somewhat simpler because it turns out some nested conditional logic should not have been conditionally applied.

### DIFF
--- a/lib/surface/formatter/phases/render.ex
+++ b/lib/surface/formatter/phases/render.ex
@@ -389,13 +389,30 @@ defmodule Surface.Formatter.Phases.Render do
           []
         end
 
-      {:<<>>, _, nodes} when is_list(nodes) ->
+      [{:do, string}] when is_binary(string) ->
+        if String.contains?(string, "\n") do
+          [string]
+        else
+          []
+        end
+
+      {operation, _, nodes} when is_atom(operation) and is_list(nodes) ->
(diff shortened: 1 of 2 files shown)


paulstatezny:
Thank you @simonmcconnell for [reporting this](https://github.com/surface-ui/surface_formatter/issues/49)! 🚀

paulstatezny:
I added a commit with a patch from v0.7.4. [The problem scenario is described here.](https://github.com/surface-ui/surface_formatter/issues/55) Thanks @maennchen!

msaraiva:
@paulstatezny thanks! ❤️

Issue #538, opened 2021-12-06
Is it supported to use Surface stateless components without LiveView?


guisehn:
I'm testing adding Surface to an existing Phoenix project that currently doesn't have LiveView, and I'm wondering if the use case of stateless components on regular pages (non-LiveView) is officially supported.

I was able to render a stateless Surface component on a regular controller by overriding the `render("index.html", assigns)` of my view file:

```ex
defmodule PhoenixProjectWeb.PageView do
  use PhoenixProjectWeb, :view

  alias PhoenixProjectWeb.Components.Button

  def render("index.html", assigns) do
    ~F"""
    <Button>I'm a Surface button</Button>
    """
  end
end
```

I wasn't able to render it on my template though (`templates/page/index.html.heex`).  There is a page on the docs talking about [usage with Phoenix templates](https://surface-ui.org/usage_with_phoenix_templates) but it only covers LiveView templates, not regular ones.

Tried changing the extension of my template file from `.heex` to `.sface` but there's no configured engine for it.

So is this use case supported? And is there a way to use Surface components on a controller with a template file? (without defining `render/2` on the view file)

msaraiva:
Implemented in #543 and #548.

PR #543, opened 2021-12-27
Add support for rendering .sface as template for regular controllers

guisehn:
Creates a new `Surface.View` module, that can be used to enable render of `.sface` templates on regular Phoenix controllers (without LiveView).

This can be useful to take advantage of Surface stateless components on regular pages, and would solve https://github.com/surface-ui/surface/issues/538.

To use it, add `use Surface.View, root_path: "lib/your_app_web/templates"` to a view module. At compile time, it will look for `#{root_path}/templates/#{view}/*.html.sface` templates, and generate a `render("template_name.html", assigns)` clause for each template.

To avoid repeating the `root_path` on all views, it can be added to your `your_app_web.ex` as exemplified on the `Surface.View` moduledoc.

P.S.: I initially tried implementing it using a `Phoenix.Template.Engine`, but had no luck because `Surface.Compiler.compile/3` requires an `env` that is only available on the compilation phase, so I switched to a `@before_compile` approach similar to [Surface.Renderer](https://github.com/surface-ui/surface/blob/master/lib/surface/renderer.ex).


guisehn:
I've created a repo with a demo app at https://github.com/guisehn/sface-on-view-demo

Commit history shows the setup process after `mix surface.init`: https://github.com/guisehn/sface-on-view-demo/commits/main

Only the first commit (https://github.com/guisehn/sface-on-view-demo/commit/a2c1d2cad9d3b242e176997ab76e05d28e9f7af6) is required to make it work, the other ones are converting templates to `sface` to demo it.

msaraiva:
Hi @guisehn. Great stuff! ❤️

I'll review it today so we can merge it and make it available in `v0.7` which should be released this week.

Thank you!

lib/surface/view.ex
```diff
@@ -0,0 +1,105 @@
+defmodule Surface.View do
+  @moduledoc """
+  Use this module on regular Phoenix views to enable .sface templates
+
+  The template files need to end with `.html.sface`
```
msaraiva, line 5: Is it be possible to optionally use just `.sface`?

I mean, Surface can only be used with extended HTML (Surface) files so something like using `.json.sface` or anything similar shouldn't exist.

msaraiva (commented):
This looks great!

msaraiva:
I also wonder if we could (or should) optionally allow using a colocated template file instead of placing it in the `templates` folder. Similar to what we do with components 🤔

lib/surface/view.ex
guisehn: Yes, adjusted it: https://github.com/surface-ui/surface/pull/543/commits/f0ccd5928199a154182a7392d95ea6e254f25a0d

guisehn:
> I also wonder if we could (or should) optionally allow using a colocated template file instead of placing it in the `templates` folder. Similar to what we do with components 🤔

For components, there's a 1:1 relationship between the component and the `.sface` file so they can stay on the same folder, but for regular controllers/views we'll often have multiple templates for a single view e.g. `index.sface`, `show.sface`, `new.sface`, `update.sface`, etc.

If we support colocated templates, we'll need to define a convention different than components to prevent conflicting `.sface` templates belonging to different views. Not sure if it's worth introducing a new convention, I'd go with `templates` since it's a well known pattern.

msaraiva:
> If we support colocated templates, we'll need to define a convention different than components to prevent conflicting `.sface` templates belonging to different views.

Sure. I think supporting `.index.sface`, `.show.sface`, etc. would be a good start. But that's not what this PR is about. We can discuss something like that in the future in case anyone is interested.

Thank you! ❤️

PR #547, opened 2022-01-06
Support nested views

guisehn:
Adds support for views inside submodules, this was missing on https://github.com/surface-ui/surface/pull/543

After this change,  rendering a sface for `ExampleWeb.Foo.BarView` will look for templates on `lib/example_web/templates/foo/bar/*.sface`. Before the change, it'd search on `lib/example_web/templates/bar/*.sface`


msaraiva:
@guisehn thanks! ❤️

Issue #526, opened 2021-11-08
Dynamic Component can't handle events with target


evansorenson:
## Describe the bug
Using a dynamic component and passing event with target. Error shown below.
> invalid value for property "submit". Expected a :any, got: ["start", {:target, %Phoenix.LiveComponent.CID{cid: 1}}].

## How to reproduce it
Pass an event to dynamic component as prop (shown below).  
`<Dynamic.Component module={@game_config.components.options} game_changeset={@game_changeset} submit={"start", target: @myself} change={"validate", target: @myself}/>`

## The behavior you expected
Should be able to pass event with target for dynamic component. 

## Your Environment
Surface: v0.6.1
LiveView: v0.16.4
Elixir: v1.11.4

Malian:
@evansorenson Thanks! I can confirm the bug.

After few investigations, it seems that the `event` properties are not treated as `:event` but `:any`.

I'll look at this during the we.

Malian:
It happens because we convert all prop to  `:any` here

https://github.com/surface-ui/surface/blob/d1c6b9e8df02c21507b35bbd4c882e165be771fe/lib/surface/type_handler.ex#L235-L239

@tiberiuc I suspect we will have the same issue with the new `Dynamic.LiveComponent`

tiberiuc:
@Malian yes, we are having the same issue with Dynamic.LiveComponent, when I will have time and nobody fix it in the meantime I will try to fix it

msaraiva:
@evansorenson and @tiberiuc, in case any of you want to give master a try, please let me know if you still run into any issue.

Thanks for reporting this problem!

@Malian we have now a new private type called `:dynamic` which is automatically set for all dynamic components' attributes. This way we can skip the compile-time type validation/conversion but still validate/convert the values to the proper types at runtime.

PR #553, opened 2022-01-12
Properly handle dynamic component attributes values

msaraiva:
Fixes https://github.com/surface-ui/surface/issues/526.


lib/surface/directive/events.ex
```diff
@@ -67,25 +67,19 @@ defmodule Surface.Directive.Events do
...
     expr_meta = Helpers.to_meta(expr_meta, meta)
+    quoted = Surface.TypeHandler.expr_to_quoted!(original, name, :event, expr_meta)
+    expr = AST.AttributeExpr.new(quoted, original, expr_meta)
 
-    value =
-      original
-      |> Surface.TypeHandler.expr_to_quoted!(name, :event, expr_meta)
-      |> case do
-        [name, opts] when is_binary(name) and is_list(opts) -> Keyword.put(opts, :name, name)
-        [name | opts] when is_binary(name) and is_list(opts) -> Keyword.put(opts, :name, name)
```
msaraiva, line 85: @lnr0626 I don't see how this could ever match. All tests are passing. I believe it's just a leftover.

lib/surface/type_handler.ex
```diff
@@ -204,16 +205,16 @@ defmodule Surface.TypeHandler do
...
-  def runtime_prop_value!(module, name, value, node_alias) do
-    type =
+  def runtime_prop_value!(module, name, clauses, opts, node_alias, original, ctx) do
+    {type, _opts} =
       attribute_type_and_opts(module, name, %{
         node_alias: node_alias || module,
-        caller: __ENV__,
-        file: __ENV__.file,
-        line: __ENV__.line
+        caller: %Macro.Env{module: ctx.module},
```
lnr0626, line 212: is this going to potentially cause issues if we expect caller to be a fully populated env and it only has the module?

lnr0626 (approved):
this looks good to me, only had one minor question

lib/surface/type_handler.ex
msaraiva: @lnr0626 I'm not sure what mean. The problem with `runtime_prop_value!` is that since it's called only at runtime, the warnings emitted were all pointing to the the TypeHandler file/module instead of the caller component. The only way I managed to point to the caller was to pass a clean env overriding the module.

lnr0626: that makes sense - perhaps a better phrasing of my question is will `%Macro.Env{module: ctx.module}` result in a fully populated `Macro.Env` struct?

Though that's only potentially an issue if we use `ctx.caller` elsewhere in the codebase and expect more than just the `module` attribute to be set

msaraiva: oh, I see. Well, the caller context (ctx) was meant to hold as minimum information as possible as it has to be passed as a literal value to `build_assigns`, so we should never have a `ctx.caller`.

PR #559, opened 2022-01-19
Updated Surface.View doc for added clarity

### LINKED ISSUE #558: Use with live layout
Is it possible to use Surface components on the `live.html.heex` layout? 

I've updated my application according to the recommended instructions [here](https://surface-ui.org/usage_with_phoenix_templates) by updating all macros to use `Surface` modules. 

But when I try to use a Surface component on the `heex` file I get a compilation error saying that I have an invalid tag. Note that using that component in another `Surface` live view works.

gabrielrinaldi:
closes #558

Just adding a little bit more context to the documentation.

### DIFF
--- a/lib/surface/view.ex
+++ b/lib/surface/view.ex
@@ -10,7 +10,12 @@ defmodule Surface.View do
         # ...
 
         def view do
+          use Phoenix.View,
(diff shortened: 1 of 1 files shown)


msaraiva:
@gabrielrinaldi thanks!

PR #564, opened 2022-02-03
Support Inputs' prop for as string

msaraiva:
The underlying phoenix's [inputs_for](https://github.com/phoenixframework/phoenix_html/blob/v3.2.0/lib/phoenix_html/form.ex#L438) function accepts it so we should accept it too.


miguel-s (approved):
lgtm!

I guess that in a form backed by a changeset you'll get an error if you use a string in the `for` prop (like with regular fields).

msaraiva:
Yeah, I believe so.

PR #565, opened 2022-02-07
fix set `@moduledoc false` in catalogue examples

herminiotorres:
To prevent receiving `:hidden`, instead of `%{"en" => "..."}` in
`Surface.Catalogue.get_metadata/1`, when using set `@moduledoc false`.


msaraiva:
@herminiotorres thanks! ❤️

PR #574, opened 2022-03-10
config hooks extension

onelesd:
I know there is another PR for this, but it has stalled and I also think this is simpler and is not specific to any language. I use it for typescript and in Phoenix 1.6 it requires nothing other than renaming *.js files to *.ts.


simonmcconnell:
What if you have one written in typescript, one in JavaScript, one in ReScript, one in PureScript, one in ... ?

onelesd:
Seems like a bit of a corner-case. I suppose if you were porting existing hooks, for example, it could be useful.

It could easily be done with a list of extensions and looking for a match in order of preference.

Typescript being a superset of Javacript, and esbuild working with Typescript out-of-the-box, the Javascript index that the surface compiler creates can `import` the hooks without modification. Would this be true for PureScript, ReScript, others? I am unfamiliar with them, but it may require overriding `index_content/0` and `index_content/1` to understand their imports.

PR #575, opened 2022-03-10
Remove old slot template syntax

tiagoefmoraes:
So we have only the new one.

* `<#template>` -> `<:default>`
* `<#template slot="header">` -> `<:header>`

Also created the 0.8 converter to handle this migration.
The current formatter already converted `<#template slot="name">` to `<:name>`


msaraiva (approved):
@tiagoefmoraes fantastic job!

Thanks for adding the converter as well 🔥🔥🔥

PR #576, opened 2022-03-12
Generate assets for `.hooks.` with the `js`, `jsx`, `ts`, `tsx` extensions, using the first one

### LINKED ISSUE #444: Feature: Add Typescript Support

### LINKED ISSUE #482: Custom hooks extension support

tiagoefmoraes:
It looks for `js`, `jsx`, `ts`, `tsx` and uses the first one found.

It also introduces `%Diagnostics{}` support to `Mix.Tasks.Compile.Surface`

* Fixes #444
* Fixes #482
* Closes #483
* Closes #574


msaraiva (approved):
@tiagoefmoraes fantastic job! ❤️

msaraiva:
@yusufyildirim and @onelesd thank you so much for your contributions. The discussions in #483 and #574 helped a lot. We decided to take a step forward and implement a more robust solution that could handle all accepted extensions and that could also handle assets in deps. For instance, if a project was using, `TS` and a component from a dependency was using `JS`, the previous PRs would not work as expected. This PR addresses that along with proper diagnostics handling.

Feel free to try it out on master and let us know in case you run into any issue.

Thanks again!

andresgutgon:
Hi, I discovered today Surface. Looks a great way of colocating JS/TS and Elixir code. But one question I have is how Surface deal with JS ecosystem.

How I build a component that has NPM dependencies?

PR #579, opened 2022-03-14
Stop perpetually indenting expression strings with newlines

paulstatezny:
### What is this?

This PR fixes a formatter bug reported here: https://github.com/surface-ui/surface_formatter/issues/59
And fixed separately in legacy `surface_formatter` package here: https://github.com/surface-ui/surface_formatter/pull/61

### What is the bug?

Any string with newlines inside of an expression (_not_ an attribute expression) would be indented further each time the formatter ran. Like this:

```elixir
{foo("bar
  baz
  qux
")}
```
```elixir
{foo("bar
    baz
    qux
  ")}
```
```elixir
{foo("bar
      baz
      qux
    ")}
```
```elixir
{foo("bar
        baz
        qux
      ")}

### DIFF
--- a/lib/surface/formatter/phases/render.ex
+++ b/lib/surface/formatter/phases/render.ex
@@ -36,6 +36,12 @@ defmodule Surface.Formatter.Phases.Render do
           expression
           |> String.trim()
           |> Code.format_string!(opts)
+          |> to_string()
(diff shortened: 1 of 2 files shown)


msaraiva:
@paulstatezny thank you! ❤️

PR #586, opened 2022-04-15
Props can be marked as deprecated

Zurga:
Closes #539


msaraiva:
I'm closing this as any new addition of that kind should be addressed in the new declarative API (v0.18) of LV itself.

PR #592, opened 2022-04-20
add some missing void tags (per w3.org)

rubysolo:
(ref https://www.w3.org/TR/2014/REC-html5-20141028/syntax.html#void-elements -- note that the `command` element is not in that list, but it is used as a void element in the examples here: https://www.w3.org/TR/2011/WD-html5-author-20110809/the-command-element.html.  🤷 )

We ran into a problem with the `embed` element, which seems to be missing from the list of void elements.  Without this change, the formatter re-formats `<embed />` as `<embed>`, and then subsequent formats blow up because the embed has no end tag.

### DIFF
--- a/lib/surface/compiler/tokenizer.ex
+++ b/lib/surface/compiler/tokenizer.ex
@@ -14,15 +14,18 @@ defmodule Surface.Compiler.Tokenizer do
     "base",
     "br",
     "col",
+    "command",
+    "embed",
     "hr",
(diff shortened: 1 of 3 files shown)


mackshkatz (approved):

tiagoefmoraes (approved):

msaraiva:
@rubysolo thanks! ❤️

Issue #596, opened 2022-05-02
Live Reload for layout .sface file is not patched with mix surface.init --layout


Malian:
## Describe the bug

When modifying a layout patched by running `mix surface.init --layout`, live reload does not work.

## How to reproduce it

1. Install fresh project with mix phx.new
2. Install Surface with mix surface.init --layout
3. Modify a patched layout (root.sface for exemple)

## The behavior you expected

The page reload automatically

## Fix
In `dev.exs`
```
~r"lib/we_live_chat_web/templates/.*(eex)$"
```
should be
```
~r"lib/we_live_chat_web/templates/.*(eex|sface)$"
```
after `mix surface.init --layout`

## Your Environment
Surface: master
LiveView: v0.17
Elixir: v1.13.4

joshprice:
@Malian can you please confirm this resolves the issue on main?

Malian:
@joshprice Yes, it's solved! Thanks for your contribution :heart:

PR #598, opened 2022-05-06
Patch live reload pattern for surface layouts

joshprice:
This assumes that the reload pattern should be applied regardless of whether the `--layout` option is used.

Fixes issue #596


msaraiva:
@joshprice thank you! ❤️

PR #603, opened 2022-05-12
Move "Unknown property" warning to mix compiler

tiagoefmoraes:
Runtime warnings, for dynamic components, are still in the TypeHandler

### DIFF
--- a/test/mix/tasks/compile/surface/validate_components_test.exs
+++ b/test/mix/tasks/compile/surface/validate_components_test.exs
@@ -6,6 +6,33 @@ defmodule Mix.Tasks.Compile.Surface.ValidateComponentsTest do
   alias Mix.Tasks.Compile.Surface.ValidateComponents
   alias Mix.Task.Compiler.Diagnostic
 
+  defmodule StringProp do
+    use Surface.Component
+    prop text, :string
+    def render(assigns), do: ~F[{@text}]
+  end
+
+  test "should return diagnostic when unkwnown prop is passed to Component" do
+    component =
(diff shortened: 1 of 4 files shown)


test/mix/tasks/compile/surface/validate_components_test.exs
```diff
@@ -6,6 +6,33 @@ defmodule Mix.Tasks.Compile.Surface.ValidateComponentsTest do
...
+  defmodule StringProp do
+    use Surface.Component
+    prop text, :string
+    def render(assigns), do: ~F[{@text}]
+  end
+
+  test "should return diagnostic when unkwnown prop is passed to Component" do
+    component =
+      quote do
+        ~F[<StringProp unknown />]
```
msaraiva, line 18: I usually test attribute-related validation having them defined not in the first line, this way we can make sure the information about the position (line) is correct and not just getting default value. Like in the replaced test:

```html
<div>
  <Button
     ...
     nonExistingProp="1"
   />
</div>
```
Is this been tested somewhere else? If so, does it cover any case?

BTW, the same question goes to the other PR https://github.com/surface-ui/surface/pull/602/.

tiagoefmoraes: It is being tested, and I intend write more focused tests for this when we include column information in the diagnostics.

https://github.com/surface-ui/surface/blob/33a45d37825d9730fdbdd78be1df2e7a200cbfc4/test/mix/tasks/compile/surface/validate_components_test.exs#L226

PR #604, opened 2022-05-19
add priv folder for Surface Formatter

herminiotorres:
Only for applications who have been using the Elixir version previous than `< 1.13.0`.

### DIFF
--- a/lib/mix/tasks/surface/surface.init/project_patchers/formatter.ex
+++ b/lib/mix/tasks/surface/surface.init/project_patchers/formatter.ex
@@ -37,7 +37,7 @@ defmodule Mix.Tasks.Surface.Init.ProjectPatchers.Formatter do
     %{
       name: "Add `surface_formatter` dependency",
       update_deps: [:surface_formatter],
-      patch: &FilePatchers.MixExs.add_dep(&1, ":surface_formatter", ~S("~> 0.6.0")),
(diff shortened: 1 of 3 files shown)


lib/mix/tasks/surface/surface.init/project_patchers/formatter.ex
```diff
@@ -57,7 +57,7 @@ defmodule Mix.Tasks.Surface.Init.ProjectPatchers.Formatter do
   def add_surface_inputs_to_formatter_config() do
     %{
       name: "Add file extensions to :surface_inputs",
-      patch: &FilePatchers.Formatter.add_config(&1, :surface_inputs, ~S(["{lib,test}/**/*.{ex,exs,sface}"])),
+      patch: &FilePatchers.Formatter.add_config(&1, :surface_inputs, ~S(["{lib,test,priv}/**/*.{ex,exs,sface}"])),
```
msaraiva, line 60: Would something like `{lib,test,priv/catalogue}/**/*.{ex,exs,sface}` work? Since we may have other folders in `priv` that we don't want to touch by default.

msaraiva:
@herminiotorres thanks for the PR!

> Should I update the surface formatter to 0.7.4 since the current version on this patcher is 0.6.0?

Yes, please. That would be great!

msaraiva (approved):

PR #606, opened 2022-05-25
Add catalogue_test/1 macro and multiple examples

msaraiva:
## New features

* Add `catalogue_test/1` macro to generate basic tests for catalogue examples and playgrounds
* Add `Surface.Catalogue.Examples` to allow defining multiple stateless examples in a single module
* Add private `Surface.components/2` function to be used by builtin tools (compiler, catalogue, etc.) to retrieve the list of available components.

@tiagoefmoraes after this PR gets merged, the compiler should also use `Surface.components/2` with the proper `only_current_project` and `only_web_namespace` options, which should cover all cases we have. I'm already using that function in both, `surface_catalogue` and the `catalogue_test/2` macro.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -6,7 +6,8 @@ locals_without_parens = [
   data: 3,
   data: 2,
   slot: 1,
-  slot: 2
+  slot: 2,
+  catalogue_test: 1
 ]
 
 [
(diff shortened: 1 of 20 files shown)


tiagoefmoraes (approved):
Making catalogue examples will be much easier!

PR #609, opened 2022-05-31
Support phoenix components with slots

### LINKED ISSUE #580: Unable to use Phoenix.Component slots from within surface.
## Describe the bug
It does not seem possible to use `slots` of regular vanilla `Phoenix.Components` inside surface. E.g. Doing something like this inside a surface liveview render `~F` does not work:
```
<.my_component>
  <:header>
    <div style={"background-color: lightgreen;"}>Content for header slot.</div>
  </:header>

  <div style={"background-color: lightblue;"}>Content for body/default slot.</div>

  <:footer>
    <div style={"background-color: lightgray;"}>Content for footer slot.</div>
  </:footer>
</.my_component>
```

## How to reproduce it
1. I've created a repo to easily reproduce the issue: [slot_sample](https://github.com/LQJacques/slot_sample). Please see the README there for details.

## The behavior you expected
I would expect to be able use regular `Phoenix.Component`s with slots (such as the sample above) from within a surface component/live view.

## Your Environment
Surface: v0.7.1
LiveView: v0.17.7
Elixir: v1.12.2
```
$ elixir -v
Erlang/OTP 24 [erts-12.0] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [jit]
Elixir 1.12.2 (compiled with Erlang/OTP 24)
```

msaraiva:
Fixes #580.

### DIFF
--- a/lib/surface/compiler/eex_engine.ex
+++ b/lib/surface/compiler/eex_engine.ex
@@ -498,44 +498,31 @@ defmodule Surface.Compiler.EExEngine do
   end
 
   # Function component
-  defp build_slot_props(%AST.FunctionComponent{fun: fun} = component, buffer, state, _context_var)
-       when fun != nil do
-    slot_info =
-      component.slot_entries
-      |> Enum.map(fn {name, slot_entries} ->
-        state = %{state | scope: [:slot_entry | state.scope]}
+  defp build_slot_props(%AST.FunctionComponent{fun: fun} = component, buffer, state, _ctx_var) when fun != nil do
+    for {name, slot_entries} <- component.slot_entries,
+        state = %{state | scope: [:slot_entry | state.scope]},
+        nested_slot_entries = handle_slot_entries(component, slot_entries, buffer, state),
+        not Enum.empty?(nested_slot_entries) do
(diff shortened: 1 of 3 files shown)


lib/surface/compiler/eex_engine.ex
```diff
@@ -498,44 +498,31 @@ defmodule Surface.Compiler.EExEngine do
...
-       when fun != nil do
-    slot_info =
-      component.slot_entries
-      |> Enum.map(fn {name, slot_entries} ->
-        state = %{state | scope: [:slot_entry | state.scope]}
+  defp build_slot_props(%AST.FunctionComponent{fun: fun} = component, buffer, state, _ctx_var) when fun != nil do
+    for {name, slot_entries} <- component.slot_entries,
+        state = %{state | scope: [:slot_entry | state.scope]},
+        nested_slot_entries = handle_slot_entries(component, slot_entries, buffer, state),
+        not Enum.empty?(nested_slot_entries) do
```
herminiotorres, line 505: I have a question here, instead of using `Enum.empty?(nested_slot_entries)` can we check `nested_slot_entries != []`? If it is possible can we change? WDYT?

PR #613, opened 2022-06-06
New generator and argument API for Slots

tiagoefmoraes:
## New Slot Argument API

```elixir
defmodule Component do
  prop users, :list
  slot user, arg: %{name: :string, age: :integer}
  slot count, arg: :integer

  def render(assigns) do
    ~F"""
    <#slot {@count, length(@users)} />
    {#for user <- @users}
      <#slot {@user, name: user.name, age: user.age} />
    {/for}
    """
  end
end
...
def render(assigns) do
  ~F"""
  <Component users={@users}>
    <:count :let={user_count}>{user_count} user(s)</:count>
    <:user :let={name: name}>Name: {name}</:user>
  </Component>
  """
end
```

## New Slot Generator API

```elixir
defmodule Component do
  prop items, :generator, root: true
  slot default, generator_prop: :items, arg: :integer

  def render(assigns) do
    ~F"""
    {#for {item, index} <- Enum.with_index(@items)}
      <#slot {@default, index} generator_value={item} />
    {/for}
    """
  end
end
...
def render(assigns) do
  ~F"""
  <Component {l <- ["one", "two"]} :let={i}>
    #{i + 1} {l}
  </Component>
  """
end
```


msaraiva (commented):
@tiagoefmoraes this looks great! Are we replacing the type `:list` with `:generator` in this PR or in a separate one?

tiagoefmoraes:
> @tiagoefmoraes this looks great! Are we replacing the type `:list` with `:generator` in this PR or in a separate one?

In this one!

tiagoefmoraes:
This are my notes on future improvements to this API.

- Argument
  - [ ] validations for `<Component>`
    - [ ] `:let={...}` is present but no `default` `slot` defined
    - [ ] `:let={...}` is present but no `arg:` defined in `slot`
  - [ ] validations for `<#slot />`
    - [ ] `arg={...}` is missing if `arg:` defined in `slot`
    - [ ] `arg={...}` is present but no `arg:` defined in `slot`
  - [ ] validations for `<:slot />`
    - [ ] `:let={...}` is present but no `arg:` defined in `slot`
  - [ ] validations for `slot`
    - [ ] validate that `arg:` value is:
      - [ ] a member of `@type` from `lib/surface/api.ex`
        - ex: `slot default, arg: :integer`
      - [ ] or a map with atom keys and values from `@type`
        - ex: `slot default, arg: %{form: :any}`
      > implement in `defp validate_opt_ast!(:slot, :arg, arg_ast, _caller) do`?
- Generator
  - [ ] validations for `<#slot />`
    - [ ] `generator_value={...}` is missing if `generator_prop:` defined in `slot`
      currenctly unexpected error
    - [ ] `generator_value={...}` is present but no `generator_prop:` defined in `slot`
  - [ ] validations for `prop`
    - [ ] if type `:generator` is used and is not associated with a `slot`

- [ ] documentation for slot argument
- [ ] include information of where `:let` and `:arg` were defined (file, line, ...) in `cannot match slot argument against :let`?

Malian:
Hi @tiagoefmoraes ! Thanks for this amazing work.

Could you please provide the reasons why this changes in the api have been implemented? The reason why I am asking is that it's not obvious to me the reasons behind these changes. To avoid any misunderstanding, I am in favour of these changes but I am wondering if we introduced it for technical reasons, because LV, or anything else.


Most of these changes are related to naming things. For example:
```
<#slot name="count">
# in favour of
<#slot for={@count}>
```

```
:args
# in favour of
:arg
```
knowing that the docs says:
> arg: quoted expression representing argument__S__ for this slot


Then, could we add a `converters` to help users upgrade their application? Is it feasible from a technical point of view? I am concerning about the time it will take to update an application by hand.

Finally, is the PR ready to be tested on a real project? If so, I would be happy to do it next week (16/08) as I have to update one of our project in production.

tiagoefmoraes:
Hello @Malian, thank you for reviewing the PR.

> I am wondering if we introduced it for technical reasons, because LV, or anything else.

The change from `args` to `arg` (`slot default, arg: :integer` and `<#slot arg={10}>`) was to be compatible with [LV render_slot/2](https://hexdocs.pm/phoenix_live_view/0.17.11/Phoenix.LiveView.Helpers.html#render_slot/2). Previous versions of Surface only allowed the syntax `slot default, args: [:index]` and `<#slot arg={index: 10}>`, in other words we always required the argument to be a `map`. Now the `arg` can be any value and in `:let` we can pattern match more freely, and allow compatibility for vanilla live_view slots/components.

Some examples:

- `<slot arg={10}>` -> `<... :let={index}>`
- `<slot arg={index: 10}>` -> `<... :let={index: i}>` (the argument will be %{index: 10})
- `<slot arg={index: 10}>` -> `<... :let={%{index: i}>`
- `<slot arg={index: 10}>` -> `<... :let={map}>`
- `<slot arg={[index: 10]}>` -> `<... :let={[index: i]}>`
- `<slot arg={[index: 10]}>` -> `<... :let={keyword_list}>`

About `<#slot name="count" index={index} /> to `<#slot for={@count} />`, `name` and `index` have been "scheduled" [to be deprecated](https://github.com/surface-ui/surface/blob/b1c934ad144b992cac5015479d0961b10cdfd371/lib/surface/compiler.ex#L1283-L1302) for some time, we just decided to do it in this PR, `for` was already available and it's name was was chosen with the help of Surface users, @msaraiva may remember more details. 

`generator_prop` and `generator_value` was more technical to separate the generator value from the original slot argument, for better compatibility with LV.

> knowing that the docs says:
> 
> > arg: quoted expression representing argument__S__ for this slot

I updated the docs, it is just one argument but the user can give a map, list, keyword list or any other data type.

> Then, could we add a `converters` to help users upgrade their application? Is it feasible from a technical point of view? I am concerning about the time it will take to update an application by hand.

We can break this PR in 3 changes

- generator - this is a complex change to `prop`, `slot` and `<#slot />`, this feature is not used frequently. We can check if the compiler errors/warnings are enought to help with the migration.

    ```elixir
    # before
    prop items, :list
    slot cols, args: [item: ^items]
    ~F"""
    {#for item <- @items}
      <#slot :args={item: item} />
    {/for}
    """
    ```

    ```elixir
    # after
    prop items, :generator
    slot cols, generator_prop: :items
    ~F"""
    {#for item <- @items}
      <#slot generator_value={item} />
    {/for}
    """
    ```

- `<#slot />` argument, `arg` - as the value will stay the same (a map) this can be a change the `converter` can do, It's basically changing the `:args` directive into the `arg` attribute. The compiler errors/warnings could help with this too.

    ```elixir
    # before
    ~F"""
    <#slot :args={item: item} />
    """
    ```

    ```elixir
    # after
    ~F"""
    <#slot arg={item: item} />
    """
    # new possibility
    ~F"""
    <#slot arg={item} />
    """
    ```

- `slot` `arg:` option - in the new API this will only be used as documentation for the catalogue. The compiler gives a deprecation warning when using `:args`. This is also complex and the value of the option has changed drastically, only the user can do this change.

    ```elixir
    # before
    slot cols, args: [:name, :age]
    ~F"""
    <#slot :args={name: name, age: age} />
    """
    ```

    ```elixir
    # after
    slot cols, arg: %{name: :string, age: :integer}
    ~F"""
    <#slot arg={name: name, age: age} />
    """
    ```

> Finally, is the PR ready to be tested on a real project? If so, I would be happy to do it next week (16/08) as I have to update one of our project in production.

We're still discussing some changes on the API, but the internals will remain mostly the same. Doing a test to see if the application compiles and runs would really help us find out if it needs improvements.

Malian:
@tiagoefmoraes what an amazing answer! Thanks :heart: 

I'll keep you informed when I upgrade my application with this branch.

msaraiva:
@Malian you should probably wait a bit before upgrading to this branch, I talked to Tiago yesterday and we agreed that the best syntax would be to just map what we have in LV, which is the `render_slot/2` call. So we'll change `<#slot>` to accept a `render` prop that will receive whatever `render_slot/2` does. The prop will be set as `root: true` so we'll be able to suppress its name.

## Examples

* Default slot

  - Without args
    `<#slot/> or <#slot {@default}/>`

  - With args
    `<#slot {@default, info}/>`

* Named slots

  - Without args
   `<#slot {@header}/>`

  - With args
    `<#slot {@header, info}/>`

* Slot entry (var)

  - Without args
    `<#slot {col}/>`

  - With args
    `<#slot {col, item}/>`

msaraiva:
@tiagoefmoraes fantastic work! 🔥🔥🔥

PR #617, opened 2022-06-13
add a config option to suppress undefined prop warnings

mayel:
replaces https://github.com/surface-ui/surface/pull/588


msaraiva:
Hi @mayel!

Thanks for the PR. 

Closing this now since we'll be using LV's declarative API soon so if something like that necessary, we'll have to implement it there.

Thanks again!

PR #618, opened 2022-06-14
Add tests to dynamic components with the function attribute

Malian:


msaraiva:
@Malian thank you! ❤️

PR #621, opened 2022-07-04
Scoped CSS

msaraiva:
This PR adds support for scoped CSS using a similar approach used by Vue.js as specified [here](https://vuejs.org/api/sfc-css-features.html#scoped-css).

The main advantages of this approach is:

- Better code organization as each component style can be defined along with the component itself
- Avoid CSS conflicts between components
- Prevent CSS rules from the parent component from leaking into children elements or other components
- Allows injecting dynamic elixir expressions into the CSS style keeping the CSS syntax valid.

Initial features planned for this PR:

- [x] Define scoped CSS rules using `<style>`
- [x] Define scoped CSS rules using colocated files with extension `.css` (only module components for now)
- [x] Inject dynamic elixir expression in the CSS using `s-bind`
- [x] Support `:deep()` pseudo-class 
- [x] Update the `surface.init` task to set up the project for scoped CSS 

## Setup

1. Add the `surface` compiler (in case you haven't yet)
  ```elixir
  def project do
    [
     ...
      compilers: [:gettext] ++ Mix.compilers() ++ [:surface],
    ]
  end
  ```
2. Add `@import "./_components.css";` at the top of the `app.css` file:
  ```css
  @import "./_components.css";

  @tailwind base;
  @tailwind components;
  @tailwind utilities;
  ...
  ```

The `./_components.css` will be automatically generated by the surface compiler.

## Example

```elixir
defmodule Demo do
  use Surface.LiveView

  data count, :integer, default: 0
  data color, :string, default: "red"

  def handle_event("inc", _params, socket) do
    count = socket.assigns.count + 1
    {:noreply, assign(socket, count: count)}
  end

  def render(assigns) do
    ~F"""
    <style>
      .btn {
        background-color: s-bind('Enum.at(["red", "green", "blue"], rem(@count, 3))');
        @apply ml-4 inline-flex border-0 py-2 px-6 rounded text-lg;
      }
    </style>

    <div>
      <h1 class="p-8 text-xl">{@count}</h1>
      <hr>
      <button :on-click="inc" class="btn">
      +++
      </button>
    </div>
    """
  end
end
```


lib/surface/compiler/eex_engine.ex
```diff
@@ -894,6 +894,24 @@ defmodule Surface.Compiler.EExEngine do
 
   defp to_html_attributes([]), do: []
 
+  # defp to_html_attributes([
```
tiagoefmoraes, line 897: We're leaving this commented out code?

tiagoefmoraes (approved):
This looks great Marlus!

lib/surface/compiler/eex_engine.ex
msaraiva: No way! :)

Issue #620, opened 2022-06-29
Confusing error message when incorrectly using the default slot


davydog187:
## Describe the bug
When trying to render the default slot as a named slot, you will get the following error message

> Error: cannot render <div> (templates are only allowed as children elements of components, but found template for slot)

## How to reproduce it
1. Create a component with a default slot `slot default`
2. In the template, render the slot as named `<:slot />`
3. Observe the above error

## The behavior you expected
Have a better error message, such as:

> No slot found named `slot`. Did you mean `<#slot />`?

## Your Environment
Surface: v0.7.4
LiveView: v0.17.10

Hex:    1.0.1
Elixir: 1.13.4
OTP:    25.0

msaraiva:
Fixed in #628.

PR #628, opened 2022-07-29
add a better error message when incorrectly using the default slot

herminiotorres:
Add a better error message as Dave reporting on this issue here: https://github.com/surface-ui/surface/issues/620

Screen Shot 2022-07-29 at 11 47 30

Screen Shot 2022-07-29 at 11 48 12

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -879,8 +879,7 @@ defmodule Surface.Compiler do
   defp validate_tag_children([]), do: :ok
 
   defp validate_tag_children([%AST.SlotEntry{name: name, meta: meta} | _]) do
-    {:error,
-     "slot entries are not allowed as children of HTML elements. Did you mean <##{name} />?", meta}
+    {:error, "slot entries are not allowed as children of HTML elements. Did you mean <##{name} />?", meta}
   end
 
   defp validate_tag_children([_ | nodes]), do: validate_tag_children(nodes)
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -878,7 +878,8 @@ defmodule Surface.Compiler do
   defp validate_tag_children([]), do: :ok
 
   defp validate_tag_children([%AST.SlotEntry{name: name} | _]) do
-    {:error, "slot entries are only allowed as children elements of components, but found slot entry for #{name}"}
+    {:error,
+     "Slot entries are only allowed as children elements of components. No slot found named slot. Did you mean <##{name} />"}
   end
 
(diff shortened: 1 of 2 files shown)


lib/surface/compiler.ex
```diff
@@ -878,7 +878,8 @@ defmodule Surface.Compiler do
   defp validate_tag_children([]), do: :ok
 
   defp validate_tag_children([%AST.SlotEntry{name: name} | _]) do
-    {:error, "slot entries are only allowed as children elements of components, but found slot entry for #{name}"}
+    {:error,
+     "Slot entries are only allowed as children elements of components. No slot found named slot. Did you mean <##{name} />"}
```
davydog187, line 882: I think this should only be the message if it's the default slot

herminiotorres: I add other named slots and tried to ensure this only happen with the default slot.

herminiotorres: good catch, I add a new function for named slots, not just only default ones, and it raises the same error message when using `:` instead of `#`

davydog187, line 882: I think you need a conditional, something like below. I also cleaned up the grammar of the message

```suggestion
   message =
     case name do
       "slot" -> "No slot found named slot. Did you mean the default slot? Use <#slot />"
       _ -> "Slot entries are only allowed as children elements of components, but found slot entry for #{name}"
     end
    {:error, message}
```

davydog187: Note i haven't tested this, I'm just trying to provide an example

herminiotorres: but the syntax for named slots is:

```
slot header

~F"""
  <div>
    <:slot name="body" />
  </div>
"""
```

it will match with the "slot" condition in `case`, and I need to track the props attributes values to see if it is a default slot or a named slot. I am thinking in this way:


```elixir
  defp validate_tag_children([%AST.SlotEntry{name: name, props: []} | _]) do
    {:error,
     "Slot entries are only allowed as children elements of components. No slot found named slot. Did you mean <##{name} />"}
  end

  defp validate_tag_children([%AST.SlotEntry{name: name, props: props} | _]) do
    [%Surface.AST.Attribute{value: attrs_values}] = props
    named_value = Map.get(attrs_values, :value, nil)

    {:error,
     "Slot entries are only allowed as children elements of components. No slot found named slot. Did you mean <##{name} name=\"#{named_value}\" />"}
  end
```

WDYT? @davydog187

msaraiva:
@herminiotorres thank you! ❤️

PR #632, opened 2022-08-05
fix when directive are specified multiple times

herminiotorres:
it should show a warning when specified directive multiple times, like:

```html
<div :on-click="a" :on-click="b"></div>
```

### DIFF
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -272,8 +272,6 @@ defmodule Surface.Compiler do
             the directive `:#{format_directive_name(directive.name)}` has been passed multiple times. Considering only the last value.
 
             Hint: remove all redundant definitions.
-
-            See the directives are supported: https://surface-ui.org/template_syntax#directives
             """
 
             IOHelper.warn(message, meta.caller, meta.file, meta.line)
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -282,7 +282,7 @@ defmodule Surface.Compiler do
               do: MapSet.put(processed_directives, directive.name),
               else: processed_directives
 
-          {mod.process(directive, node), MapSet.put(processed_directives, directive.name)}
+          {mod.process(directive, node), processed_directives}
       end
 
     directives
--- a/lib/surface/compiler.ex
+++ b/lib/surface/compiler.ex
@@ -262,11 +262,25 @@ defmodule Surface.Compiler do
   defp node_type(_), do: :text
 
   defp process_directives(%{directives: directives} = node) when is_list(directives) do
+    {directives, _} =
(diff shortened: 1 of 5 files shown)


test/surface/compiler_test.exs
```diff
@@ -756,6 +756,35 @@ defmodule Surface.CompilerSyncTest do
     assert line == 2
   end
 
+  test "warning when directive are specified multiple times" do
```
msaraiva, line 759:
```suggestion
  test "warning when a directive is specified multiple times in an HTML element" do
```

test/mix/tasks/compile/surface/validate_components_test.exs
```diff
@@ -106,6 +106,31 @@ defmodule Mix.Tasks.Compile.Surface.ValidateComponentsTest do
            ]
   end
 
+  test "should return diagnostic when directive are specified multiple times" do
```
msaraiva, line 109:
```suggestion
  test "should return diagnostic when a directive is specified multiple times in a component" do
```

lib/surface/compiler.ex
```diff
@@ -262,11 +262,30 @@ defmodule Surface.Compiler do
...
 
   defp process_directives(%{directives: directives} = node) when is_list(directives) do
+    {directives, _} =
+      for %AST.Directive{module: mod, meta: meta} = directive <- directives,
+          function_exported?(mod, :process, 2),
+          reduce: {node, MapSet.new()} do
+        {node, processed_directives} ->
+          if MapSet.member?(processed_directives, directive.name) and match?(%AST.Tag{}, node) do
+            message = """
+            the directive `:on-#{directive.name}` has been passed multiple times. Considering only the last value.
```
msaraiva, line 272: What if the directive is not an `:on-*`? E.g. `:if`. We should probably have a test for this case too.

lib/surface/compiler.ex
```diff
@@ -262,11 +262,30 @@ defmodule Surface.Compiler do
   defp node_type(_), do: :text
 
   defp process_directives(%{directives: directives} = node) when is_list(directives) do
+    {directives, _} =
+      for %AST.Directive{module: mod, meta: meta} = directive <- directives,
+          function_exported?(mod, :process, 2),
+          reduce: {node, MapSet.new()} do
+        {node, processed_directives} ->
+          if MapSet.member?(processed_directives, directive.name) and match?(%AST.Tag{}, node) do
```
msaraiva, line 270:
```suggestion
          if match?(%AST.Tag{}, node) and MapSet.member?(processed_directives, directive.name) do
```

msaraiva (commented):
@herminiotorres great work! 

I left a couple of comments for your consideration.

msaraiva:
@herminiotorres thank you! ❤️

Issue #631, opened 2022-08-03
Support `export default` when using a single hook


msaraiva:
Setting directive `:hook` without a value should set its value to "default".

This allows us to export the hook as just:

```
export default {
  mounted() {
   ...
  }
}
```

and using it in the template as:

```
<div id={@id} :hook>
```

Currently, this works but you need to set it explicitly:

```
<div id={@id} :hook="default">
```

msaraiva:
Fixed in #636.

PR #636, opened 2022-08-12
add support to export default when use a single hook

herminiotorres:
Setting directive `:hook` without a value should set its value to "default".

Close this issue https://github.com/surface-ui/surface/issues/631


herminiotorres:
Hey @paulstatezny 

I working on this implementation, and when I run `mix format`, the format change:
```elixir
<div :hook></div>
```

into this:

```elixir
<div :hook={true}></div>
```

And @msaraiva and I, take a look at the surface format, and we found working around to not format this directive, since `:hook` is not a boolean directive and this implementation brings us a new situation never pass before when this directive could assume as a default hook when doesn't have any value.

msaraiva:
@paulstatezny for now, we could add a new clause:

```elixir
  defp render_attribute({":hook", true, _meta}, _opts),
    do: "#{name}"
```

above this one:

https://github.com/surface-ui/surface/blob/0f49505ee8a4535919d88ca7db7c62d4a29aa44d/lib/surface/formatter/phases/render.ex#L334

Then later we could implement a more generic long-term solution that would be to store in the metadata the information about the original format (e.g. `:dir` vs `:dir={true}`) so we can properly format any directive or attribute.

WDYT?

paulstatezny:
Thanks for the contribution @herminiotorres :)

@msaraiva I agree with your suggestion completely. We should add that clause that targets `:hook` until the metadata includes information to handle it more generically.

msaraiva:
@herminiotorres in this case, I believe it's better to implement this small change in the formatter in the scope of this PR.

msaraiva:
@herminiotorres thank you! ❤️

PR #638, opened 2022-08-17
Adds option to override the web module on init

supersimple:
I talked to @davydog187 about this...

I had an issue today. My project is named `RandomUUID` but the file is `random_uuid.ex`, not `random_u_u_i_d.ex`.
Given the last part is an acronym, it is correct for the entire part to be capitalized, but I never use the naming convention in phoenix to make that happen automagically (because it looks terrible.)
In this case, I just find/replace the module name after running the phoenix generator. However, this caused a problem when running the Surface init task. It assumes the web module was `RandomUuidWeb`.
This option in the task makes it possible to override that.
It seems like a small footprint change, and I hope it is useful to more people than just me. :D

### DIFF
--- a/lib/mix/tasks/surface/surface.init.ex
+++ b/lib/mix/tasks/surface/surface.init.ex
@@ -52,6 +52,8 @@ defmodule Mix.Tasks.Surface.Init do
 
     * `--no-install` - do not fetch and install added/updated dependencies.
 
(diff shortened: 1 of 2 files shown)


davydog187 (commented):
Thanks @supersimple, this PR looks great to me! I'm not familiar with this part of the codebase so I'll let one of the other @surface-ui/core members give their approval

msaraiva:
@supersimple thanks! ❤️

PR #640, opened 2022-08-26
add test for :hook directive without any attribute

herminiotorres:


paulstatezny (approved):
❤️ Thank you!

test/surface/formatter/plugin_test.exs
```diff
@@ -106,4 +110,16 @@ defmodule Surface.Formatter.PluginTest do
...
+  test ":hook directive without any attribute" do
+    assert_formatter_doesnt_change(
+      "sface_files.sface",
+      """
+      <div :hook />
+      """,
+      """
+      <div :hook />
+      """
+    )
```
paulstatezny, line 123: Actually, for `assert_formatter_doesnt_change`, I think the point of it is that you only have to specify the code once. (See `test/surface/formatter_test.exs`)

Can we make that change?

paulstatezny (dismissed):
(See request about `assert_formatter_doesnt_change`)

PR #642, opened 2022-08-29
Hide internals of formatter in docs

paulstatezny:
This PR hides formatter modules that aren't part of the public API, leaving only two:

1. `mix surface.format`
2. `Surface.Formatter.Plugin`

### DIFF
--- a/lib/mix/tasks/surface/surface.format.ex
+++ b/lib/mix/tasks/surface/surface.format.ex
@@ -2,8 +2,10 @@ defmodule Mix.Tasks.Surface.Format do
   @shortdoc "Formats Surface ~F sigils and .sface files in the given files/patterns"
 
   @moduledoc """
-  **To format Surface code using Elixir 1.13 or later, use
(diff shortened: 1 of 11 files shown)


lib/mix/tasks/surface/surface.format.ex
```diff
@@ -2,8 +2,10 @@ defmodule Mix.Tasks.Surface.Format do
   @shortdoc "Formats Surface ~F sigils and .sface files in the given files/patterns"
 
   @moduledoc """
-  **To format Surface code using Elixir 1.13 or later, use
-  `Surface.Formatter.Plugin`.**
+  > #### Formatter Plugin Preferred {: .tip}
+  >
+  > To format Surface code using Elixir 1.13 or later, use
+  > `Surface.Formatter.Plugin`.
```
paulstatezny, line 8: After we update to the latest `ex_doc`, this admonition block will render like this:

Screen Shot 2022-08-29 at 2 49 47 PM

PR #644, opened 2022-09-01
Add elixir 1.14.0 to CI

capitalist:


davydog187:
It looks like the CI fails. I ran it twice so I don't believe it's a flaky test

msaraiva:
It has been fixed on `main`.

@capitalist thanks! 💚

capitalist:
Just fyi @paulstatezny brought up a good point on our repos that we might not want warnings and formatter errors for 1.13 and 1.14 in the event that 1.14 releases an update where that satisfies the compiler but emits warnings on older compilers. We currently didn't have an instances of this in our decently sized internal monorepo, but it's something to be aware of.

Issue #552, opened 2022-01-12
v0.7 introduces compile time dependencies


dvic:
I believe https://github.com/surface-ui/surface/blob/master/lib/surface/base_component.ex#L62 introduces various compile time dependencies. I just cleaned up my project to get rid of the troublesome so-called "compile-connected" dependencies and with the latest version (as of time of writing this is d80b3e347537d07990d7cd2d3be7495a2af1fad1) many compile dependencies returned.

Is there another way to implement liveview 0.17 support without the need for these `require` statements? Maybe generate another module at compile time to avoid requiring the original module?

msaraiva:
Hi @dvic!

> I believe https://github.com/surface-ui/surface/blob/master/lib/surface/base_component.ex#L62 introduces various compile time dependencies.

Yes. Currently, module components (as opposed to pure function components ) must be defined as compile-time dependencies so we can do static checking based on the information introspected from those components. This has always been the case. One exception recently discovered was that if you're using Surface, running on elixir v1.13, due to some compiler optimizations, some of those dependencies will not be created, bringing a bunch of problems. That was the reason we needed to change the way we force the dependencies in `BaseComponent`.

> Is there another way to implement liveview 0.17 support without the need for these `require` statements? Maybe generate another module at compile time to avoid requiring the original module?

Yes. I'm already working on a unified [Declarative API](https://github.com/phoenixframework/phoenix_live_view/pull/1747) for both Surface and Liveview that will introduce a separate compiler that will do those checks without the need to add compile-time dependencies. Until then, we need those dependencies.

Besides the created dependencies (which is expected), are you facing any other issue? Errors? Warnings? Out-of-sync modules?, etc.

dvic:
> Yes. Currently, module components (as opposed to pure function components ) must be defined as compile-time dependencies so we can do static checking based on the information introspected from those components. This has always been the case. One exception recently discovered was that if you're using Surface, running on elixir v1.13, due to some compiler optimizations, some of those dependencies will not be created, bringing a bunch of problems. That was the reason we needed to change the way we force the dependencies in BaseComponent.

We are indeed on Elixir 1.13. Not sure how it was done before, but surface 0.6 did not introduce compile dependencies when I call a component inside another component (reported by `mix xref trace`).

> Besides the created dependencies (which is expected), are you facing any other issue? Errors? Warnings? Out-of-sync modules?, etc.

[...]

PR #653, opened 2022-09-15
Remove compile-time deps from components

msaraiva:
Removes `compile` deps from components and turn them into `export`. 

The basic rules you can expect when changing a component from now on are:

* Changes in the public API (props, slots, component type, etc.) trigger recompilation only of direct callers (no more transitive deps)
* Other changes, including changes inside `render/1`, trigger recompilation only of the component itself
* For macro components and components implementing `transform/1`, the behavior is exactly as an elixir macro, so it creates a `compile` dep which keeps track of transitive compile-time deps.

### DIFF
--- a/lib/surface/base_component.ex
+++ b/lib/surface/base_component.ex
@@ -133,18 +133,96 @@ defmodule Surface.BaseComponent do
 
     components = Enum.uniq_by(components_calls, & &1.component)
 
-    requires =
-      for %{component: mod, line: line} <- components, mod != env.module do
-        quote line: line do
-          require(unquote(mod)).__info__(:module)
+    {imports, requires} =
+      for %{component: mod, line: line, dep_type: dep_type} <- components, mod != env.module, reduce: {[], []} do
+        {imports, requires} ->
+          case dep_type do
+            :export ->
+              {[
+                 quote line: line do
+                   import unquote(mod), warn: false
+                 end
+                 | imports
+               ], requires}
+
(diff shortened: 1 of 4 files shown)


tiagoefmoraes (approved):

dvic:
Wow, amazing :D This PR reduces in our project the compile-connected deps from 800+ to just 4!! 🎉

We can say definitly it fixes #552 😉

dvic:
I just tried updating to 0.8.2, but I get these weird dialyzer errors:

```
lib/my_web/some_file.ex:1:pattern_match
The pattern can never match the type.

Pattern:
false

Type:
true
```

I tried resetting the cache but it didn't help. Any idea?

dvic:
Maybe it's because of the following new code?

```elixir
    imports_block =
       quote do
         if true do
           unquote(imports)
         end
       end
```

msaraiva:
The `if` block there is just a simple way to scope the imports so they don't leak to the rest of the module. We just want to make sure we create the `export` dependencies there. We don't need to actually import anything. I guess I'll have to replace that with something else that can satisfy dialyzer. 

Any suggestion? I don't use dialyzer at all :)

dvic:
Hmm... removing that if statement does not resolve the issue and I can't reproduce it in a new repo. Have to investigate this further... Makes me want to drop dialyzer altogether 😂

dvic:
See #654, it's introduced in a commit before this PR

PR #655, opened 2022-09-18
Fix dialyzer issue in EExEngine

### LINKED ISSUE #654: dialyzer complains since v0.8.2
Since e8aea39, I'm getting dialyzer errors across my project in the form of

```
lib/my_web/some_file.ex:1:pattern_match
The pattern can never match the type.

Pattern:
false

Type:
true
```

cc @tiagoefmoraes

dvic:
Fixes #654

### DIFF
--- a/lib/surface/compiler/eex_engine.ex
+++ b/lib/surface/compiler/eex_engine.ex
@@ -606,7 +606,7 @@ defmodule Surface.Compiler.EExEngine do
                 unquote(context_var)
               } ->
                 unquote(
-                  quote line: let_line do
+                  quote line: let_line, generated: true do
                     if !match?(unquote(no_warnings_let), argument) do
                       raise ArgumentError,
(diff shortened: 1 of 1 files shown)


dvic:
I also added `generated: true` to the second block, because although my project didn't trigger that part, I think it suffers from the same problem.

msaraiva:
@dvic thanks! ❤️

msaraiva:
Hey @dvic! We're still getting those dialyzer errors even after the PR. We've made some changes that made the problem disappear for us. Can you check if that branch also solves it for you? https://github.com/surface-ui/surface/tree/v0.8

janwillemvd:
@dvic is away this week, that's why I'm responding.

I can confirm that but [our dialyzer issue](https://github.com/surface-ui/surface/issues/654) was gone since ffaef38f874d9e93e4a1bc4bb94f83f009b3be79.

Also, https://github.com/surface-ui/surface/tree/v0.8 doesn't give any dialyzer warnings or errors 💪🏻.

We're still on LV 0.17 with this project.

PR #664, opened 2022-10-11
Deprecate `:for` directive.

djgoku:
Removing the use of the `:for` directives in tests required updating the tests' expected results.

Running the tests emits `:for` directive deprecation warnings. Tests need to still be included on the `:for` directive so this is a necessary evil.

### DIFF
--- a/lib/surface/directive/for.ex
+++ b/lib/surface/directive/for.ex
@@ -2,9 +2,8 @@ defmodule Surface.Directive.For do
   use Surface.Directive
 
   def extract({":for", {:attribute_expr, value, expr_meta}, attr_meta}, meta) do
-    # TODO: deprecate :for after releasing v0.7
-    # message = "directive :for has been deprecated. Please use the {#for ...} block instead."
-    # Surface.IOHelper.warn(message, meta.caller, expr_meta.line)
+    message = "directive :for has been deprecated. Please use the {#for ...} block instead."
+    Surface.IOHelper.warn(message, meta.caller, expr_meta.line)
 
     %AST.Directive{
       module: __MODULE__,
--- a/test/mix/tasks/compile/surface/validate_components_test.exs
+++ b/test/mix/tasks/compile/surface/validate_components_test.exs
@@ -342,7 +342,9 @@ defmodule Mix.Tasks.Compile.Surface.ValidateComponentsTest do
 
     def render(assigns) do
       ~F"""
-      <span :for={v <- @prop}>value: {v}</span>
+      {#for v <- @prop}
+        <span>value: {v}</span>
+      {/for}
       """
     end
   end
--- a/test/surface/components/context_test.exs
+++ b/test/surface/components/context_test.exs
@@ -90,9 +90,11 @@ defmodule Surface.Components.ContextTest do
     def render(assigns) do
       ~F"""
       <Context put={field: "field from OuterWithNamedSlots"}>
-        <span :for={slot <- @my_slot}>
-          <#slot {slot} />
-        </span>
+        {#for slot <- @my_slot}
+          <span>
+            <#slot {slot} />
+          </span>
+        {/for}
       </Context>
       """
     end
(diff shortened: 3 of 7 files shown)


msaraiva:
Hi @djgoku!

Thank you so much for the PR. I believe we need to rethink the decision to add this deprecation and keep `:for`, now that HEEx has introduced it to its own syntax. As far as I know, there might be some new LV features coming in the future that may depend on that syntax. 

My suggestion would be to remove the deprecation warning and update the docs accordingly. @Malian, @miguel-s and @davydog187, WDYT?

@djgoku please let me know if you'd be willing to work on that and thank you again for the PR.

miguel-s:
Really it all depends on this part:

> there might be some new LV features coming in the future that may depend on that syntax.

Any spoilers you can shere? 😁 

If Surface's template syntax `{#for}` can support these new features I would still vote for deprecating `:for`. Having two different ways of achieving the same thing is more work in maintenance and also more confusing to document.

If `{#for}` can't support these new features then having both "for" versions with different capabilities has the potential to become quite confusing. Maybe `{#for}` should be deprecated?

Malian:
> Maybe {#for} should be deprecated?

One of the main reason I use Surface is that the template engine allows us to have clearer templates compare to what Heex offers. I know this is maybe subjective and just my opinion, but my fear is that if we decide to deprecate `{#for}` why should we keep the other blocks expressions? And, if we decide to keep the other block expressions, I am pretty sure we will receive requests to add `{#for}` every month ^^

I am also afraid that if we decide to keep the deprecation, it will affect the Surface user adoption. In conclusion, I am in favor of keeping `:for`.

miguel-s:
> One of the main reason I use Surface is that the template engine allows us to have clearer templates compare to what Heex offers.

+1 on this. I love the template syntax and `{#for}`!

Maybe I worry too much, but I think having `{#for}` and `:for` will be confusing if the functionality of both diverges.

Malian:
> Maybe I worry too much, but I think having {#for} and :for will be confusing if the functionality of both diverges.

Malian:
Sorry, miss click for the close/open the pull request ^^

> Maybe I worry too much, but I think having {#for} and :for will be confusing if the functionality of both diverges.
It could be an issue indeed.

msaraiva:
There's no way will ever deprecate `{#for}` as it is the only way that allow us to work with multiple nodes. The question is only if we should keep `:for`.

My idea is to keep it for now until we have a better picture of how it will evolve in LV. There are some ideas to use it in cases where `{#for}` won't fit. One example is if we ever want to allow dynamic slots, which is been a common request by users, there might be a chance to implement it in a more natural way using `:for` as it keeps the requirement of the slot being a direct child of the parent component defining the slot. There might also be some developments regarding using streams to dynamically update view content.

LV already accept `<%= for ... do %>` and `:for` as well as the `if` counterparts. We already agreed we're keeping `:if`, so it doesn't sound that odd to me. If in the future `:for` turns out to be redundant, we can circle back to the discussion.

miguel-s:
> LV already accept <%= for ... do %> and :for

That's a great point. Always forget about the LV syntax...

> We already agreed we're keeping :if

Missed this discussion, sorry. I agree that if we keep `:if` it does make sense to also keep `:for`.

davydog187:
@msaraiva I would be in favor of keeping `:for` in its current form, for two reasons

1. I believe there is actual utility in it over `{#for}` for building concise lists
2. For the reasons you highlighted with the future of LV

djgoku:
> My suggestion would be to remove the deprecation warning and update the docs accordingly. @Malian, @miguel-s and @davydog187, WDYT?

@msaraiva I'll remove the deprecation warning. But I am not sure what is meant by update the docs. I'll see both `{#for` [1] and `:for`
[2] are documented. So I am not sure if there are any other places that need to be updated.

Should I keep all my test changes that switched from `:for` to `{#for`?

Sorry for the late reply.

[1] https://github.com/surface-ui/surface_site/blob/e8058cd5fa4d5c684743325f7e8a850437bd571f/lib/surface_site_web/live/template_syntax.ex#L207-L227
[2] https://github.com/surface-ui/surface_site/blob/e8058cd5fa4d5c684743325f7e8a850437bd571f/lib/surface_site_web/live/template_syntax.ex#L272-L273

msaraiva:
Since LV implemented `:for` as well, we're not deprecating it anymore. Thank you all for the discussion.

PR #666, opened 2022-10-17
Missing shortdoc on surface.init prevents task discovery from `mix help`

gangstead:
I've been a long time user of surface but was recently setting it up on a new project for the first time.  

I noticed the `surface.init` mix task from the docs wasn't showing up in the list of tasks from `mix help`. 

A missing @shortdoc seems to be the cause. The other two tasks have it and show up in the help output.


Malian:
❤️

PR #671, opened 2022-11-07
Click away

bryannaegele:
Expose the phx click away event that was added recently.

https://hexdocs.pm/phoenix_live_view/bindings.html#click-events


Malian:
Thanks for the PR @bryannaegele !

I believe a prop is missing in `Surface.Components.Events` to fix the tests.

Finally, could you also please add a test to `Surface.Components.EventsTest`.

After that, we will be able to merge.

bryannaegele:
Awesome. I couldn't figure out which file to do it in. I'll fix that up soon.

PR #678, opened 2023-02-14
Fix scoped css per module and add variants

### LINKED ISSUE #677: Hot code reloading of Surface View depends on mix (in running server)
## How to reproduce it
1. Open a running server's remote console and attempt to live patch the code of a Surface View module.

## The behavior you expected
Expected the code to be patched and the new module code to be in use. Instead a "module Mix is not available" error appears:

image

## Your Environment
Surface: v0.9.1
LiveView: v0.18.3
Elixir: v1.14.3

msaraiva:
# Changes

* Fixes the problem of creating a new scope for each function component. Currently, function components cannot have their own css file so they need to share the styles with the main component implementing `render/1`. This fix makes any function component in the same module share the same scope.
* Introduces an experimental feature the allows the compiler to generate Tailwind variants for `props` or `data` assigns by using `css_variant: true` as an option. This feature is disable by default and must be explicitly enabled using the `enable_variants` compiler option in `config.exs`.
* When generating the combined `_components.css` file, we now move all `@import` entries collected from the components to the top of the file to adhere to the CSS spec. Without this change, only the imports of the first component would apply.

### Boolean

```elixir 
data active, :boolean, default: false, css_variant: true
```

```html 
<div class="bg-gray-400 active:bg-green-400">
  ...
</div>

<div class="bg-green-400 not-active:bg-gray-400">
  ...
</div>
```

###  Choices

```elixir 
data size, :string, values: ["small", "medium", "large"], default: "small", css_variant: true
```

```html 
<div class="size-small:text-sm size-medium:text-base size-large:text-lg">
  ...
</div>
```

###  Lists

```elixir 
data items, :list, default: [], css_variant: true
```

```html 
<div class="hidden has-items:block">
  ...
</div>

<div class="no-items:hidden">
  Number of items: {length(@items)}
</div>
```

## Notes

* The generated variants are scoped so they don't conflict with variants from other components
* All generated variants can also be used inside the colocated css file

## TODO

* [x] Allow users to change the name of the variant
* [ ] Validate variant name against existing pre-defined variants from TW. Emit a warning instructing the user to rename it.


tiagoefmoraes (approved):
Looks great @msaraiva!

PR #684, opened 2023-03-23
chore: upgrade phoenix_live_view to v0.18.18

### LINKED ISSUE #683: Compile error on form.ex
I get a compiler error on form.ex

```== Compilation error in file lib/surface/components/form.ex ==
** (CompileError) lib/surface/components/form.ex:101: the function "inner_block" cannot handle clauses with the -> operator because it is not a macro. Please make sure you are invoking the proper name and that it is a macro
    expanding macro: Surface.sigil_F/2
```

Have tried clean build, etc. 

## Your Environment
Surface: v0.9.4
LiveView: v0.18.17
Elixir: v1.14.3

frankdugan3:
Fixes #683


Menkir:
An important note: Because the new TagEngine is now used, tag attributes are no longer dasherized. This means that the form attributes like :phx_submit are converted to "phx_submit". So the event handling doesn't work anymore. 

[see here the latest LiveVIew change](https://github.com/phoenixframework/phoenix_live_view/commit/6b414a96baa49c2857e37a8663552cad05840c59#diff-58d0de731ce58b8b66b6248c9a33e29ae35ea024d832e7375d556c1fc0a5039fR893)
```elixir
  def attributes_escape(attrs) do
    # We don't want to dasherize keys, which Phoenix.HTML does for atoms,
    # so we convert those to strings

    attrs
    |> Enum.map(fn
      {key, value} when is_atom(key) -> {Atom.to_string(key), value}
      other -> other
    end)
    |> Phoenix.HTML.attributes_escape()
  end
```

[Surface form options](https://github.com/surface-ui/surface/blob/main/lib/surface/components/form.ex#L88)
```elixir
    # form.ex
    event_opts =
      event_to_opts(assigns.change, :phx_change) ++
        event_to_opts(assigns.submit, :phx_submit) ++
        event_to_opts(assigns.auto_recover, :phx_auto_recover)
```

msaraiva:
@frankdugan3 thank you for the PR! 

I merged it into [update-lv-to-0.18.18](https://github.com/surface-ui/surface/tree/update-lv-to-0.18.18) so I can fix the tests and work on other compatibility issues.

@Menkir thank you for reporting this. I'll take a look at it.

msaraiva:
> An important note: Because the new TagEngine is now used, tag attributes are no longer dasherized. This means that the form attributes like :phx_submit are converted to "phx_submit" which is not the same as "phx-submit". So the event handling doesn't work anymore.

@Menkir are you sure about this? I'm running the tests using LV `0.18.18` and Phoenix `1.7.2` and according to the following test, the changes don't seem to affect the generated code:

https://github.com/surface-ui/surface/blob/3adc9c544e49ca29b4225f9f4718037731ca572e/test/surface/components/form_test.exs#L124-L135

PR #685, opened 2023-03-28
avoids get_attribute from crashing when including a component in a MacroComponent

mayel:
Not sure if the best fix, please let me know :)

### DIFF
--- a/lib/surface/compiler/eex_engine.ex
+++ b/lib/surface/compiler/eex_engine.ex
@@ -1145,7 +1145,7 @@ defmodule Surface.Compiler.EExEngine do
     caller = component.meta.caller
 
     caller_is_module_component? =
-      Module.get_attribute(caller.module, :component_type) && caller.function == {:render, 1}
+      get_module_attribute(caller.module, :component_type) && caller.function == {:render, 1}
 
     # NOTE: Using this is not optimized as it will always assume the component propagates
     # context into into its slots.
@@ -1417,7 +1417,7 @@ defmodule Surface.Compiler.EExEngine do
   end
 
   defp get_propagate_context_to_slots_map(caller) do
-    Module.get_attribute(caller.module, :propagate_context_to_slots_map) ||
+    get_module_attribute(caller.module, :propagate_context_to_slots_map) ||
       Surface.BaseComponent.build_propagate_context_to_slots_map()
   end
 
@@ -1436,4 +1436,13 @@ defmodule Surface.Compiler.EExEngine do
 
     {call, [slots: slots] ++ meta, args}
   end
+
+  defp get_module_attribute(module, attribute) do
+    if function_exported?(module, :__info__, 1) do
+      module.__info__(:attributes)[attribute]
+    else
+      Module.get_attribute(module, attribute)
+    end
+  end
+
 end


msaraiva:
Hey @mayel! Thank you for the PR.

Could you provide a minimal example of a MacroComponent reproducing this issue. I want to make sure this is the best way to handle it.

mayel:
Sure thing! This seems to do it:

```elixir
Application.put_env(:sample, SamplePhoenix.Endpoint,
  http: [ip: {127, 0, 0, 1}, port: 5001],
  server: true,
  live_view: [signing_salt: "aaaaaaaa"],
  secret_key_base: String.duplicate("a", 64)
)

Mix.install([
  {:plug_cowboy, "~> 2.5"},
  {:jason, "~> 1.0"},
  {:surface, "~> 0.9.4"},
  {:phoenix_live_view, "<= 0.18.4", override: true},
  {:iconify_ex, git: "https://github.com/bonfire-networks/iconify_ex"}
])

defmodule SamplePhoenix.ErrorView do
  def render(template, _), do: Phoenix.Controller.status_message_from_template(template)
end

defmodule SamplePhoenix.Icon do
    use Surface.MacroComponent

    # any icon from iconify: https://icones.js.org
    prop iconify, :string, required: false, static: true

    prop class, :css_class, default: "w-4 h-4"

    def expand(attributes, _content, meta) do
        static_props =
        Surface.MacroComponent.eval_static_props!(__MODULE__, attributes, meta.caller)

        class = Surface.AST.find_attribute_value(attributes, :class) || ""

        icon = static_props[:iconify]

        quote_surface do
            ~F"""
            <Iconify class={^class} iconify={^icon} />
            """
        end
    end
end
defmodule SamplePhoenix.SampleLive do
  use Surface.LiveView, layout: {__MODULE__, :live}

  def mount(_params, _session, socket) do
    {:ok, assign(socket, :count, 0)}
  end

  def render("live.html", assigns) do
    ~H"""
    <script src="https://cdn.jsdelivr.net/npm/phoenix@1.7.0-rc.2/priv/static/phoenix.min.js"></script>
    <script src="https://cdn.jsdelivr.net/npm/phoenix_live_view@0.18.2/priv/static/phoenix_live_view.min.js"></script>
    <script>
      let liveSocket = new window.LiveView.LiveSocket("/live", window.Phoenix.Socket)
      liveSocket.connect()
    </script>
    <style>
      * { font-size: 1.1em; }
      .full { width: 100%; height: 100%; }
    </style>
    <%= @inner_content %>
    """
  end

  def render(assigns) do
    ~F"""

    <#SamplePhoenix.Icon iconify="solid:heart" class="full" />
    """
  end
end

defmodule Router do
  use Phoenix.Router
  import Phoenix.LiveView.Router

  pipeline :browser do
    plug(:accepts, ["html"])
  end

  scope "/", SamplePhoenix do
    pipe_through(:browser)

    live("/", SampleLive, :index)
  end
end

defmodule SamplePhoenix.Endpoint do
  use Phoenix.Endpoint, otp_app: :sample
  socket("/live", Phoenix.LiveView.Socket)
  plug(Router)
end

{:ok, _} = Supervisor.start_link([SamplePhoenix.Endpoint], strategy: :one_for_one)
Process.sleep(:infinity)
```

msaraiva:
Hey @mayel!

Whenever using `quote_surface`, we need to make sure the AST is generated using the correct caller, otherwise, it will use `__ENV__`, which for this case is the macro component itself, not the caller component. To fix this, all you have to do is to explicitly pass the caller, like:

```elixir
quote_surface caller: meta.caller do
  ~F"""
  <Iconify.iconify class={^class} icon={^icon} />
  """
end
```

I fell into this pitfall a couple of times already. I wonder if we should make the `caller` option required to force the user to always reason about which caller should be used.

mayel:
Aha, thanks! Maybe this would work as way of documenting that?
```
  defp get_module_attribute(module, attribute) do
    if function_exported?(module, :__info__, 1) do
      raise "Cannot apply `Module.get_attribute(#{module}, #{attribute})` because the module is already compiled. Hint: If using `quote_surface`, you need to make sure the AST is generated using the correct caller, otherwise, it will use __ENV__, which for this case is the macro component itself, not the caller component. To fix this, all you have to do is to explicitly pass the caller, i.e.: `quote_surface caller: meta.caller do ...`"
    else
      Module.get_attribute(module, attribute)
    end
  end
  ```

msaraiva:
I'd rather avoid such change for now but we could certainly add a note to the [quote_surface docs](https://github.com/surface-ui/surface/blob/main/lib/surface.ex#L133). Something like:

> **NOTE:** If you're using `quote_surface` inside the `expand/3` callback of a macro component, you mostly want to pass the caller explicitly to make sure the AST is generated using the correct caller component, otherwise, it will leave the macro component itself as the caller, which may lead to confusing compile-time errors.

Then we could also change the example to show the code inside an `expand/3` call using `quote_surface` passing `caller: meta.caller`.

WDYT?

Feel free to submit a PR with those changes, if you want.

mayel:
👍 Will do

PR #691, opened 2023-05-29
Fix Context.get/put in LiveComponent update

tiagoefmoraes:
Allow calling `Context.get(socket, :field)` and `Context.put(socket, field: "value")` in [`update/2`](`c:Phoenix.LiveComponent.update/2`)

Calling it giving `assigns` will not work because there's no context information there. It's different from a `render/1` assigns.

If the user is defining `prop`s or `data`s with `from_context` they will have to call `assign(socket, assigns)` for them to be updated (as with any other assigns).

### DIFF
--- a/lib/surface/components/context.ex
+++ b/lib/surface/components/context.ex
@@ -140,8 +140,9 @@ defmodule Surface.Components.Context do
 
   Without scope:
 
-      Context.put=(key1: @value1, key2: "some other value")
+      Context.put(key1: @value1, key2: "some other value")
 
+  > **Note**: When using `put/3` in [`update/2`](`c:Phoenix.LiveComponent.update/2`) call it giving the `socket`.
   """
   def put(socket_or_assigns, scope \\ nil, values)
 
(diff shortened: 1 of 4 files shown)


lib/surface/components/context.ex
```diff
@@ -140,8 +140,9 @@ defmodule Surface.Components.Context do
 
   Without scope:
 
-      Context.put=(key1: @value1, key2: "some other value")
+      Context.put(key1: @value1, key2: "some other value")
 
+  > **Note**: When using `put/3` in [`update/2`](`c:Phoenix.LiveComponent.update/2`) call it giving the `socket`.
```
msaraiva, line 145:
```suggestion
  > **Note**: Whenever using `Context.put/3` inside the [`update/2`](`c:Phoenix.LiveComponent.update/2`) callback, make sure you call it passing the `socket`, not the `assigns`.
```

WDYT?

lib/surface/components/context.ex
```diff
@@ -187,6 +188,8 @@ defmodule Surface.Components.Context do
       ~F"\""
       <MyTextInput form={form} field={field} />
       "\""
+
+  > **Note**: When using `put/3` in [`update/2`](`c:Phoenix.LiveComponent.update/2`) call it giving the `socket`.
```
msaraiva, line 192:
```suggestion
  > **Note**: Whenever using `Context.get/3` inside the [`update/2`](`c:Phoenix.LiveComponent.update/2`) callback, make sure you call it passing the `socket`, not the `assigns`.
```

WDYT?

PR #720, opened 2023-12-28
fix: ranges for 1.16

mhanberg:
Fixes the remaining ranges that cause runtime warnings in Elixir 1.16


mhanberg:
The tests don't seem to actually pass on elixir 1.16, but they passed on 1.14. once the CI is approved, we'll see if they pass on 1.15.

mhanberg:
Sorry, got carried away, will open a PR for all 1.16 fixes

PR #721, opened 2023-12-28
test: fix for 1.16

mhanberg:
error messages are more elaborate in 1.16, this uses a regex to allow tests to pass on all supported versions


mhanberg:
Sorry, got carried away, will open a PR for all 1.16 fixes, same as the other PR

Issue #713, opened 2023-09-23
Compatibility with LiveView 0.20


mayel:
Our codebase compiles and runs with no issues using LV 0.20.0 and Surface 0.11.0 :)

It seems that support for [debug annotations](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#module-debug-annotations) needs to be added to Surface though. That's all I've noticed so far.

noozo:
Related to enhancement released in LV 0.20.0: 

```
Add heex debug annotations via config :phoenix_live_view, :debug_heex_annotations: true, 
which provides special HTML comments that wrap around rendered components to help you 
identify where markup in your HTML document is rendered within your function component tree
```

endorphin3d:
I installed LiveView 0.20.2 with Surface 0.11.1 and I discovered one bug: When you embed a `Surface.LiveComponent` from a `Surface.LiveView`, the LiveComponent's `update/3` function gets called twice on page load (I just put an `IO.inspect` in the `update` body). I then downgraded LiveView to 0.19.0 and found that `update` was now only being called once.

endorphin3d:
On further testing, I've realized this isn't an issue with Surface. The same thing happens to me when I have a plain Phoenix live component embedded by a live view.

dougw-bc:
Another LiveComponent issue I noticed, which might be related, and is not fixed in `fix-lv-20-compat` is all my Surface LiveComponents when calculating diffs (i think) it errors with `more than a single root node`.
There is a single root node, but it is Surface component, by using a plain `div` as the root node of the LC fixes it....

I think `update/3` being called twice could have something to do with it...

msaraiva:
@dougw-bc is that a server or client error? Live components must have a single root node but that root node cannot be another component. As far as I can remember we even have validation for that.

dougw-bc:
It is not a LiveComponent but a regular Component at the root of a LiveComponent - which was working prior to liveview 0.20.2. If this was a deliberate upstream liveview change - no problem adding a wrapper `div` seems to solve the problem...

LiveView now throws this error - but as mentioned ^ maybe it is intentional: 

```
** (ArgumentError) error on BCWeb.DirectInviteModal.render/1 with id of "direct-invite-modal". Stateful components must have a single static HTML tag at the root
  (phoenix_live_view 0.20.3) lib/phoenix_live_view/diff.ex:344: Phoenix.LiveView.Diff.component_to_rendered/3
```

[...]

PR #725, opened 2024-01-03
Compatibility with LV v0.20 

msaraiva:
* Fix Malformed HTML in rendered.js
* Fix error when passing special chars as literals inside an expression


tiagoefmoraes (approved):

dougw-bc:
I have run into some other issues that are not fixed with this branch related to LiveComponents, I made a comment about it here: https://github.com/surface-ui/surface/issues/713

dougw-bc:
Ignore my previous comment, this branch is OK on liveview `0.20.3`

Users do need to be careful, and make sure all `LiveComponent`'s use static html at the root - but it is an easy enough change. 

### For example:

<details>
<summary>In previous versions the following works OK, (but maybe poorly optimized?):
 </summary>
 
 ```elixir
defmodule StaticComponent do
  @moduledoc "Static component used by a livecomponent"
  use Surface.Component 
  
  slot default
  
  def render(assigns) do
   ~F"""
    <div>
     <#slot />
    </div>
   """
  end
end

defmodule ALiveComponent do
  @moduledoc "live component with static component at root"
  use Surface.LiveComponent 

  def render(assigns) do
    ~F"""
     <StaticComponent>
       <div>
         Some LiveComponent Content
       </div>
     </StaticComponent>
    """
  end
end

defmodule ALiveView do
  @moduledoc "LiveView that renders the livecomponent"
  use Surface.LiveView

  def render(assigns) do
    ~F"""
     <ALiveComponent id="a_live_component" />
    """
  end
end
```
</details>

<details>
<summary>This PR requires the following change to the above example live component (additional wrapping static node): </summary>

```elixir
defmodule ALiveComponent do
  @moduledoc "additional static html node at root of live component"
  use Surface.LiveComponent 

  def render(assigns) do
    ~F"""
     <div>
      <StaticComponent>
        <div>
          Some LiveComponent Content
        </div>
      </StaticComponent>
     </div> 
    """
  end
end
```
</details>

PR #728, opened 2024-01-05
LiveView 0.20.2+

harmon25:
- updates to lv 0.20.3
- adds `phoenix_html_helpers` for compatibility
- replaces uses of `Phoenix.HTML.Form` with `PhoenixHTMLHelpers.Form`

PR #735, opened 2024-01-26
Update deps

tiagoefmoraes:
- Depend on `:jason` only for tests
- Lower the requirement on `:phoenix_live_view` to `~> 0.19.0` and test it in CI
- Update other dependencies
- Test with updated dependencies in CI
- Declare `:phoenix_html` as dependency to avoid updating to 4.0.0 that [has breaking changes](https://github.com/phoenixframework/phoenix_html/blob/main/CHANGELOG.md#v400-2023-12-19), we'll update later, the changes are in https://github.com/surface-ui/surface/pull/728
- Update `:floki`
- Update GitHub actions to avoid Node 16 deprecated warnings
  - after https://github.com/erlef/setup-beam/pull/249 is merged we can depend on the version again


msaraiva (approved):

msaraiva:
@tiagoefmoraes thank you! 💚

PR #740, opened 2024-02-22
ci: fixes tests by reinstating phoenix_html 3.x requirement

grzuy:
Hi!

Seems like after the latest extraction to surface_form, tests are running against incompatible (right?) phoenix_html 4.x.


tiagoefmoraes:
Thanks for the PR, we actually want to support the newer version too. [This commit](https://github.com/surface-ui/surface/commit/27e076aad9b3600ab7975f03b8c0994fa97d94ff) removes the problematic dependency on `Phoenix.HTML.Tag` that was used only on the tests.

grzuy:
Great! Better :-)

PR #743, opened 2024-04-02
Support LV >= 0.20.10

marpo60:
Hi again 👋 

[In a previous PR](https://github.com/surface-ui/surface/pull/742), I though that the CI was not running correctly in latest LV.

After seeing the changes in Phoenix Live View 0.20.10 I can confirm that I was mistaken and the implementation of debug annotation in LV changed so surface needs to adjust accordingly

References:
[Add HEEx debug annotations](https://github.com/phoenixframework/phoenix_live_view/pull/2813) in 0.20.0
[Move body annotation as a HTMLEngine callback](https://github.com/phoenixframework/phoenix_live_view/commit/d7c24e8e3bf17fe5ffe1abfceddd206f065440f3) in 0.20.10

Open questions:
Surface 0.11.2 is released already and in the mix.exs its support `~> 0.19.0 or ~> 0.20.0` This is wrong in the sense that 0.11.2 is not able to run with 0.20.10. Not sure what we want to do there, maybe add something in the Changelog that explains this.

Let me know what you think.


tiagoefmoraes:
Thank you @marpo60 and @grzuy! This was merged in https://github.com/surface-ui/surface/commit/784ffdf6117d58f7ae099e200c772327086f777e

Issue #754, opened 2024-10-09
Error on printing diagnostics when hot code reloading


mayel:
## Describe the bug

```
warning: variable "opts" is unused (if the variable is not meant to be used, prefix it with an underscore)
  lib/static_generator/maybe_static_generator_plug.ex:13:32
** (Protocol.UndefinedError) protocol String.Chars not implemented for {39, 9} of type Tuple. This protocol is implemented for the following type(s): Atom, BitString, Cldr.Calendar.Duration, Cldr.Currency, Cldr.LanguageTag, Cldr.LanguageTag.T, Cldr.LanguageTag.U, Cldr.Unit, Complex, Date, DateTime, Decimal, [...]
    (elixir 1.17.3) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir 1.17.3) lib/string/chars.ex:22: String.Chars.to_string/1
    (elixir 1.17.3) lib/exception.ex:769: Exception.format_file_line/3
    (elixir 1.17.3) lib/exception.ex:638: Exception.format_stacktrace_entry/1
    (elixir 1.17.3) src/elixir_errors.erl:79: :elixir_errors."-print_diagnostic/2-lc$^0/1-0-"/1
    (elixir 1.17.3) src/elixir_errors.erl:79: :elixir_errors.print_diagnostic/2
    (elixir 1.17.3) src/elixir_errors.erl:106: :elixir_errors.emit_diagnostic/6
    (surface 0.12.0) lib/mix/tasks/compile/surface.ex:200: anonymous fn/2 in Mix.Tasks.Compile.Surface.print_diagnostics/1
    (elixir 1.17.3) lib/enum.ex:2531: Enum."-reduce/3-lists^foldl/2-0-"/3
    (surface 0.12.0) lib/mix/tasks/compile/surface.ex:199: Mix.Tasks.Compile.Surface.print_diagnostics/1
    (surface 0.12.0) lib/mix/tasks/compile/surface.ex:191: Mix.Tasks.Compile.Surface.handle_diagnostics/2
    (mix 1.17.3) lib/mix/task.ex:495: anonymous fn/3 in Mix.Task.run_task/5
    (phoenix 1.7.14) lib/phoenix/code_reloader/server.ex:307: Phoenix.CodeReloader.Server.run_compilers/3
    (phoenix 1.7.14) lib/phoenix/code_reloader/server.ex:335: Phoenix.CodeReloader.Server.with_logger_app/2
[...]

PR #755, opened 2024-10-09
quick fix for hot code reload error

mayel:
This fixes https://github.com/surface-ui/surface/issues/754 though we may want to use the column instead of ignore it?

### DIFF
--- a/lib/mix/tasks/compile/surface.ex
+++ b/lib/mix/tasks/compile/surface.ex
@@ -196,9 +196,14 @@ defmodule Mix.Tasks.Compile.Surface do
   end
 
   defp print_diagnostics(diagnostics) do
+    IO.inspect(diagnostics)
     for %Diagnostic{message: message, severity: severity, file: file, position: position} <- diagnostics do
       print_diagnostic(message, severity, file, position)
     end
+  end
+
+  defp print_diagnostic(message, :warning, file, {line, _column}) do
+    print_diagnostic(message, :warning, file, line)
   end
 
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/compile/surface.ex
```diff
@@ -201,6 +201,10 @@ defmodule Mix.Tasks.Compile.Surface do
     end
   end
 
+  defp print_diagnostic(message, :warning, file, {line, _column}) do
+    print_diagnostic(message, :warning, file, line)
+  end
+
```
msaraiva, line 207: Hey @mayel! Thanks for the PR. 

I wonder if could have something like this, instead:

```elixir
if Version.match?(System.version(), ">= 1.14.0") do    
  defp print_diagnostic(message, :warning, file, {line, col}) do
    IO.warn(message, file: file, line: line, column: col)
  end
end

# TODO: Remove this clause in Surface v0.13 and set required elixir to >= v1.14
defp print_diagnostic(message, :warning, file, line) do
  rel_file = file |> Path.relative_to_cwd() |> to_charlist()
  IO.warn(message, [{nil, :__FILE__, 1, [file: rel_file, line: line]}])
end
```

This way most users could already benefit from the column info and have better diagnostics. 

@tiagoefmoraes WDYT?

tiagoefmoraes: Best of both worlds!

msaraiva:
Fixed in #755.

@mayel thank you!

